Issue #9, opened 2016-11-07
64 Bit integers


jimhester:
### Issue Description and Expected Result
64 Bit integers are able to be read and written from databases. Currently they are converted to 32 bit native R integers (i.e. nonsense).

### Database
All databases that support 64 bit integers.

### Reproducible Example
``` r
library(DBI)
library(odbconnect)
con <- dbConnect(odbconnect::odbconnect(), dsn = "PostgreSQL")
dbGetQuery(con, "SELECT 10000000000 as a")
#>            a
#> 1 1410065408
```

### Session Info
<details>

``` r
devtools::session_info()
#> Session info -------------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.3.0 (2016-05-03)
#>  system   x86_64, darwin13.4.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  tz       America/New_York            
#>  date     2016-11-07
#> Packages -----------------------------------------------------------------
#>  package    * version     date       source                            
#>  assertthat   0.1         2013-12-06 CRAN (R 3.3.0)                    
#>  blob         0.0.0.9000  2016-11-03 Github (jimhester/blob@d677541)   
#>  clipr        0.2.1       2016-06-23 CRAN (R 3.3.0)                    
#>  colorspace   1.2-7       2016-10-11 cran (@1.2-7)                     
#>  crayon       1.3.2       2016-06-28 cran (@1.3.2)                     
#>  DBI        * 0.5-12      2016-10-13 Github (rstats-db/DBI@4f00863)    
#>  DBItest      1.3-10      2016-11-03 local (jimhester/DBItest@8284aef) 
#>  devtools     1.12.0.9000 2016-10-28 Github (hadley/devtools@f0d1c7e)  
#>  digest       0.6.10      2016-08-02 CRAN (R 3.3.0)                    
#>  evaluate     0.10        2016-10-11 cran (@0.10)                      
#>  ggplot2    * 2.1.0.9001  2016-10-28 Github (hadley/ggplot2@cd6a691)   
#>  gtable       0.2.0       2016-02-26 CRAN (R 3.3.0)                    
#>  htmltools    0.3.5       2016-03-21 CRAN (R 3.3.0)                    
#>  knitr        1.14.15     2016-11-07 Github (yihui/knitr@8cbc61a)      
#>  lazyeval     0.2.0.9000  2016-06-15 Github (hadley/lazyeval@c155c3d)  
[...]

PR #15, opened 2016-11-26
Improve behavior with bigint types

bquinart:
To address #9. Change mapping of bigint to double instead of integer.


bquinart:
Apologies for the lack of follow-up.
Cancelling the PR as I do not have enough experience to dig into DBItest and figure out the problem.
From the little I understand, it looks like 64 bit tests are disabled, so I am not sure why it fails.

Till a "true 64 bit R type" gets implemented. I still think it could make sense to map DB bigint type to R numeric type. This is also what RODBC does.

PR #29, opened 2017-02-20
Signal custom conditions on unknown fields

### LINKED ISSUE #26: Unknown field type -11 with uniqueidentifier and SQL Server
dbGetQuery reports "Unknown field type (-11) in column..." with uniqueidentifier

It's a warning only, results look ok (character)

### Database
SQL Server 13.0.1722 under Windows 64

odbc 1.0.1 from CRAN

jimhester:
Rather than throwing a warning.

They can be caught with
  withCallingHandlers(expr, odbc_unknown_field_type = function(e) {...} )

Fixes #26

### DIFF
--- a/src/condition.h
+++ b/src/condition.h
@@ -1,0 +1,36 @@
+#ifndef CONDITION_H_
+#define CONDITION_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include <Rinternals.h>
+
+static SEXP signal_condition(const char * msg, const char * class_, SEXP env = R_GlobalEnv) {
+  SEXP condition, c, signalConditionFun, out;
+
+  condition = PROTECT(Rf_mkNamed(VECSXP, (const char *[]) { "message", ""}));
+
+  PROTECT(c = Rf_allocVector(STRSXP, 2));
+  SET_STRING_ELT(c, 0, Rf_mkChar(class_));
+  SET_STRING_ELT(c, 1, Rf_mkChar("condition"));
+
+  SET_VECTOR_ELT(condition, 0, Rf_mkString(msg));
+  Rf_setAttrib(condition, R_ClassSymbol, c);
+  signalConditionFun = Rf_findFun(Rf_install("signalCondition"), R_BaseEnv);
+
+  SEXP call = PROTECT(Rf_lang2(signalConditionFun, condition));
+  out = Rf_eval(call, env);
+
+  UNPROTECT(3);
+
+  return out;
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* CONDITION_H_ */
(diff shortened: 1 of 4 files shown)


src/odbc_result.h
```diff
@@ -435,7 +436,10 @@ class odbc_result {
             break;
           default:
             types.push_back(string_t);
-            Rcpp::warning("Unknown field type (%s) in column %s", type, r.column_name(i));
+
+            char buf[100];
```
hadley: Make this a `odbc_unknown_field_type()` function?

src/condition.h
```diff
@@ -0,0 +1,36 @@
...
+  PROTECT(c = Rf_allocVector(STRSXP, 2));
+  SET_STRING_ELT(c, 0, Rf_mkChar(class_));
+  SET_STRING_ELT(c, 1, Rf_mkChar("condition"));
+
+  SET_VECTOR_ELT(condition, 0, Rf_mkString(msg));
+  Rf_setAttrib(condition, R_ClassSymbol, c);
+  signalConditionFun = Rf_findFun(Rf_install("signalCondition"), R_BaseEnv);
+
+  SEXP call = PROTECT(Rf_lang2(signalConditionFun, condition));
+  out = Rf_eval(call, env);
```
hadley: Doesn't this need a `PROTECT` too?

src/condition.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+#include <Rinternals.h>
+
+static SEXP signal_condition(const char * msg, const char * class_, SEXP env = R_GlobalEnv) {
+  SEXP condition, c, signalConditionFun, out;
```
hadley: And why are you defining all these up front?

jimhester: I was being strictly c89 compliant in this header, which requires variable declarations at the top of a block.

src/condition.h
jimhester: I don't know, if you are immediately going to be unprotecting it in the next statement I thought that would be unnecessary, but maybe not?

hadley: That's true, but isn't it safer to always protect? Otherwise you'll have to modify previous code if you add anything extra to this function

jimhester: From https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Garbage-Collection I get the impression the GC would only be called from R functions or macros, so this is probably safe as is. On the other hand the performance difference is likely negligible, so might as well add it!

hadley: Your argument would be more compelling if the first code example wasn't 

```c
ab = PROTECT(allocVector(REALSXP, 2));
REAL(ab)[0] = 123.45;
REAL(ab)[1] = 67.89;
UNPROTECT(1);
```

😉 

(but we should write up a consistent philosophy at https://github.com/hadley/r-internals)

jimhester: I think the idea there is potentially `REAL()` could have code that could trigger the garbage collector (even though it does not), but anyway we agree it is safer to add the PROTECT to avoid issues if additional code is added.

PR #31, opened 2017-03-01
Update example in Odbc driver

mnel:
change capitalisation to all lower case as per package and function name


jimhester:
Thanks!

PR #38, opened 2017-03-20
First pass at Hive data types

edgararuiz:
Based on experience, mapped datetime to "TIMESTAMP" and character  to "STRING"

### DIFF
--- a/R/DataTypes.R
+++ b/R/DataTypes.R
@@ -7,6 +7,7 @@
 #' - SQL Server
 #' - SQLite
 #' - Spark
+#' - Hive
 #'
 #' If you are using a different database and `dbWriteTable()` fails with a SQL
 #' parsing error the default method is not appropriate, you will need to write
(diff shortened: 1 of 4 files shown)


R/DataTypes.R
```diff
@@ -77,6 +78,22 @@ odbcDataType.default <- function(info, obj, ...) {
 }
 
 #' @export
+`odbcDataType.Hive` <- function(info, obj, ...) {
+  switch_type(obj,
+    factor = "VARCHAR(255)",
```
jimhester: If character is "STRING" I think factor and list should be as well, as those types are converted to character vectors on the R side before being passed to the ODBC code.

jimhester:
Thanks Edgar, looks good, one minor point above and could you add a note to NEWS.md with the change and mentioning this PR # and yourself `(#38, @edgararuiz)`.

R/DataTypes.R
edgararuiz: I thought about that, but wasn't sure how homogenous you'd like to have the type conversion for this function.  I'll make the change

jimhester:
Thanks!

PR #39, opened 2017-03-21
Adding support for Impala data types.  

edgararuiz:
Impala does not have a date type so mapping it to varchar


jimhester:
From Magnitude's documentation it does look like there is a DATE type, and looks like STRING should be CHAR?

screenshot 2017-03-22 08 28 13

I just pushed a commit (https://github.com/rstats-db/odbc/commit/ef071fd536e2a9d551aac6201fea4febfd41a6ac) that exposes a `test_roundtrip()` function, which will try to roundtrip a data frame with all of the data types supported by the odbc package. Could you try impala with that to verify this method is correct for all data types?

edgararuiz:
Here's the article that discusses the differences between HiveQL and Impala: https://www.cloudera.com/documentation/enterprise/5-8-x/topics/impala_langref_unsupported.html#langref_hiveql_unsupported

edgararuiz:
Test results: 

```
Error: Test failed: '[Impala] round tripping data.frames works' * <SQL> 'CREATE TABLE `it` ( `datetime` TIMESTAMP, `date` DATE, `binary` BINARY, `integer` INTEGER, `double` DOUBLE, `character` VARCHAR(255), `logical` VARCHAR(5) ) ' nanodbc.cpp:1587: HY000: [RStudio][ImpalaODBC] (110) [HY000] : ImpalaRuntimeException: CAUSED BY: InvalidObjectException: Invalid column type: invalid_type 1: DBI::dbWriteTable(con, "it", it, overwrite = TRUE) 2: DBI::dbWriteTable(con, "it", it, overwrite = TRUE) 3: .local(conn, name, value, ...) 4: dbGetQuery(conn, sql) 5: dbGetQuery(conn, sql) 6: dbSendQuery(conn, statement, ...) 7: dbSendQuery(conn, statement, ...) 8: OdbcResult(connection = conn, statement = statement) 9: new_result(connection@ptr, statement)
```

jimhester:
Thanks Edgar!

PR #64, opened 2017-05-31
Fix preview and columns in viewer contract

javierluraschi:
Two fixes for  the viewer contract:
1. Previewing tables uses `dbQuoteIdentifier` but conneciton is missing.
2. Contract requires `table` parameter under `odbcListColumns`


jimhester:
Thanks Javier!

PR #84, opened 2017-07-05
FIX: dbSendStatement should NOT call result_execute

ruiyiz:
According to DBI specification, results are executed when **dbBind** or **dbGetRowsAffected** is called. Executing results in **dbSendStatement** causes errors when parameterized queries are being sent.

The change is tested successfully against SQL Server.


jimhester:
Thanks!

PR #93, opened 2017-07-27
Added Vertica data types

khotilov:
I took PostgreSQL types to start, and used more common type names. Reference: https://my.vertica.com/docs/7.2.x/HTML/index.htm#Authoring/SQLReferenceManual/DataTypes/SQLDataTypes.htm

Also, since vertica doesn't care about the order of rows unless specifically asked, I've added a parameter to `test_roundtrip` to be able to enforce the same row order before the comparison. The roundtrip succeeds with `force_sorted=TRUE`.


jimhester:
This looks good thanks! Can you add a bullet to NEWS.md with the change and `@` mention yourself and this PR number. Thanks!

khotilov:
@jimhester Done.

jimhester:
Looks good, thanks!

PR #99, opened 2017-08-09
Adds temp table support to Oracle DBs

edgararuiz:


edgararuiz:
cc @nwstephens
cc @jimhester

R/db_oracle.R
```diff
@@ -0,0 +1,23 @@
+#' @rdname hidden_aliases
```
jimhester: This would need a corresponding `hidden.R` file (https://github.com/cran/DBI/blame/882c3921ee58d1fa97298697bf2a50161a8e401a/R/hidden.R)

edgararuiz:
@jimhester - I can't tell if the CI failure is due to something in the code change, any pointers?

jimhester:
The failures weren't due to you, I fixed it in the master and restarted your build.

R/db_oracle.R
```diff
@@ -1,4 +1,3 @@
-#' @rdname hidden_aliases
```
jimhester: The method does need to be documented, so having this here is fine, you just need to add the hidden.R file as well. You can copy the one from DBI (https://github.com/cran/DBI/blob/882c3921ee58d1fa97298697bf2a50161a8e401a/R/hidden.R#L1-L8)

jimhester: And you will need to run devtools::document() afterwards.

edgararuiz: Ok, I will. Thanks

jimhester:
LGTM, Thanks!

PR #106, opened 2017-08-24
In dbWriteTable change parameter name from fieldTypes to field.types …

jschelbert:
…to be consistent with the DBI package

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@
 
 * Support table creation for Vertica (#93, @khotilov).
 
+* Changed parameter `fieldTypes` to `field.types` in functions `dbWriteTable()` and `sqlCreateTable()` to be compliant with DBI (#106, @jschelbert).
+
 # odbc 1.1.1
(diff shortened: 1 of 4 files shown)


jschelbert:
The DBI man page for dbWriteTable states
> The field.types argument must be a named character vector with at most one entry for each column. It indicates the SQL data type to be used for a new column.

sqlCreateTable has the additional argument "fieldTypes" in odbc (since 1.1.0).
How should consistency be done between DBI dbWriteTable (with _field.types_) and odbc sqlCreateTable (with _fieldTypes_)?

jimhester:
Yes you are correct this should be `field.types` to match the spec. Could you adjust `sqlCreateTable` accordingly as well?

jimhester:
And add a note to NEWS.md with this change, referencing this PR # and your GitHub handle?

jschelbert:
Sure thing, here you go. Keep up the good work :thumbsup:

PR #117, opened 2017-10-04
Fix forgetting negation

hoxo-m:
You might forget negation.

```r
> nzchar("")
[1] FALSE
> nzchar("something")
[1] TRUE
```


jimhester:
Yes thanks!

PR #147, opened 2018-01-10
Fixes #146

### LINKED ISSUE #146: odbc returns double quotes for empty IDENT
When using an aggregate function in `dplyr` without a group, the `GROUP BY`  clause is being operated as: `GROUP BY ""`, this causes an error. I was able to track down the issue to an empty IDENT being interpreted as an empty character, thus returning the double quote:

```
suppressPackageStartupMessages({
  library(dbplyr)
  library(dplyr)
  library(RSQLite)
  library(DBI)
})

con <- dbConnect(odbc::odbc(), Driver = "SQL Server", Server = "localhost\\SQLEXPRESS", 
  Database = "datawarehouse", Trusted_Connection = "True")

empty <- ident(as.character())
empty
#> <IDENT> [empty]

DBI::dbQuoteIdentifier(con, empty)
#> <SQL> ""

dbDisconnect(con)
```

The issue seems to be in the `dbQuoteIdentifier` method.  A simple fix should be to skip quoting if `x` is 0 length.

A PR is forthcoming for this.

edgararuiz:


hadley:
I think this would be cleaner if you did something more like if (length(x) == 0) return(SQL())

jimhester:
Yeah I am going to change it after I merge.

jimhester:
Done in https://github.com/r-dbi/odbc/commit/696dcd2a0813df04f3e2e5820eae6205ff2efeff

Unfortunately I just submitted an odbc release to CRAN this morning before this PR came in :(, but we can do another release sometime soon.

edgararuiz:
Great! Thank you.  If we can do a release prior to the rstudio::conf class would be awesome.

jimhester:
We can't, you can really only have one release a month

hadley:
Is the odbc release still in flight? If so, you might be able to abort it. 

Otherwise, I think you can blame dbplyr and say you need a patch release.

jimhester:
No, unfortunately it is accepted now. I can try to do a patch release sometime next week.

PR #149, opened 2018-01-25
Remove unused forwarders

krlmlr:
For r-dbi/DBI#218 (but I'm fixing this independently).

These looks like the remains from RKazam, need to find a better way to deal with it.


jimhester:
Thanks!

PR #182, opened 2018-05-16
Add support for sql editing from connections pane

javierluraschi:
Makes use of rstudio/rstudio#2791 to support a SQL action in the connections pane to edit SQL from an active connection.


javierluraschi:
Instead of assigning named connections, now we search for the connection in the global environment. If we don't find it, we generate a SQL preview with no connection for the user to fill out.

javierluraschi:
@jimhester we've been testing this one for a couple weeks, can we merge?

jimhester:
Thanks!

Issue #171, opened 2018-04-15
Query SQL Server table with text field hanging


renkun-ken:
### Issue Description and Expected Result

I use FreeTDS driver to query a table in SQL Server with the following definition:

```
   ORDINAL_POSITION     COLUMN_NAME        DATA_TYPE CHARACTER_MAXIMUM_LENGTH IS_NULLABLE
1                 1              ID              int                       NA          NO
2                 2       ORIGIN_ID              int                       NA         YES
3                 3            CODE          varchar                       50         YES
4                 4       CODE_NAME          varchar                       50         YES
5                 5           TITLE          varchar                      150         YES
6                 6         CONTENT             text               2147483647         YES
7                 7         TYPE_ID              int                       NA         YES
8                 8        ORGAN_ID              int                       NA         YES
9                 9          AUTHOR          varchar                      100         YES
10               10        SCORE_ID              int                       NA         YES
11               11  ORGAN_SCORE_ID              int                       NA         YES
12               12     CREATE_DATE         datetime                       NA         YES
13               13        PAUSE_ID              int                       NA         YES
14               14       INTO_DATE         datetime                       NA         YES
15               15           TEXT1          varchar                      200         YES
16               16           TEXT2          varchar                      200         YES
17               17           TEXT3          varchar                      200         YES
18               18           TEXT4            float                       NA         YES
19               19           TEXT5            float                       NA         YES
20               20           TEXT6            float                       NA         YES
21               21           TEXT7            float                       NA         YES
22               22           TEXT8            float                       NA         YES
23               23   PRICE_CURRENT            float                       NA         YES
24               24 CAPITAL_CURRENT            float                       NA         YES
25               25 FORECAST_RETURN            float                       NA         YES
26               26      EXPECT_MAX            float                       NA         YES
27               27      EXPECT_MIN            float                       NA         YES
28               28           TEXT9         datetime                       NA         YES
29               29          TEXT10         datetime                       NA         YES
30               30        FILEGUID uniqueidentifier                       NA         YES
31               31       ATTENTION              int                       NA         YES
32               32  ATTENTION_NAME          varchar                        8         YES
33               33      SCORE_FLAG              int                       NA         YES
34               34         TMSTAMP           bigint                       NA         YES
35               35       ENTRYDATE         datetime                       NA         YES
36               36       ENTRYTIME          varchar                       12         YES
```

[...]

PR #190, opened 2018-06-11
Use strlen to determine actual append length

renkun-ken:
This PR is an attempt to address #171 where each buffer in an iteration is likely to be terminated by multiple trailing zeros. `buffer_size - 1` is not safe enough to remove those trailing zeros. In this case, using `strlen()` is most reliable to determine the actual number of chars to append to `out` in case any early-terminating zeros are mixed into the buffer being appended.

This PR solves both string truncation and infinitely looping problems.

The upstream PR https://github.com/nanodbc/nanodbc/pull/176 has been merged.

### DIFF
--- a/src/nanodbc/nanodbc.cpp
+++ b/src/nanodbc/nanodbc.cpp
@@ -2817,14 +2817,22 @@ inline void result::result_impl::get_ref_impl<string_type>(short column, string_
                     buffer,          // TargetValuePtr
                     buffer_size,     // BufferLength
                     &ValueLenOrInd); // StrLen_or_IndPtr
+                // For col.ctype_ != SQL_C_BINARY, buffer is assumed to terminate by 0.
+                // In some cases, buffer has multiple trailing zeros which causes early
+                // termination of string. To avoid that, use strlen(buffer) to determine
+                // the actual string length first and append only this length as supposed.
+                std::size_t append_size = col.ctype_ == SQL_C_BINARY ? buffer_size : strlen(buffer);
+                if (append_size == 0) {
+                    break;
+                }
                 if (ValueLenOrInd == SQL_NO_TOTAL)
-                    out.append(buffer, col.ctype_ == SQL_C_BINARY ? buffer_size : buffer_size - 1);
+                    out.append(buffer, append_size);
                 else if (ValueLenOrInd > 0)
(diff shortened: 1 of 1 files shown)


renkun-ken:
@jimhester I'm not sure if it is appropriate to file PR on `nanodbc` locally. If this looks good for you, I'm also asking if we should also apply the same logic to case `SQL_C_WCHAR`?

renkun-ken:
I filed a PR upstream at nanodbc: https://github.com/nanodbc/nanodbc/pull/176 with exactly the same code change.

renkun-ken:
The upstream PR is merged. This one is refined to be consistent with the changed done there.

jimhester:
Could you add a note to NEWS.md detailing the change with the issue number and mentioning your GitHub username.

Otherwise I think it looks good. My main concerns are strlen being potentially expensive and if the ODBC spec requires strings be null terminated (my assumption was that it did not).

renkun-ken:
`NEWS.md` is updated.

It looks like a database/driver-specific problem, as @mloskot suggests. I test with my database and find that the trailing spaces in each text buffer are not at fixed positions, that is, they are probably not from the original text. I change the buffer size from 1024 to 256, there are always trailing spaces in the buffer. It cannot happen if the original text stored in database has interrupting zeros in the middle of the text or otherwise different choice of buffer size would result in different results if null-termination is enforced.

If the database/driver behave correctly, the text buffer should always be ended with a single zero (as the original code assumes).

renkun-ken:
Any more suggestions, @jimhester?

renkun-ken:
I tested with my database, and the performance impact of `strlen()` is very small. It is only noticeable if user is querying a very long table of many text fields with very long texts in each field. For practical tables, its performance impact can be safely ignored.

I also briefly go through the [ODBC specification](https://github.com/Microsoft/ODBC-Specification/blob/master/ODBC%204.0.md#38-variable-length-columns) on variable-length columns. It does not seem to enforce that the driver should fill the whole buffer before returning `SQL_NO_TOTAL` if more data is to be fetched.

I'd like to also file an issue at https://github.com/FreeTDS/freetds to address the issue here.

renkun-ken:
Fixed upstream.

PR #248, opened 2018-12-07
List objects

blairj09:
Add odbcListObjects.PostgreSQL method so that temp tables are not listed in the connections pane. This fix focusses on listing schemas and uses `DBI::dbGetQuery()` to query PostgreSQL schemas.


jimhester:
My worry with going down this path is that we will have to duplicate a bunch of code for every backend. I really would prefer to avoid this if possible. For now I think we should close this issue, but we can reopen and revisit if needed.

Issue #205, opened 2018-08-06
Error connect to MS Access from R32bit and R64 bit


patperu:
Hi,

I'm trying to connect to MS Access from R 3.5.1 32bit and 64bit and getting an error in both cases.

(1) R3.5.1 (r75001)/32bit and RStudio (1.0.153)
      - connection: ok
      - read: ok
      - write: error

(2) R3.5.1 (r75001)/64bit and RStudio (1.2.830)
      - connection: error
      - read: ---
      - write: ---

### Database
MS Access 2010

### Reproducible Example
Example:
```r
library(odbc)
library(DBI)

cs <- "Driver={Microsoft Access Driver (*.mdb, *.accdb)};Dbq=E:/db_test.mdb"
con <- dbConnect(odbc::odbc(), .connection_string = cs)

dbListTables(con)
# [1] "MSysAccessStorage"          "MSysACEs"
# [3] "MSysNameMap"                "MSysNavPaneGroupCategories"
# [5] "MSysNavPaneGroups"          "MSysNavPaneGroupToObjects"
# [7] "MSysNavPaneObjectIDs"       "MSysObjects"
# [9] "MSysQueries"                "MSysRelationships"
# [11] "tbl_a"

z <- dbReadTable(con, "tbl_a")
str(z)

# str(z)
# 'data.frame':	1 obs. of  6 variables:
# $ ID   : int 5
# $ Feld1: chr "A"
# $ Feld2: chr "B"
# $ Feld3: int 123
# $ Feld4: num 12.4
# $ Feld5: POSIXct, format: "2018-12-31"

dbWriteTable(con, "mtcars", mtcars)
# Error in new_result(connection@ptr, statement) :
#  nanodbc/nanodbc.cpp:1344: 42000: [Microsoft][ODBC-Treiber für Microsoft Access] Syntaxfehler in CREATE TABLE-Anweisung.

```

<details>
<summary>Session Info</summary>

```r
devtools::session_info()
# Session info ----------------------------------------------------------------
  # setting  value
# version  R version 3.5.1 Patched (2018-07-21 r75001)
# system   i386, mingw32
# ui       RStudio (1.0.153)
# language (EN)
# collate  German_Germany.1252
# tz       Europe/Berlin
# date     2018-08-04

# Packages --------------------------------------------------------------------
#  package   * version date       source
# base      * 3.5.1   2018-07-24 local
# bit         1.1-14  2018-05-29 CRAN (R 3.5.0)
[...]

Issue #79, opened 2017-06-23
dbWriteTable() throwing an "Invalid precision value" error with Access


DavisVaughan:
### Issue Description and Expected Result

`dbWriteTable()` throwing an "Invalid precision value" error when used with Access.

Below I've given an example showing what I've tried/learned:
1) Both characters and numerics throw this error.
2) The error is stemming from `result_insert_dataframe`
3) The data types `DOUBLE` and `VARCHAR` seem to be correct, at least to create the table. Creating a new `odbcDataType.ACCESS` throws a different error that looks like it couldn't create the table with the other `SQL_DOUBLE` and `SQL_VARCHAR` data types I supplied.

### Database

Microsoft Access Version 14.0. 

<details>
<summary>Connection from odbc</summary>

``` r
library(DBI)
cn <- dbConnect(odbc::odbc(), dsn = "access-odbc")
dbGetInfo(cn)
#> $dbname
#> [1] "R:\\Life_Modeling_Team\\Vaughan\\1q16_NewBusiness_since_1q15.accdb"
#> 
#> $dbms.name
#> [1] "ACCESS"
#> 
#> $db.version
#> [1] "14.00.0000"
#> 
#> $username
#> [1] "admin"
#> 
#> $host
#> [1] ""
#> 
#> $port
#> [1] ""
#> 
#> $sourcename
#> [1] "access-odbc"
#> 
#> $servername
#> [1] "ACCESS"
#> 
#> $drivername
#> [1] "ACEODBC.DLL"
#> 
#> $odbc.version
#> [1] "03.80.0000"
#> 
#> $driver.version
#> [1] "Microsoft Access database engine"
#> 
#> $odbcdriver.version
#> [1] "03.51"
#> 
#> $supports.transactions
#> [1] FALSE
#> 
#> attr(,"class")
#> [1] "ACCESS"      "driver_info" "list"
```

</details>

### Other details

I am on a Windows computer running Windows 7 Professional.

### Reproducible Example

<details>

``` r
# Connecting (works, and gives warning)
library(DBI)
cn <- dbConnect(odbc::odbc(), dsn = "access-odbc")

# A simple data frame (One has a numeric, one has a character)
df_numeric <- data.frame(col_num  = 3)
df_char    <- data.frame(col_char = "hi")

# Write it. Uh oh. Both fail with the same error

dbWriteTable(cn, "table_num", df_numeric)
[...]

PR #262, opened 2019-03-17
Fix  "Invalid precision value" error when writing tables to Access 

vh-d:
This seems to solve part of the issues with writing to a MS Access database (Issues #79 and #205).


jimhester:
Thanks for the PR

Issue #263, opened 2019-03-17
Support for ODBC drivers that do not allow binding array of parameters (e.g. MS Access)


vh-d:
### Issue Description and Expected Result
Even after fixing #79, we still cannot write data into databases when its ODBC driver does not support binding array of parameters.

### Database
MS Access

### Reproducible Example
Example:
```r
library(odbc)
library(DBI)
con <- dbConnect(odbc::odbc(), dsn = "foobardb")
dbWriteTable(con, "iris", iris)
```

```
 Error in result_insert_dataframe(rs@ptr, values, max_batch_size) : 
  nanodbc/nanodbc.cpp:1581: HYC00: [Microsoft][ODBC Microsoft Access Driver]Optional feature not implemented 
```

<details>
<summary>Session Info</summary>

```r
> devtools::session_info()
Session info ---------------------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.5.2 (2018-12-20)
 os       Windows 7 x64 SP 1          
 system   x86_64, mingw32             
 ui       RStudio                     
 language (EN)                        
 collate  English_United States.1252  
 ctype    English_United States.1252  
 tz       Europe/Prague               
 date     2019-03-17     
```
</details>

jimhester:
I think this may now be resolved by https://github.com/r-dbi/odbc/pull/313, but I do not have an access database available to verify it.

vh-d:
Passing `batch_rows = 1` to dbWriteTable (or using `options(odbc.batch_rows = 1)`) introduced in ed2bf38 works for me. Thanks!

ahjames11:
I think there is one last nuance in that `dbAppendTable()` will not pass through the `batch_rows` parameter.

There are two very usable other options, but it is good to be aware of. Happy to provide any more information or testing if needed.

```r
# This works
dbWriteTable(con, "tbl_Test", data, batch_rows = 1, append = TRUE)

# This does not work
dbAppendTable(con, "tbl_Test", data, batch_rows = 1)
# Error in result_bind(res@ptr, as.list(params), batch_rows) : 
  nanodbc/nanodbc.cpp:1611: HYC00: [Microsoft][ODBC Microsoft Access Driver]Optional feature not implemented 

# This works
options(odbc.batch_rows = 1)
dbAppendTable(con, "tbl_Test", data)
```

This is on dev version 1.2.1.9000.

PR #264, opened 2019-03-17
dtWriteTable support for ODBC drivers that do not allow binding array of parameters

vh-d:
Possible solution to #263 is to allow user set the batch_size parameter to 1 (and make it default for explicit list of drivers, such as Access).

Another possible solution is to check for support for arrays at connection to ODBC. That would require altering nanodbc.cpp though.


jimhester:
I added support for specifying the batch size in https://github.com/r-dbi/odbc/commit/ed2bf38d285c9ca3512c4c282cc6337addf2095d, so that should allow you to use a workaround for Access.

PR #285, opened 2019-06-17
Fix the case-sensitivity of dbExistsTable() 

shrektan:
Fix the case-sensitivity of dbExistsTable()  should be the same as other methods.

Let's say we have a table called "test", before this PR: 

```r
DBI::dbExistsTable(conn, 'table_test')
# [1] TRUE
DBI::dbExistsTable(conn, 'TABLE_TEST')
# [1] FALSE =========> it reports TABLE_TEST doesn't exist
DBI::dbReadTable(conn, 'table_test')
# [1] col1
# <0 rows> (or 0-length row.names)
DBI::dbReadTable(conn, 'TABLE_TEST')
# [1] col1 =============> we can use TABLE_TEST to fetch the table
# <0 rows> (or 0-length row.names)
```

After the PR: `DBI::dbExistsTable(conn, 'TABLE_TEST')` will return `TRUE`.

Moreover, since `DBI::dbWriteTable()` uses `DBI::dbExistsTable()`, before the PR when we tries to append data to "TABLE_TEST", odbc will try to create a new table "TABLE_TEST" and result in failure, due to the fact that "table_test" exists. After this PR, it will be fixed, too. 

I believe this behavior is more reasonable. 

Thanks!


R/Connection.R
```diff
@@ -219,7 +219,8 @@ setMethod(
   "dbExistsTable", c("OdbcConnection", "character"),
   function(conn, name, ...) {
     stopifnot(length(name) == 1)
-    name %in% dbListTables(conn, ...)
+    df <- connection_sql_tables(conn@ptr, table_name = name)
+    nrow(df) > 0
```
jimhester: Could we use `NROW(df)` here just to be sure we will get a numeric value?

I am pretty sure that `connection_sql_tables()` will return an empty data.frame if there are no results, but using `NROW()` will still work if for some reason it instead returns `NULL`.

shrektan: I've made the change.

shrektan:
Great, it's reasonable.

jimhester:
Great work, this package is now better thanks to you!

Thanks!! You are da bomb!

PR #294, opened 2019-08-26
new param timezone_out to control the timezone returned to R

### LINKED ISSUE #275: [FR] be able to control the returned timezone
Hi, first of all, my timezone and my database server's timezone are all Asia/Shanghai (UTC+8).

It's great that we can set the server's timezone info in the odbc connection, a.k.a., `odbc::dbConnect(...,, timezone = 'Asia/Shanghai')`. 

It's also a necessary step to ensure the value uploaded to the server is the same as local. Otherwise, the time will be converted to UTC first before getting uploaded to the server, ending up with a different time value in the server. Moreover, the timezone info get from `DBI::dbGetQuery()` would be UTC which is incorrect either.

Howerver, it causes inconvenient. **The results returned from `DBI::dbGetQuery()` are all converted to UTC timezone and it confuses the users a lot.**

For example, let's say there's a time value `2019-01-01 01:00:00` in the server. The server's timezone is `Asia/Shanghai`. What we get from `DBI::dbGetQuery()` is `2018-12-31 17:00:00 UTC`.

Yes, I know the value it's correct but it's much better if we can simply have the default `Asia/Shanghai` timezone value, a.k.a., `2019-01-01 01:00:00`, by setting somewhere before in the `odbc::dbConnect()`.

**So my proposal is, can we add a new param `timezone_out` in `odbc::dbConnect()` so that we can control the timezone returned from `DBI::dbGetQuery()`?**

Many thanks!

shrektan:
Closes #275 

This PR allows user to define a new param `timezone_out` so that the timezone of the returned object can be local timezone or whatever other than UTC. 

I believe it is very convinient for many users because usually the timezones on the local machine and the server are the same. By forcing the timezone to be UTC leads to sort-of confusing for users, especially those who are not familiar with timezone things.

## An Example

Note this result now displays the same as what's in SQL. But without this functionality, the user will get "2018-12-31 20:00:00 UTC" and have to convert it to "Asia/Shanghai" later by themselves, which is verbose and inconvinient.

``` r
con <- DBI::dbConnect(
  odbc::odbc(),
  server = '#hide#', database = '#hide#', uid = '#hide#', pwd = '#hide#',
  encoding = 'GB2312', driver = 'SQL Server', port = #hide#,
  timezone = 'Asia/Shanghai', timezone_out = 'Asia/Shanghai'
)
out <- DBI::dbGetQuery(con, "select cast('2019-01-01 04:00:00' as datetime) as datetime")
str(out)
#> 'data.frame':    1 obs. of  1 variable:
#>  $ datetime: POSIXct, format: "2019-01-01 04:00:00"
attributes(out$datetime)  
#> $class
#> [1] "POSIXct" "POSIXt" 
#> 
#> $tzone
#> [1] "Asia/Shanghai"
DBI::dbDisconnect(con)
```

<sup>Created on 2019-08-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/src/odbc_result.cpp
+++ b/src/odbc_result.cpp
@@ -416,7 +416,7 @@ double odbc_result::as_double(nanodbc::timestamp const& ts) {
 
 double odbc_result::as_double(nanodbc::date const& dt) {
   using namespace cctz;
-  auto sec = convert(civil_day(dt.year, dt.month, dt.day), c_->timezone());
+  auto sec = convert(civil_day(dt.year, dt.month, dt.day), c_->timezone_out());
   return sec.time_since_epoch().count();
 }
 
@@ -484,7 +484,7 @@ void odbc_result::add_classes(
       break;
     case datetime_t:
       x.attr("class") = Rcpp::CharacterVector::create("POSIXct", "POSIXt");
-      x.attr("tzone") = Rcpp::CharacterVector::create("UTC");
+      x.attr("tzone") = Rcpp::CharacterVector::create(c_->timezone_out_str());
       break;
     case odbc::time_t:
       x.attr("class") = Rcpp::CharacterVector::create("hms", "difftime");
--- a/src/odbc_result.cpp
+++ b/src/odbc_result.cpp
@@ -416,7 +416,7 @@ double odbc_result::as_double(nanodbc::timestamp const& ts) {
 
 double odbc_result::as_double(nanodbc::date const& dt) {
   using namespace cctz;
-  auto sec = convert(civil_day(dt.year, dt.month, dt.day), c_->timezone_out());
+  auto sec = convert(civil_day(dt.year, dt.month, dt.day), c_->timezone());
   return sec.time_since_epoch().count();
 }
 
--- a/src/odbc_result.cpp
+++ b/src/odbc_result.cpp
@@ -416,7 +416,8 @@ double odbc_result::as_double(nanodbc::timestamp const& ts) {
 
 double odbc_result::as_double(nanodbc::date const& dt) {
   using namespace cctz;
-  auto sec = convert(civil_day(dt.year, dt.month, dt.day), cctz::utc_time_zone());
+  auto sec =
+    convert(civil_day(dt.year, dt.month, dt.day), cctz::utc_time_zone());
   return sec.time_since_epoch().count();
 }
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@
 * The RStudio Connections Pane now shows the DSN, when available (#304, @davidchall).
 
 * `dbConnect()` now has a new param `timezone_out` which is useful if the user wants 
-  the datetime values be marked as a specific timezone instead of UTC (@shrektan, #294).
+  the datetime values be marked with a specific timezone instead of UTC (@shrektan, #294).
 
 ## Bugfixes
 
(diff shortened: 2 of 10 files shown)


src/odbc_result.cpp
```diff
@@ -416,7 +416,7 @@ double odbc_result::as_double(nanodbc::timestamp const& ts) {
 
 double odbc_result::as_double(nanodbc::date const& dt) {
   using namespace cctz;
-  auto sec = convert(civil_day(dt.year, dt.month, dt.day), c_->timezone());
+  auto sec = convert(civil_day(dt.year, dt.month, dt.day), c_->timezone_out());
```
krlmlr: Is this still correct after #295?

shrektan: I think I misunderstood how the double value stored in POSIXct class. I thought the value would be varied depending on the timezone but it's not. It's a fixed value compared to UTC 1970-1-1 00:00:00. 

> Class "POSIXct" represents the (signed) number of seconds since the beginning of 1970 (in the UTC time zone) as a numeric vector. 

So what should be done is: after fetching from DB, we just assign the timezone attributes that the user wants. There's no need to change the value of the underlying.

In conclusion, this line is not used at all (the function is about `nanodbc::date` and will never be called for a datetime value) and we actually don't need `c_->timezone_out()`. Only `c_->timezone_out_str()` will be used. 

I will file the correction later.

shrektan:
@krlmlr It should be correct now!

``` r
db_tz <- 'Asia/Shanghai' 
user_tz <- 'America/Los_Angeles'
con <- DBI::dbConnect(
  odbc::odbc(),
  server = '127.0.0.1', database = 'tempdb', uid = 'sa', pwd = 'Abc_123_Bcd',
  encoding = 'GB2312', driver = 'ODBC Driver 13 for SQL Server', port = 1433,
  timezone = db_tz, timezone_out = user_tz
)
out <- DBI::dbGetQuery(con, "select cast('2019-01-01 04:00:00' as datetime) as datetime, 
                       cast('2019-01-01' as date) as date")
str(out)
#> 'data.frame':    1 obs. of  2 variables:
#>  $ datetime: POSIXct, format: "2018-12-31 12:00:00"
#>  $ date    : Date, format: "2019-01-01"
attributes(out$datetime)
#> $class
#> [1] "POSIXct" "POSIXt" 
#> 
#> $tzone
#> [1] "America/Los_Angeles"
lubridate::with_tz(out$datetime, db_tz)
#> [1] "2019-01-01 04:00:00 CST"
print(out$date)
#> [1] "2019-01-01"
DBI::dbDisconnect(con)
```

<sup>Created on 2019-10-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

shrektan:
The failed check is only happened on R-devel and is caused by a warning _"checking top-level files ... WARNING"_, which seems not related to this PR.

Moreover, just to note that now it's possible to have an MS SQL Server inside of a docker container, making the tests more easily. See [quickstart-install-connect-docker](https://docs.microsoft.com/en-us/sql/linux/quickstart-install-connect-docker?view=sql-server-2017&pivots=cs1-bash) for details. 😃

jimhester:
Nice, that is useful!

jimhester:
Awesome, great work!

Thanks a bunch!

PR #304, opened 2019-10-01
Show DSN in RStudio Connections Pane (when available)

davidchall:
The RStudio Connections Pane uses a `displayName` to summarize each connection.

Currently, the odbc package shows something like `dbname - username@hostname via TCP/IP`, which can get quite long (spilling over multiple lines).

When available, the Data Source Name provides a (probably) shorter string that still uniquely identifies the database connection. It's also highly likely that the DSN is easier for the user to interpret, because it was expressly designed for this purpose.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+* The RStudio Connections Pane now shows the DSN, when available (#304, @davidchall)
+
 ## Bugfixes
 
 * Fixed an issue where `DBI::dbListFields()` could fail when used with a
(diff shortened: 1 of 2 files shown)


nwstephens:
In my personal experience, DSN's were commonly known (i.e. they were not private). As @davidchall points out, the DSN's are typically created so that it's easy for users to refer to them. I'd be a little surprised if people had "secret" DSN's.

jimhester:
Thanks @davidchall! Great work, this package is now better thanks to you!

Thanks!! You are da bomb!

PR #307, opened 2019-10-13
Use setDBIMethod()

krlmlr:
I'd like to rename the `con` argument to `sqlData()` to `conn` in r-dbi/DBI#285. This change allows me to do so in the future without introducing a breaking change which would be very unpleasant to deal with. Also, it opens up the potential to clean up the definition of the DBI interface.

See https://dbi.r-dbi.org/dev/reference/setdbimethod for details.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ LazyData: true
 ByteCompile: true
 Depends: R (>= 3.2.0)
 Imports:
-    DBI (>= 1.0.0),
+    DBI (>= 1.0.0.9003),
     methods,
     Rcpp (>= 0.12.11),
     blob (>= 1.1.0),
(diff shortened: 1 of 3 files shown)


jimhester:
I don't think we should do this, it would be better to deal with the breaking changes than to require all DBI packages to override such a foundational S4 function. 

Also if you are going to do this it likely needs to be done at load time rather than install time, overwise if DBI is updated after the client package you could still be using an older `DBI::setDBIMethod` definition in the client with newer code in DBI.

krlmlr:
AFAICT the `setMethod()` calls are (re)evaluated at load time. We can double-check what happens.

Anyway, I agree that a more "classical" approach (along the lines of `if (packageVersion("DBI") >= "...") ...`) is less risky. Let me draft something for RSQLite.

PR #308, opened 2019-10-14
Use recent DBItest

krlmlr:
Skip the tests that currently fail.

In other backends, I have turned off DBItest tests on CRAN to avoid the pain when updating.


jimhester:
Skipping on CRAN is one of the things that `skip_unless_has_test_db()` does https://github.com/r-dbi/odbc/blob/ba6588a548ed6ad0b62cac986db4c6fd15e5713d/tests/testthat/helper.R#L2-L4

jimhester:
Thanks for starting on this Kirill!

Issue #70, opened 2017-06-14
Provide fallback implementation to use when parameter binding fails


jimhester:
Some databases / drivers have issues with the way nanodbc does parameter binding for inserting data. https://github.com/rstats-db/odbc/issues/66, https://github.com/rstats-db/odbc/issues/30. To work around the issue we can provide a fallback `odbcWriteTableSQL()` function that writes the values as text rather than binding a column. This requires escaping the data specially for things like Dates, datetimes and missing data.

jimhester:
We are providing a VARCHAR fallback now, which seems to avoid the issue without having to do the proposal here.

detule:
@jimhester would you be interested in a back-port of https://github.com/nanodbc/nanodbc/commit/e99606bc8d01257c7c1316dd06f3a9c30e0a71fd (happy to do it)?

PR #313, opened 2019-10-30
Feature/describe prior to write

detule:
Hi @jimhester @krlmlr 

Please consider this patch-set as potentially a more robust alternative to https://github.com/r-dbi/odbc/issues/70.

* With the current solution (master) FreeTDS users face truncation when inserting values longer than 255 characters.
* An additional benefit could be performance enhancement when writing to wide tables, as this approach makes less round-trip calls to the server.

Looking forward to hearing your feedback.


jimhester:
Thanks for working on this!

It looks like this change is causing the tests to fail when using PostgreSQL, would you be able to look into this?

detule:
Will do - thanks.

detule:
@jimhester Should be ready for a second look.

The R-devel pipeilne failure seems to be [unrelated to this patch-set](https://travis-ci.community/t/new-warning-in-r-devel-about-the-checkbashisms-script/5551).

R/Table.R
```diff
@@ -62,6 +71,9 @@ odbc_write_table <-
         )
       rs <- OdbcResult(conn, sql)
 
+      if (!is.null(fieldDetails) && nrow(fieldDetails))
+        result_describe_parameters(rs@ptr, fieldDetails)
```
jimhester: Can you put braces around this conditional?

R/Table.R
```diff
@@ -50,6 +50,15 @@ odbc_write_table <-
       dbExecute(conn, sql)
     }
 
+    fieldDetails <-
+      tryCatch({
+        details <- odbcConnectionColumns(conn, name)
+        datails <- details[match(names(values), details$column_name)]
+        details[, c("ordinal_position", "data_type", "column_size", "decimal_digits")]
```
jimhester: This call has no effect, did you mean to assign it back to details?

R/Table.R
```diff
@@ -50,6 +50,15 @@ odbc_write_table <-
       dbExecute(conn, sql)
     }
 
+    fieldDetails <-
+      tryCatch({
+        details <- odbcConnectionColumns(conn, name)
+        datails <- details[match(names(values), details$column_name)]
```
jimhester: This could just be

```r
datails <- details[names(values)]
```

But do we need to reorder the columns at all?

R/Connection.R
```diff
@@ -69,6 +69,101 @@ setClass(
...
+        schema_name = schema_name,
+        column_name = column_name)
+    # Rename/re-order columns to make output ODBC-like
+    names(detail) <- c("column_name", "type_name", "table_name", "table_schem",
+      "table_cat", "data_type", "column_size", "buffer_length", "decimal_digits",
+      "numeric_precision_radix", "remarks", "column_def", "sql_data_type",
+      "sql_datetime_subtype", "char_octet_length", "ordinal_position",
+      "nullable")
+    detail <-
+      detail[, c(5, 4, 3, 1, 6, 2, 7, 8, 9, 10, 17, 11, 12, 13, 14, 15, 16)]
```
jimhester: Indexing by number seems too brittle, could we reorder by name first, then rename the columns if necessary.

R/Connection.R
```diff
@@ -69,6 +69,101 @@ setClass(
...
+
+    if (!is.null(column_name))
+      detail <- subset(detail, detail$column_name == column_name)
```
jimhester: Could we add braces here,

Also generally it is best to avoid using subset inside functions, the type of non-standard evaluation in `subset()` is not robust enough.

`detail <- detail[detail$column_name == column_name, ]` is the equivalent.

Also it is somewhat surprising that you would need to do an additional filtering, if you search for a specific column in `SQLColumns` only that column should be returned.

R/Table.R
detule: Jim - the output of the `tryCatch` block is assigned to `fieldDetails` -  the line you quoted there is what gets assigned in the success code-path.

Let me know if you think there is a better approach.

R/Connection.R
detule: Will change - thanks.

R/Connection.R
detule: Will change - thanks.

R/Table.R
detule: Will do - thanks.

R/Table.R
detule: Jim, the intention here was to:
* Subset the rows of the output of SQLColumns to those (columns) we are writing to / we need to describe.
* I thought some care was needed with the ordering - AFAICT there is no guarantee that the columns in `values` (and therefore the parameters in the INSERT statement) are ordered the same way as the columns in the table we are writing to (i.e. the rows in `details`).

Having said that, I think part of the confusion is the use of `ordinal_position` to index the parameters we are describing - this seems incorrect.  After re-ordering the rows in `details` to match the column order in `values` as is done here, the index that gets passed `nanodbc::statement::describe_parameters` should be `1:ncol(values)`

detule:
@jimhester Thanks for the feedback!

* Patched conditional braces, numeric column indexing
* Fixed bug with pairing table column descriptions with INSERT statement parameters.  Added a test.

`test_roundtrip` is perhaps not the best home for this test - happy to move it to a more appropriate location if you have any suggestions.

R/Connection.R
```diff
@@ -69,6 +69,98 @@ setClass(
...
+        catalog_name = catalog_name,
+        schema_name = schema_name,
+        column_name = column_name)
+    # Re-order/rename columns to make output ODBC-like
+    detail <- detail[, c("catalog_name", "schema_name", "table_name", "name",
+      "data_type", "field.type", "column_size", "buffer_length",
+      "decimal_digits", "numeric_precision_radix", "nullable", "remarks",
+      "column_default", "sql_data_type", "sql_datetime_subtype",
+      "char_octet_length", "ordinal_position")]
+    names(detail)[c(1, 2, 4, 6, 13)] <- c("table_cat", "table_schem",
```
jimhester: I think I would prefer to _not_ rename. And do we need to reorder the columns?

R/Table.R
jimhester: It was due to the indentation style used, I would normally put the `tryCatch()` line on the same line as the assignment.

jimhester:
I would prefer a separate test outside of the roundtrip that creates a table and appends to it.

R/Table.R
detule: Will update - thanks.

R/Connection.R
detule: Jim my original thinking here was to make the output as close to the output of the ODBC `SQLColumns` function as possible.   It seemed to me that, since we are offering the end-user the ability to write their own implementation, then it is more natural to anchor on the field naming of the ODBC's `SQLColumns` API, than on the choices made when implementing the internal `connection_sql_columns`.

If you also think that's worthwhile then we either re-name the columns here, or change the output of `connection_sql_columns`.  The latter seemed less desirable since even though internal, I wasn not sure if there aren't folks that have developed dependencies.

The re-ordering was more thinking in the same direction - making the output align with what I can see when I execute `sp_columns` in my SQL Server client for example.

I don't feel strongly here - happy to make this just a wrapper around `connection_sql_columns` if you think that's the best way forward.

jimhester: I would prefer it to just be a wrapper around `connection_sql_columns`

detule:
@jimhester Thanks again - appreciate the comments.

Let me know if I missed anything.

jimhester:
Thanks!! You are da bomb!

detule:
Jim: thank you very much - I know it was a chunky PR, and I appreciate the thoughtful comments.

Any thoughts on when this might make its way to CRAN?

jimhester:
I was aiming for a CRAN release on odbc next week sometime.

PR #320, opened 2019-11-25
fix a missing comma in odbc_write_table()

### LINKED ISSUE #319: after #313 dbWriteTable() is case-sensitive in terms of columns
After PR #313 (7c4bc6ec) being merged, the `dbWriteTable()` method is now case-sensitive in terms of matching columns agains the db table. But before this PR, it's case insensitive (at least for MSSQLServer). This breaks many of my R scripts...

The database version is:

```
Microsoft SQL Server 2014 - 12.0.2000.8 (X64) 
	Feb 20 2014 20:04:26 
	Copyright (c) Microsoft Corporation
	Enterprise Edition (64-bit) on Windows NT 6.2 <X64> (Build 9200: )
```

## The example code

(Before the PR, it runs successfully without errors. )

``` r
conn <- DBI::dbConnect(...)
DBI::dbExecute(conn, "create table test (aaa int) ")
#> [1] 0
DBI::dbWriteTable(conn, 'test', value = data.frame(AAA = 1L), append = TRUE)
### I have to change `data.frame(AAA = 1L)` to ` data.frame(aaa = 1L)`. ###
#> Error in result_describe_parameters(rs@ptr, fieldDetails): Index out of bounds: [index='param_index'].
DBI::dbExecute(conn, "drop table test")
#> Warning in new_result(connection@ptr, statement): Cancelling previous query
#> [1] 0
DBI::dbDisconnect(conn)
```

shrektan:
Fixes #319

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -62,9 +62,6 @@
 * `dbWriteTable()` now always writes `NA_character` as `NULL` for
   data.frame with only one row (@shrektan, #288).
 
-* `dbWriteTable()` now is case-insensitive as before when matching agains 
-  the database's table fields (@shrektan, #320).
-
 # odbc 1.1.6
 
 ## Features
--- a/NEWS.md
+++ b/NEWS.md
@@ -62,6 +62,9 @@
 * `dbWriteTable()` now always writes `NA_character` as `NULL` for
   data.frame with only one row (@shrektan, #288).
 
+* `dbWriteTable()` now is case-insensitive as before when matching agains 
+  the database's table fields (@shrektan, #320).
+
 # odbc 1.1.6
 
 ## Features
--- a/R/Table.R
+++ b/R/Table.R
@@ -55,7 +55,7 @@ odbc_write_table <-
     fieldDetails <- tryCatch({
       details <- odbcConnectionColumns(conn, name)
       details$param_index <- match(details$name, names(values))
-      details[!is.na(details$param_index)]
+      details[!is.na(details$param_index), ]
     }, error = function(e) {
       return(NULL)
     })
--- a/R/Table.R
+++ b/R/Table.R
@@ -54,9 +54,6 @@ odbc_write_table <-
 
     fieldDetails <- tryCatch({
       details <- odbcConnectionColumns(conn, name)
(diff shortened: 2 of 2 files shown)


jimhester:
Whether case sensitivity matters is database dependent, some databases it doesn't matter, some it does and in some it matters dependant on a configuration parameter. We can't assume case insensitivity in the odbc package.

jimhester:
But maybe we have to do this, I am not sure why we are getting case mismatches when describing the columns...

detule:
The first commit is a bugix (sorry!).  

Presumably with that commit only, if there is a case mismatch between the columns of `values` and the target table the fallback should continue to work gracefully.

@shrektan does that work?

detule:
Having said that, the entirety of the PR, including the fuzzy (case) matching on the [R] side is probably a good idea.

Nice work.

shrektan:
I think we may need to provide the column name argument when calling odbcConnectionColumns(conn, name) and let the databases verify if the fields provided can be matched or not. I will tweak and verify my “theory” later.

shrektan:
@detule You are right. Fix `details[!is.na(details$param_index), ]` only is enough for this case, in fact. 

Regarding to fieldDetails, the only robust way I can think of is to call odbcConnectionColumns() and let the server to find the correct columns. 

However, this approach has a drawback that `catalog::columns catalog::find_columns()` can only handle scalar column_name, so we have to call the function multiple times, which may result in some performance overheads when there're many fields in the table (). The code will be like this:

```r
      # the cpp function connection_sql_columns() now is only able to handle
      # scalar column_name.
      details <- lapply(names(values), function(col) {
        odbcConnectionColumns(conn, name, column_name = col)
      })
      # use the first match for each field and set up the param_index
      details <- lapply(seq_len(ncol(values)), function(i) {
        if (nrow(details[[i]])) details[[i]][1L,]$param_index <- i
      })
      do.call(rbind, details)
```

So for this PR, I will just fix the missing comma `,`, and let you and @jimhester to decide if need further improvement or not. 

Thanks.

jimhester:
Ok, I am fine with just doing the comma for now, thanks again for the PR @shrektan

PR #338, opened 2020-01-07
Use correct length for varchar columns

krlmlr:
on SQL Server.

Before:

``` r
odbc:::`odbcDataType.Microsoft SQL Server`(NULL, character())
#> Warning in max(nchar(as.character(x)), na.rm = TRUE): no non-missing arguments
#> to max; returning -Inf
#> [1] "varchar(255)"

odbc:::`odbcDataType.Microsoft SQL Server`(NULL, letters)
#> [1] "varchar(255)"
```

<sup>Created on 2020-01-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

After:

``` r
odbc:::`odbcDataType.Microsoft SQL Server`(NULL, character())
#> [1] "varchar(1)"

odbc:::`odbcDataType.Microsoft SQL Server`(NULL, letters)
#> [1] "varchar(1)"
```

<sup>Created on 2020-01-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/DataTypes.R
+++ b/R/DataTypes.R
@@ -316,13 +316,25 @@ object_type <- function(obj) {
 }
 
 varchar <- function(x, type = "varchar") {
-  max_length <- max(c(1, nchar(as.character(x))), na.rm = TRUE)
-  paste0(type, "(", min(255, max_length), ")")
+  # at least 255 characters, use max if more than 8000:
+  max_length <- max(c(255, nchar(as.character(x))), na.rm = TRUE)
+
+  if (max_length > 8000) {
+    max_length <- "max"
+  }
+
+  paste0(type, "(", max_length, ")")
 }
 
 varbinary <- function(x, type = "varbinary") {
-  max_length <- max(c(1, lengths(x)), na.rm = TRUE)
-  paste0(type, "(", min(255, max_length), ")")
+  # at least 255 bytes, use max if more than 8000:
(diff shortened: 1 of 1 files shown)


jimhester:
This is not correct, the intention was to use varchar(255) _unless_ the maximum length of one of the strings was greater than 255, in which case we would have to use that instead.

Dealing with the 0 length vector case is fine I guess though...

krlmlr:
Interesting. The limit is 8000 though, beyond that we need `max`: https://docs.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver15.

I'll tweak.

krlmlr:
``` r
odbc:::`odbcDataType.Microsoft SQL Server`(NULL, character())
#> [1] "varchar(255)"

odbc:::`odbcDataType.Microsoft SQL Server`(NULL, letters)
#> [1] "varchar(255)"

odbc:::`odbcDataType.Microsoft SQL Server`(NULL, paste(rep("a", 10000), collapse = ""))
#> [1] "varchar(max)"
```

<sup>Created on 2020-01-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jimhester:
Thanks!

PR #340, opened 2020-01-10
wrap table name in toString() when throwing errors from odbc_write_table

### LINKED ISSUE #339: DBI::dbWriteTable() fails with wrong error message if table exists and table name is a DB::Id
`DBI::writeTable()` fails with the not very helpful error message `"no method for coercing this S4 class to a vector"` if the target table exists and both append and overwrite are false AND the table name is a `DBI::id()`. I think the problem is this part of  `odbc_write_table()` in Table.R that cannot handle Id objects:

```
  if (found && !overwrite && !append) {
      stop("Table ", name, " exists in database, and both overwrite and",
        " append are FALSE", call. = FALSE)
    }
```

s-fleck:
fixes #339

### DIFF
--- a/R/Table.R
+++ b/R/Table.R
@@ -38,7 +38,7 @@ odbc_write_table <-
 
     found <- dbExistsTable(conn, name)
     if (found && !overwrite && !append) {
-      stop("Table ", name, " exists in database, and both overwrite and",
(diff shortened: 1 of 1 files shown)


jimhester:
Thanks!

PR #359, opened 2020-04-28
WIP: Catch errors from calling SQLGetData()

### LINKED ISSUE #358: Better behavior for "Invalid Descriptor Index" problem
I'm now convinced that the problem must be solved by using an appropriate driver, or putting long columns at the end of the query.

Still: Can we do better than throwing `"Invalid Descriptor Index"` at the user?

``` r
library(DBI)
library(tidyverse)

con <- dbConnect(odbc::odbc(), "mssql-test-ms", uid = "kirill", pwd = keyring::key_get("mssql", "kirill"))

data <- tibble(a = 1:3, b = "a", c = 1)

copy_to(con, data, name = "##data1", types = c("integer", "varchar(max)", "real")) %>% collect()
#> Error in result_fetch(res@ptr, n): nanodbc/nanodbc.cpp:2908: 00000: [Microsoft][ODBC Driver 17 for SQL Server]Invalid Descriptor Index
#> Warning in dbClearResult(res): Result already cleared
```

<sup>Created on 2020-04-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Follow-up to #309.

krlmlr:
This is a proof of concept to catch errors from drivers that don't implement the `SQL_GD_ANY_COLUMN` extension. It shows that we can query more columns and rows even after `SQLGetData()` fails.

We could then:

- distinguish 07009 errors from others, https://docs.microsoft.com/en-us/sql/odbc/reference/appendixes/appendix-a-odbc-error-codes
- record all failures
- raise a warning that indicates which columns could not be retrieved
- suggest remedies in that warning, and point to appropriate resources
    - we can't rely on `SQLGetInfo()` for all drivers, but in case of a failure we can check if the `SQL_GD_ANY_COLUMN` extension is supported and warn appropriately)

Closes #358.

``` r
library(DBI)
library(tidyverse)

con <- dbConnect(odbc::odbc(), "mssql-test-ms", uid = "kirill", pwd = keyring::key_get("mssql", "kirill"))

data <- tibble(a = 1:3, b = "a", c = 1)

copy_to(con, data, name = "##data1", types = c("integer", "varchar(max)", "real")) %>% collect()
#> # A tibble: 3 x 3
#>       a b         c
#>   <int> <chr> <dbl>
#> 1     1 <NA>      1
#> 2     2 <NA>      1
#> 3     3 <NA>      1
```

<sup>Created on 2020-04-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/src/odbc_result.cpp
+++ b/src/odbc_result.cpp
@@ -770,15 +770,19 @@ void odbc_result::assign_string(
     // SQLGetData()
     // has been called for that column (i.e. after get() or get_ref() is
     // called).
-    auto str = value.get<std::string>(column);
-    if (value.is_null(column)) {
-      res = NA_STRING;
-    } else {
-      if (c_->encoding() != "") {
-        res = output_encoder_.makeSEXP(str.c_str(), str.c_str() + str.length());
-      } else { // If no encoding specified assume it is UTF-8 / ASCII
-        res = Rf_mkCharCE(str.c_str(), CE_UTF8);
+    try {
+      auto str = value.get<std::string>(column);
+      if (value.is_null(column)) {
+        res = NA_STRING;
+      } else {
+        if (c_->encoding() != "") {
+          res = output_encoder_.makeSEXP(str.c_str(), str.c_str() + str.length());
+        } else { // If no encoding specified assume it is UTF-8 / ASCII
+          res = Rf_mkCharCE(str.c_str(), CE_UTF8);
+        }
       }
+    } catch (const nanodbc::database_error& e) {
+      res = NA_STRING;
     }
   }
   SET_STRING_ELT(out[column], row, res);


r2evans:
I realize your PR doesn't check for the column/info (yet), but I don't think that self-reporting of `SQL_GD_ANY_COLUMN` will always be sufficient:

```lang-r
con <- DBI::dbConnect(..., driver = "FreeTDS")
DBI::dbExecute(con, "create table r2test (a int, b nvarchar(max), c real)")
# [1] 0
DBI::dbWriteTable(con, "r2test", tibble(a = 1:3, b = "a", c = 1), create = FALSE, append = TRUE)
DBI::dbGetQuery(con, "select * from r2test")
#   a b c
# 1 1 a 1
# 2 2 a 1
# 3 3 a 1
str(DBI::dbGetInfo(con))
# List of 15
#  $ dbname                       : chr "racers"
#  $ dbms.name                    : chr "Microsoft SQL Server"
#  $ db.version                   : chr "13.00.1742"
#  $ username                     : chr ""
#  $ host                         : chr ""
#  $ port                         : chr ""
#  $ sourcename                   : chr ""
#  $ servername                   : chr "FPTC-DATA01"
#  $ drivername                   : chr "libtdsodbc.so"
#  $ odbc.version                 : chr "03.80.0000"
#  $ driver.version               : chr "01.01.9999"
#  $ odbcdriver.version           : chr "03.50"
#  $ supports.transactions        : logi TRUE
#  $ getdata.extensions.any_column: logi FALSE
#  $ getdata.extensions.any_order : logi FALSE
#  - attr(*, "class")= chr [1:3] "Microsoft SQL Server" "driver_info" "list"
DBI::dbDisconnect(con)

con <- DBI::dbConnect(..., driver = "ODBC Driver 17 for SQL Server"), ...)
DBI::dbGetQuery(con, "select * from r2test")
# Error in result_fetch(res@ptr, n) : 
#   nanodbc/nanodbc.cpp:2966: 07009: [Microsoft][ODBC Driver 17 for SQL Server]Invalid Descriptor Index 
# Warning in dbClearResult(rs) : Result already cleared
```

jimhester:
This should be solved by #381

PR #360, opened 2020-04-28
Fix immediate = TRUE for FreeTDS

krlmlr:
For direct execution, opening then closing a connection is redundant.


krlmlr:
All tests pass on SQL Server with FreeTDS driver with this PR.

jimhester:
To clarify for myself, using the statement constructor with a connection opens the connection, and `execute_direct()` first closes the connection if it is opens, so there is no need to pass the connection to the initial statement constructor.

PR #363, opened 2020-05-14
Catch exception by value to avoid OOB access

krlmlr:
I have no idea why this is necessary, nanodbc and Rcpp seem to be doing the right thing otherwise. Should we also catch by value when fetching the result?

Reproducible with:

```
R -d valgrind -q -e 'pkgload::load_all(); DBI::dbConnect(odbc(), list(dsn = "bogus"))'
```


krlmlr:
A clean rebuild solves the problem.

PR #364, opened 2020-05-14
Correct blob roundtrip

krlmlr:
Enables two DBItest tests.

### DIFF
--- a/R/DataTypes.R
+++ b/R/DataTypes.R
@@ -318,8 +318,16 @@ object_type <- function(obj) {
 is_blob <- function(obj) {
   if (is(obj, "blob")) return(TRUE)
   if (is.object(obj) && any(class(obj) != "AsIs")) return(FALSE)
-  # Assuming raw inside naked lists, not checking
-  is.list(obj)
+  if (!is.list(obj)) return(FALSE)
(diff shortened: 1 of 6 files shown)


R/DataTypes.R
```diff
@@ -309,12 +309,19 @@ object_type <- function(obj) {
...
   if (is(obj, "difftime")) return("time")
 
   return(typeof(obj))
 }
 
+is_blob <- function(obj) {
+  if (is(obj, "blob")) return(TRUE)
+  if (is.object(obj) && any(class(obj) != "AsIs")) return(FALSE)
+  # Assuming raw inside naked lists, not checking
+  is.list(obj)
```
jimhester, line 322: Why do we need this assumption? Couldn't we check it with something like `is.list(x) && length(x) > 0 && is.raw(x[[1]])`

krlmlr: Because checking `x[[1]]` is not enough, it may be `NULL`. In the worst case we'd iterate over the entire vector just for checking -- safer but slower. Happy to add that if you prefer.

krlmlr: Added.

PR #379, opened 2020-07-19
add Pacman to configure file

she3o:
Helps users of Archlinux-based distros confirm which package (core/unixodbc) to install from Pacman repositories in case `sql.h` was not present.


jimhester:
Thanks a bunch!

PR #381, opened 2020-07-30
Prcandidate/out of order blob

detule:
Hi @jimhester 

This is one approach to tackling the infamous "invalid descriptor" problem.  The idea is: If driver is such that out-of-order-retrieval via `SQLGetData` is not possible then unbind all `nanodbc` buffers past the first unbound (blob) column.

First two patches are a clean-ish backport of the upstream patches that introduce the 'result::unbind' endpoint - but more importantly, route calls to `get<T>` to `SQLGetData` if there is no bound buffer.  Previously this (no buffer bound to the column) was the case only for blobs, but now it's possible for non-blobby columns if the user asked these to be unbound.

Second two are `package:odbc` patches that implement the "unbind-if-necessary" logic.  `connection::get_data_any_order` is particularly ugly - happy to make updates there and elsewhere.

Thanks

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -11,6 +11,7 @@ Authors@R:
              family = "Wickham",
              role = "aut",
              email = "hadley@rstudio.com"),
+      person(given = "detule", role = "ctb"),
       person(given = "lexicalunit",
              role = "cph",
              comment = "nanodbc library"),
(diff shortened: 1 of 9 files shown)


jimhester:
Hi Oliver, thanks so much for working on this!

Both the upstream patch to nanoodbc and this PR, I apologize for the delay in getting to the review but I think the code looks good as is.

As this is a significant contribution I would like to add you as a co-author to the package, would you prefer I use your full name or just the `detule` alias?

detule:
@jimhester thanks again.  Appreciate being added as a contributor - full name is fine.

jimhester:
Hey! Thanks!

PR #400, opened 2020-10-25
Fix compilation on Alpine Linux

gaborcsardi:
* Pass down `CXXFLAGS` to libcctz. On https://github.com/r-hub/r-minimal
  we set `__MUSL__`. This is not a standard, but nevertheless some
  R packages (e.g. Rcpp) rely on it already, and at least the user
  can set it in a user Makevars file.
* Use `__MUSL__` to select the right `strerror_r` signature. I don't know
  why this is not happening automatically. It might be a libcctz
  bug or a musl bug.


jimhester:
Thanks!

PR #412, opened 2020-12-17
should check validity before clearing results

shrektan:
For certain cases, e.g., `select A/B from Table` where the B column contains zero, the `rs` object would have become invalid already after fetching. This leads to a case that odbc throws warning complaining "Result already cleared".

This PR fixes that.

I think it's really trival so I think without a NEWs item and test should be fine. However, if you think they are necessary, please let me know.

Thanks.

### DIFF
--- a/R/Connection.R
+++ b/R/Connection.R
@@ -352,7 +352,7 @@ setMethod(
 setMethod("dbGetQuery", signature("OdbcConnection", "character"),
   function(conn, statement, n = -1, params = NULL, ...) {
     rs <- dbSendQuery(conn, statement, params = params, ...)
-    on.exit(dbClearResult(rs))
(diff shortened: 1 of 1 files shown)


detule:
Hey @shrektan 

I think there may be some subtlety here.  In https://github.com/r-dbi/odbc/pull/334 for example, I think we tried to make sure that the `result_release` that's part of the `dbClearResult` call, gets called unconditionally (valid result or not) so as to make sure that all `nanodbc` resources get promptly destroyed [ i am not sure, but this in turn may issue some commands to reset the cursor ].

Is the warning (after a bad query) the only issue here?

shrektan:
@detule

Ha, it seems that there's no apparent-better solution for this. Live with the warning is ok for me. So I'm closing this PR for now.

Issue #414, opened 2020-12-18
invalid descriptor index fails with param-binding


r2evans:
### Issue Description and Expected Result

This is a continuation of #358, #309, and several others.

The recent fixes seem to work well with "regular" text, but when we introduce bound parameters, it fails.

### Database

SQL Server 2017, running on docker. Versions:

```r
DBI::dbGetQuery(conss, "select @@version")
#                                                                                                                                                                                                      
# 1 Microsoft SQL Server 2017 (RTM-CU13) (KB4466404) - 14.0.3048.4 (X64) \n\tNov 30 2018 12:57:58 \n\tCopyright (C) 2017 Microsoft Corporation\n\tDeveloper Edition (64-bit) on Linux (Ubuntu 16.04.5 LTS)

str(DBI::dbGetInfo(conss))
# List of 15
#  $ dbname                       : chr "master"
#  $ dbms.name                    : chr "Microsoft SQL Server"
#  $ db.version                   : chr "14.00.3048"
#  $ username                     : chr "dbo"
#  $ host                         : chr ""
#  $ port                         : chr ""
#  $ sourcename                   : chr ""
#  $ servername                   : chr "c755900f6200"
#  $ drivername                   : chr "msodbcsql17.dll"
#  $ odbc.version                 : chr "03.80.0000"
#  $ driver.version               : chr "17.06.0001"
#  $ odbcdriver.version           : chr "03.80"
#  $ supports.transactions        : logi TRUE
#  $ getdata.extensions.any_column: logi FALSE
#  $ getdata.extensions.any_order : logi FALSE
#  - attr(*, "class")= chr [1:3] "Microsoft SQL Server" "driver_info" "list"
```

### Reproducible Example

```r
# docker run -d -p "31433:1433" --name testss -e "ACCEPT_EULA=Y" -e "MSSQL_SA_PASSWORD=Mysecretpassword1" -e "MSSQL_PID=Developer" --cap-add SYS_PTRACE microsoft/mssql-server-linux
library(DBI)
conss <- DBI::dbConnect(drv = odbc::odbc(), driver = "ODBC Driver 17 for SQL Server", 
                        server = "127.0.0.1,31433", database = "master", include_password = TRUE, 
                        uid = "SA", pwd = "Mysecretpassword1")

DBI::dbExecute(conss, "
  create table quux1 (
    id UNIQUEIDENTIFIER DEFAULT NEWSEQUENTIALID() PRIMARY KEY NOT NULL,
[...]

PR #415, opened 2020-12-23
Invalid descriptor with prepared SELECTs

detule:
Hi @jimhester - this moves the `unbind_if_needed` call to be part of the fetch operation - rather than when the the result is constructed.

Fixes https://github.com/r-dbi/odbc/issues/414


jimhester:
Great, thanks again!

PR #417, opened 2020-12-23
tests: sqlserver: fix timestamp roundtrip

detule:
Jim fixing my own mess:

I am not sure how this passed the test in https://github.com/r-dbi/odbc/pull/413 when this test was added.


krlmlr:
I think this broke independently, I saw it before the other PR was merged too.

detule:
Great - I think this PR is still needed as it fixes my derp for the test I wrote couple of days ago.

But sounds like there are new DBItest / SQLite failures, which I think are unrelated from any code I touched.

PR #418, opened 2020-12-24
Fail if database is missing

krlmlr:
if `DBITEST_BACKENDS` environment variable is set.

I'm adding odbc to the downstream tests in https://github.com/r-dbi/DBItest/pull/224, I'd like the tests to fail in this case if no database connection can be established.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -2,10 +2,19 @@ skip_unless_has_test_db <- function(expr) {
   if (!identical(Sys.getenv("NOT_CRAN"), "true")) {
     return(skip("On CRAN"))
   }
-  tryCatch({
-    DBItest:::connect(expr)
+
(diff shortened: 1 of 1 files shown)


tests/testthat/helper.R
```diff
@@ -2,10 +2,19 @@ skip_unless_has_test_db <- function(expr) {
   if (!identical(Sys.getenv("NOT_CRAN"), "true")) {
     return(skip("On CRAN"))
   }
-  tryCatch({
-    DBItest:::connect(expr)
+
+  # Failing database connection should fail the test in DBItest backends
+  if (Sys.getenv("DBITEST_BACKENDS") != "") {
```
shrektan, line 7:
```suggestion
  if (nzchar(Sys.getenv("DBITEST_BACKENDS"))) {
```

krlmlr:
I fixed upstream too so that tests now pass.

PR #424, opened 2020-12-28
Adds Hive subclass and dbQuoteString method for it

rnorberg:
Hive uses C-style escape for string literals, so a single-quote inside a string literal should be replaced with `\'` and not `''` (as is currently the case). See r-dbi#184 for details.


jimhester:
It looks like you are missing documentation for the new methods, which is why the checks are failing. e.g. https://github.com/r-dbi/odbc/pull/424/checks?check_run_id=1619012948#step:10:74

jimhester (approved):

jimhester:
Thanks!

PR #426, opened 2020-12-28
Adds sqlCreateTable method specifically for DB2 connections

rnorberg:
This method correctly creates a DB2 temporary table when `temporary = TRUE`. 

Tested on a DB2 database at my job:

```R
db2_conn <- DBI::dbConnect(...)
class(db2_conn)
# [1] "DB2/AIX64"
DBI::dbWriteTable(db2_conn, DBI::Id(schema = "SESSION", table = "mtcars_tmp"), mtcars, temporary = TRUE)
mtcars_copy <- DBI::dbReadTable(db2_conn, DBI::Id("SESSION", "mtcars_tmp"), row.names = TRUE)
identical(mtcars, mtcars_copy)
# [1] TRUE
```


jimhester:
Looks like the checks are failing due to not documenting the new methods.

jimhester:
Can you also please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

rnorberg:
Sorry, I don't know how to document this. I was just following the style of the other methods in the file. Is `@rdname hidden_aliases` not sufficient?

rnorberg:
Oh I just needed to Cntrl+Shift+D. I'm an idiot.

jimhester:
not at all, not at all. However the class name `DB2/AIX64` does seem to be causing problems with the documentation and `R CMD check`. I cannot figure out a way to escape it so that both work. We might have to sanitize the class name and convert the `/` to a `_` character when we create the class.

jimhester (approved):

rnorberg:
Nice idea to sanitize class names. A few questions:

* I know this is pedantic, but technically an underscore is not an alphanumeric character, so you're replacing all non-alphanumeric characters with a non-alphanumeric character. I don't think this is a bug necessarily, but it's maybe a "code smell"? I don't know.
* The regular expression `"[^[:alnum:]]"` matches whitespace. I'm afraid this might impact functionality specific to SQL Server. SQL Server connections are objects of class "Microsoft SQL Server", which will now be of class "Microsoft_SQL_Server". I'm afraid this might break things [here](https://github.com/r-dbi/odbc/blob/master/R/DataTypes.R#L194) or [here](https://github.com/r-dbi/odbc/blob/master/src/odbc_connection.cpp#L99). Perhaps it would be better to just replace all *punctuation* with underscores? (e.g. `"[[:punct:]]"` instead of `"[^[:alnum:]]"`)
* Your change actually changes the `info$dbms.name` attribute of the created connection, as well as the class. I'm not sure this matters, but to overcome the `Rd` issue we only need to change the class, not the `info$dbms.name` attribute. I guess what I'm asking is, do you think there are things downstream that are predicated on the `info$dbms.name` that might break?

jimhester:
I decided to scrap the class sanitizing, I figured out a way to work around the R CMD check issues without having to change the class name by documenting the methods differently.

Thanks again for your work and opening the pull request!

jimhester:
Thanks again for working on this!

Thank you!

rnorberg:
Wow, that `man/sqlCreateTable-methods.Rd` is black magic. I have no idea what's going on there. Nice job getting to the bottom of it. Thanks for reviewing this!

PR #431, opened 2021-01-14
Encode non-ASCII Column Names to UTF-8 

### LINKED ISSUE #430: The Encoding of the column names is not UTF-8

When the SQL contains non-ASCII strings as the field names, the returned **column names** of the data.frame is not encoded as UTF-8.

### Database

SQL Server 2014 

### Reproducible Example

Example:
```r
library(odbc)
library(DBI)
con <- dbConnect(odbc::odbc(), dsn = "SQL Server")
x <- DBI::dbGetQuery(con, "select '汉语' as 中文")
Encoding(x[[1]]) # returns UTF-8
Encoding(colnames(x)) # returns unknown and are actually not UTF-8 encoded
```

image

### LINKED ISSUE #432: Garbage error message when the encodings of the server and client are different

When the database's encoding is GBK, the client's encoding is UTF-8 and the SQL contains non-ASCII letters, the error message becomes garbage letters.

The reason is that odbc converts the SQL to the encoding of the Database first. Thus, when nanodbc throw errors, if the error message contains the SQL string, the encoding of the error message is the same as the database, which differs from the encoding of the Client. 

Thus, the error message displayed on the client becomes garabage.

The fix I believe is to convert the encoding of the error message to the client's encoding.

### Database

MS SQL Server 2014

### Reproducible Example

- A database with encoding other than UTF-8, e.g., GBK
- A client (the user machine) uses encoding UTF-8
- A SQL with non-ASCII letters

```r
con <- DBI::dbConnect(odbc::odbc(), drv = "SQL Server", encoding = "GBK", ...)
sql <- "select '中文' from 汉语"
# this is garabage message
DBI::dbGetQuery(con, sql)
# this fixes the message
tryCatch(DBI::dbGetQuery(con, sql), error = function(e) stop(iconv(e$message, "GBK", "UTF-8")))
```

image

shrektan:
Closes #430 
Closes #432

# Code

``` r
x <- DBI::dbGetQuery(con, "select '汉语' as '中文'")
x
Encoding(x[[1]])
Encoding(colnames(x))

DBI::dbGetQuery(con, "select '汉语' from '中文'”)
```

# Before the PR

image

# After the PR

image


detule:
Looks good -

Should we add some documentation in `odbc_result.h`?  Character enconding always confuses the hell out of me.

src/odbc_result.cpp
```diff
@@ -51,7 +51,9 @@ void odbc_result::execute() {
       num_columns_ = r_->columns();
     } catch (const nanodbc::database_error& e) {
       c_->set_current_result(nullptr);
-      throw odbc_error(e, sql_);
+      // #432: odbc_error() expects UTF-8 encoded strings but both nanodbc and sql are
+      // encoded in the database encoding, which may differ from UTF-8
+      throw odbc_error(CHAR(to_utf8(e.what())), CHAR(to_utf8(sql_)));
```
detule, line 56: Hey @shrektan

Nice work here - help me with this conversion:  Isn't the encoding  of `sql_` same as [R]/client/native; and not necessarily as specified by `db-enc`?

detule:
@shrektan Thanks for writing up the documentation; I was able to follow along better.  Nice work again.

This is very minor, but I was wondering about this:

Rather than introducing `to_utf8` where we guard the usage of `output_encoder_` with that `if c_->encoding() == ""` statement; how about we *always* call `output_encoder_`, unconditionally - both in `assign_string`, as well as in `column_names`?

As far as I can tell in `Iconv.cpp`, both `output_encoder_.makeSEXP` and `output_encoder_.makeString` are pretty smart / should just pass through to `Rf_mkCharLenCE(...., CE_UTF8)` if no conversion is needed (UTF8->UTF8).  The latter one, `output_encoder_.makeString` might be easiest to use in `column_names` and would perhaps allow you to stay with the current data types. 

We would probably need something like this in the constructor:
```
output_encoder_(Iconv(c_->encoding() == "" ? "UTF-8" : c_->encoding(), "UTF-8"))
```

The way I see it, this would save us from having character encoding helper methods both in `odbc_result` as well as the object in `Iconv`.

What do you think?

jimhester:
@shrektan any thought on @detule's last comment above? I have to do a odbc release to fix a compilation warning for CRAN, it would be nice to get this in there as well, but it has to be submitted by tomorrow. If we can't get it done by then it can wait until the next CRAN release, so it is not a huge deal if you don't have time to work on it right now.

shrektan:
@jimhester Unfortunately, I don't have time for this right now. Let's merge this in the next CRAN release.

detule:
Hey @shrektan 

Sorry this has been out here for so long; you did some excellent work here.

Hope you don't mind - [I tried to see](https://github.com/detule/odbc/commits/colname-utf8-rebased) what the changes would look like if, rather than adding a `to_utf8` routine in `odbc_result`, we just unconditionally funneled the strings to `output_encoder`, and let it decide if any conversion is needed.

What do you think?

Nice work again.

shrektan:
@detule It's great to see that these issues are starting to be addressed gradually. I've briefly reviewed your approach, and it looks good to me. Thanks.

PR #444, opened 2021-02-28
Fix single schema RStudio Connections pane issue

meztez:
HI,

First, thanks to the maintainers for all the great contributions. We all appreciate the time you put in.

The following statements might be wrong as they are my personal understanding.

I think the conditions in `odbcListObjects` should be greater than 0 instead of greater than 1. I could not a reason to recheck for nullity with %||% it is checked one line above.

Furthermore, I've tested it on MSSQL, Netezza. Works fine.

It fixes an issue with a db that has only one schema as the tables would be listed, but the Preview would fail because no schema would be fed to the preview function.

Have an excellen day.


jimhester:
Thanks for working on this, I think your understanding is correct, thanks!

meztez:
Hot damn. My English is horrifying when I type after an all-nighter. @jimhester  I'm sorry you had to read that.

PR #461, opened 2021-06-25
Db2 find temp tables in dbExistsTable and dbListTables

emgns04:
DB2 subclass- find temporary tables in dbExistsTable and dbListTables if possible. Defaults to normal behavior.
Thanks to @rnorberg for the suggestion.


detule:
Completed via https://github.com/r-dbi/odbc/pull/823.

PR #494, opened 2022-05-26
Avoid warning with connection observer callback conditions eval

meztez:
In R 4.2.0, when dbConnect is called within RStudio using something like:

```
odbc::dbConnect(odbc::odbc(), "DSN")
```
In https://github.com/r-dbi/odbc/blob/dee7ae0cc2d3d0d10d54d412d5c448e7870414c9/R/Driver.R#L123,
`as.character(expr[[1]])` will have values `[1] odbc      ::          dbConnect` and condition evaluates to `FALSE FALSE FALSE` thus cause a warning similar to

```
warning messages from top-level task callback '1'
Warning message:
In is.call(expr) && as.character(expr[[1]]) %in% c("<-", "=") :
  'length(x) = 3 > 1' in coercion to 'logical(1)'
```

I suggest using tail with a single element to return the expr name to the right if the `::`.


JannesRed:
Not sure what it would take to get this reviewed and approved?

detule:
Hi - 

Thanks for the submission.  Tailing could work.  Alternatively, rather than converting the expression to character, it may be better converting the assignment operators to expressions.  Something like:

```
op <- expr[[1]]
identical(op, quote(`<-`)) || identical(op, quote(`=`))
```

Please also add a bullet in the NEWS markdown file highlighting the fix.

Thanks again

detule (approved):

R/Driver.R
```diff
@@ -120,7 +120,7 @@ setMethod(
       addTaskCallback(function(expr, ...) {
         tryCatch({
           if (is.call(expr) &&
-              as.character(expr[[1]]) %in% c("<-", "=") &&
```
hadley, line 123: I think an even better fix here would be to switch to rlang's `is_call()`: `is_call(expr[[1]], c("<-", "="))`. Looks like rlang is already an dependency of odbc.

meztez: Did a couple tests, seems to work alright.

R/Driver.R
```diff
@@ -78,6 +78,7 @@ setMethod(
 #' \url{https://www.connectionstrings.com} is also a useful resource of example
 #' connection strings for a variety of databases.
 #' @aliases dbConnect
+#' @importFrom rlang is_call
```
hadley, line 81: It's not super important here, but in tidyverse packages we general just do `@import rlang`.

meztez: I'm curious to know the reason behind this preference. Could not find a reference to it in the tidyverse style guide (https://github.com/tidyverse/style). Have a great day.

meztez: Got it. Thanks for taking the time to write down an explanation. Let me switch it.

detule:
Thanks again!🏆

PR #513, opened 2022-10-17
Workflow updates

detule:
1. Update github actions image
2. Install pkgbuild prior to calling testthat::test_local since it is a depedency to pkgload ( in some circumstances )

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -29,12 +29,12 @@ jobs:
           - {os: windows-latest, r: '3.6'}
 
           # Use older ubuntu to maximise backward compatibility
-          - {os: ubuntu-18.04,   r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-18.04,   r: 'release'}
-          - {os: ubuntu-18.04,   r: 'oldrel-1'}
-          - {os: ubuntu-18.04,   r: 'oldrel-2'}
-          - {os: ubuntu-18.04,   r: 'oldrel-3'}
-          - {os: ubuntu-18.04,   r: 'oldrel-4'}
+          - {os: ubuntu-20.04,   r: 'devel', http-user-agent: 'release'}
+          - {os: ubuntu-20.04,   r: 'release'}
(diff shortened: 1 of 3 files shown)


.github/workflows/R-CMD-check.yaml
```diff
@@ -29,12 +29,12 @@ jobs:
           - {os: windows-latest, r: '3.6'}
 
           # Use older ubuntu to maximise backward compatibility
-          - {os: ubuntu-18.04,   r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-18.04,   r: 'release'}
-          - {os: ubuntu-18.04,   r: 'oldrel-1'}
-          - {os: ubuntu-18.04,   r: 'oldrel-2'}
-          - {os: ubuntu-18.04,   r: 'oldrel-3'}
-          - {os: ubuntu-18.04,   r: 'oldrel-4'}
+          - {os: ubuntu-20.04,   r: 'devel', http-user-agent: 'release'}
```
hadley, line 32: I think this should be latest not 20; you might be better off running `usethis::use_github_action_check_full()` to get our latest changes.

.github/workflows/db.yaml
```diff
@@ -106,6 +106,7 @@ jobs:
       - name: Install dependencies
         run: |
           remotes::install_deps(dependencies = TRUE)
+          remotes::install_cran('pkgbuild')
```
hadley, line 109: I think it would be better to switch to `uses: r-lib/actions/setup-r-dependencies@v2` rather than adding more  exceptions.

hadley (commented):
Thanks for taking a look at this!

.github/workflows/db.yaml
detule: Done / migrated to using setup-r-dependencies.

I am not entirely convinced the package cache ( in this step / setup-r-dependencies ) is used in any way;  I can chase that down later, if that works for you.

detule:
Thanks for the feedback! Let me know if you had something else in mind.

hadley (approved):

hadley:
Thanks!

PR #517, opened 2022-11-04
Feature/query timeout option

### LINKED ISSUE #515: Query timeout option (SQL_ATTR_QUERY_TIMEOUT)
Previously the RODBCext package provided an option to set the [query timeout](https://github.com/zozlak/RODBCext/blob/master/R/odbcSetQueryTimeout.R). As of 2020 the package is deprecated and the better alternative is [odbc](https://github.com/r-dbi/odbc) ([cran](https://cran.r-project.org/web/packages/odbc/index.html)).

This odbc package doesn't seem to provide the described option. I was wondering if there is another way around to implement a query timeout while using odbc (perhaps something like the [C code](https://github.com/zozlak/RODBCext/blob/master/src/RODBCext.c#L426) of the RODBCext package which sets the value of SQL_ATTR_QUERY_TIMEOUT).

(R does have a [withTimeout](https://stackoverflow.com/q/70993017/1974344) function, but I want to have a reliable/ correct implementation)

The nanodbc.cpp file does have a [timeout function](https://github.com/r-dbi/odbc/blob/main/src/nanodbc/nanodbc.cpp#L1419) which sets the required SQL_ATTR_QUERY_TIMEOUT attribute, but this doesn't seem to be available in the R code?

### Database
Microsoft SQL Server

### Reproducible Example
con <- dbConnect(
  odbc::odbc(),
  driver = "SQL Server",
  server = "SERVERNAME",
  trusted_connection = TRUE,
  timeout = 5,
  encoding = "Latin1")

samplequery <- dbGetQuery(con, "
     SELECT * FROM [Something]   
")

Odraio:
# Issue

Current odbc main repo does not provide a query_timeout option.

Closes #515

# Feature description

`dbGetQuery()`, `dbSendStatement()` and `dbSendQuery()` provide a `query_timeout` option (internally sets `SQL_ATTR_QUERY_TIMEOUT`).

`query_timeout`: The number in seconds before query timeout (can be used to stop long running queries). 
Default is 0 indicating no timeout. 

**Example:**
``` r
# long running query stops after 60 seconds
# provided statement is a placeholder

dbGetQuery(con, "SELECT flight, tailnum, origin FROM flights ORDER BY origin", query_timeout = 60)

dbSendStatement(con, "SELECT flight, tailnum, origin FROM flights ORDER BY origin", query_timeout = 60)

dbSendQuery(con, "SELECT flight, tailnum, origin FROM flights ORDER BY origin", query_timeout = 60)

```

**Additional request:**
- After merging, please create a new (CRAN) release.


detule:
Hi - thanks for the submission.

It would be challenging to review this PR because of the way the diff comes through.  You may be using text editing software (windows?) that alters the line endings.  Please resubmit with a clean diff.

Also,  commenting out the ASSERT is likely not the right approach.  I would leave that line unaltered in your next submission; we can address elsewhere.

Odraio:
@detule I've pushed a new version of the branch which results in a clean diff.

Odraio:
There is one issue while using the query_timeout and params. I will check and fix this!

Odraio:
A new fix has been pushed which resolves the issue of a query using _params_ and _query_timeout_.

Odraio:
@detule Did you have an opportunity to review and approve the pull request?

shrektan:
It looks good to me overall.

- Could you please add the doc for `query_timeout` as well? 
- In addition, it's better to set the default value to `0` instead of `Inf`.
- Lastly, it would be great if you can update the `NEWS.md` file.

Thanks.

R/Result.R
```diff
@@ -9,11 +9,11 @@ NULL
 #' @docType methods
 NULL
 
-OdbcResult <- function(connection, statement, params = NULL, immediate = FALSE) {
+OdbcResult <- function(connection, statement, params = NULL, immediate = FALSE, query_timeout = 0L) {
```
shrektan, line 12: better keep the default value exactly the same.

```suggestion
OdbcResult <- function(connection, statement, params = NULL, immediate = FALSE, query_timeout = 0) {
```

src/odbc_result.cpp
```diff
@@ -7,20 +7,21 @@
...
+    std::shared_ptr<odbc_connection> c, std::string sql, bool immediate, long query_timeout)
     : c_(c),
       sql_(sql),
       rows_fetched_(0),
       num_columns_(0),
       complete_(0),
       bound_(false),
-      output_encoder_(Iconv(c_->encoding(), "UTF-8")) {
+      output_encoder_(Iconv(c_->encoding(), "UTF-8")),
+      query_timeout_(query_timeout){
```
shrektan, line 18:
```suggestion
      query_timeout_(query_timeout) {
```

shrektan (approved):

Odraio:
@detule and @shrektan Thanks for reviewing! I will update the pull request later on this week.

It seems there is always some confusion between a _connection_ timeout and an actual _query_ timeout. To avoid this, I've choosen for the _query_timeout_ parameter name. Is it still required to rename _query_timeout_ to _timeout_?

@detule 
> Move the documentation out of the README to the roxygen doc blocks for all [R] methods that gained a timeout parameter.

Why should it be moved out of the README? I think it should be available as well in the README section on the GitHub home page of the project. Users will immediatly have an example.

detule:
Hi:
* Parameter name: The use of the ```timeout``` argument to ```dbConnect``` is documented in the roxygen tags for that function; I think that should clear up ambiguity between ```dbConnect(..., timeout,... )``` and ```dbGetQuery(..., timeout, ...)```.  In addition, and admittedly this is a bit subjective, but ```dbGetQuery(..., query_timeout, ...)``` feels like repeated usage of ```"query"```.
* In terms of the README: I think that document covers some basic functionality of the odbc package - features that are (mostly) guaranteed to work across databases and various drivers; note there are many details missing, especially some that are not (guaranteed to be) universally supported.  I think the ```timeout``` parameter to the query methods falls in that latter category.  It is a very useful addition - I thought about maybe adding your examples in an ```@examples``` block,  but I don't even think there is enough complexity in the usage of the new ```timeout``` parameter that would warrant that.  Its purpose and usage is pretty clear-cut!

  Between the announcement in the NEWS file, and the added documentation, I think the feature will be well advertised and understood.

Thanks again for the submission.

hadley:
@Odraio can you reproduce build failures locally? It works for my when I test with my local sql server, so I'm a bit mystified as to what is going wrong.

Issue #299, opened 2019-09-10
Setting pre-connect connection attributes for Azure SQL Server connection


kjaanson:
### Connecting to Azure SQL database using access token via pre-connect connection attributes

Is it possible to set pre-connect attributes on a connection using r odbc? I want to use Active Directory access token to connect to Azure SQL Data Warehouse. Azure docs specifiy that the token needs to be given to the driver as a pre-connect attribute SQL_COPT_SS_ACCESS_TOKEN (https://docs.microsoft.com/en-us/sql/connect/odbc/using-azure-active-directory?view=sql-server-2017).

I know that it is doable in pyodbc but I have not seen any implementation or mention of pre-connect connection attributes in r-odbc docs.

jimhester:
No, there is currently no support for specifying pre-connection attributes. Adding support for it would likely require a decent amount of work, it is unlikely to be done without an external PR.

samssann:
This would be a very welcome feature.

hongooi73:
@jimhester I've received a number of requests for something like this. What is required to make this happen? Is there anything I can contribute?

jimhester:
As mentioned in https://github.com/nanodbc/nanodbc/issues/215 we would have to modify nanodbc to support this, similar to what was done in pyodbc, and then expose an R interface for it in this odbc package.

olesendan:
Hi everybody.

I'm one of persons, trying to implement this in my workflow. I would like to contribute aswell. I consider myself a newbie in software development, and only know R. But I am willing to learn an have some time to work on a solution.

julieinsan:
Upvote and offer to contribute.  We are dealing with this issue presently as the only means of connecting to Azure SQL DB to satisfy our security team.

JoshRosenstein:
Bump. This would be extremely helpful.

innir:
Now that nanodbc seems to have support for it (see https://github.com/nanodbc/nanodbc/commit/b0fa192eb7289bba5450f90ad57075dec3a6581c), things might get going :-)

detule:
Yes / should have a solution that works for us here soon.

BaukjeStienstra:
That would be very nice! But I do have a question about this enhancement of nanodbc: Is there also a need to modify the odbc/DBI R-packages to make it available in an R-environment? And if so, when will this be available?

kummerer94:
[...]

PR #521, opened 2022-11-29
Ability to set (pre) connection attributes ....

detule:
... and use this functionality to enable users to authenticate against Azure/MSSQL using Azure AD token.

This closes https://github.com/r-dbi/odbc/issues/299

Notes:

* First commit, in src/nanodbc, is an adaptation of [recent upstream work](https://github.com/nanodbc/nanodbc/commit/b0fa192eb7289bba5450f90ad57075dec3a6581c) to enable setting of generic pre-connection attributes.  As mentioned in the commit message, needed to adapt the upstream solution since there it requires C++17.
* Second commit plumbs this functionality though to the [R] API.  Initially [i added an](https://github.com/detule/odbc/commit/730410e0cc11732b70ad470079f090c0c7c8316d) ```azure.token``` parameter to ```dbConnect```.  This ended up looking pretty ugly for various reasons.  The solution as coded here would have the user do something like
```
r_attr = new("MSSQLConnectionAttributes", azure_token = .token)
conn <- dbConnect(
  odbc::odbc(),
  dsn = "my_azure_mssql_db",
  Encrypt = "yes",
  attributes = r_attr)
```
This should scale better, and IMO is just plain prettier.


R/Driver.R
```diff
@@ -61,6 +62,8 @@ setMethod(
 #' name for the OdbcConnect object returned from `dbConnect()`. However if the
 #' driver does not return a valid value it can be set manually with this
 #' parameter.
+#' @param attributes An S4 object of connection attributes that are passed
```
hadley, line 65: Why make this an S4 object and not (e.g.) just a named list?

detule: It seemed a bit awkward housing connection attributes that may be specific to individual DBs in the same structure.
Without needing dispatch, I probably would not have used S4, but its already all over DBI so it seemed OK.

Happy to go a different route if you think that's an overkill.

hadley: Ooooh, this isn't a generic way to pass name-value pairs to the driver, it's specifically tailored to solve the Azure issue?

detule: Sorry for the confusion.

It is intended to pass name-value pairs to the driver.  However these come in different flavors.  Some are generic  ODBC attributes ( timeout, catalog, etc ) that are (expected to be) implemented across all back-ends, and in time we can support these through this interface; they would fit naturally as slots of the parent ```ConnectionAttributes``` object.  However there may be other attributes that are back-end specific.  The azure token is one of them.

hadley: Hmmm, in that case I still think it would be slightly easier to provide a named list. You make it generic by providing a an `odbc_serialize()` generic which you'd apply to each element of the list. And for now you'd just start with `AzureToken`. Does that make sense?

hadley: Yeah, that sounds good.

detule: Thanks / should be done; let me know if I missed anything.

Pipeline looks to be failing, but I believe it is for an unrelated reason, [fixed here.](https://github.com/r-dbi/odbc/pull/522)

R/Connection.R
```diff
@@ -1,6 +1,30 @@
 #' @include Driver.R
 NULL
 
+#' Supported Connection Attributes
+#'
+#' These (pre) connection attributes are supported and can be passed as
+#' part of the `dbConnect` call in the named list `attributes` parameter:
+#'
+#' * azure_token: This should be a string scalar; in particular Azure Active
```
hadley, line 9:
```suggestion
#' * `azure_token`: This should be a string scalar; in particular Azure Active
```

R/Connection.R
```diff
@@ -1,6 +1,30 @@
...
+#' @examples
+#' \dontrun{
+#' conn <- dbConnect(
+#'   odbc::odbc(),
+#'   dsn = "my_azure_mssql_db",
+#'   Encrypt = "yes",
+#'   attributes = list("azure_token" = .token)
+#' }
+#' @export
+SUPPORTED_CONNECTION_ATTRIBUTES <-
```
hadley, line 25: Do you want to export this object, or just make this a home for documenting any additional attributes?

hadley (approved):

R/Connection.R
detule: Good eye - un-exported.

Thanks for all the feedback.

PR #523, opened 2022-12-09
backport: nanodbc: ensure_pdata: Return unique_ptr so as to not leak resources

detule:
This patches an obvious memory leak, likely manifesting as a very small but persistent loss of resources affecting SQL Server users.

The `ensure_pdata` code-path is traversed with the `test_out_of_order_blob` unit test.

### DIFF
--- a/src/nanodbc/nanodbc.cpp
+++ b/src/nanodbc/nanodbc.cpp
@@ -2724,7 +2724,7 @@ public:
 
 private:
     template <typename T>
-    T* ensure_pdata(short column) const;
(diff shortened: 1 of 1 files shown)


hadley:
How confident are you that this is correct? I'm happy to trust you, but if you think it needs to be reviewed by someone who can double check your work, I'll need to ask someone else.

detule:
Hi Hadley:

Additional feedback is always welcome.  However, I am pretty confident.  When I wrote the upstream patch I had a test jig that I used to iterate over with valgrind and monitor the leak.

Thanks

Issue #502, opened 2022-08-05
Limit on query for previewing large tables


MariaSemple:
### Issue Description and Expected Result
When calling `previewObject` even with a small row limit such as 50, it still takes a long time for the query to return when the table in question is quite large. This can cause an RStudio Session to hang while the query is being performed.

Ideally, if we limit the number of rows in the preview, we could query only that many rows to avoid the hang in the IDE. A `LIMIT` or a `TOP` statement [here](https://github.com/r-dbi/odbc/blob/09ef0a23a191b6f1496e9243b91157f83a32d9ae/R/Viewer.R#L241) would fit that need, but that may not be possible due to database language support differences.

### Database
I'm unsure if this is specific to one database or another, but we have a reproducible example on the IDE team, so I will get the information about that table and its Database and update it here (it has something like 30000 rows).

### Reproducible Example

Specific details about loading the database to come.

1. In an RStudio Session, navigate through the Connections pane until you find a large table. 
2. Preview the table by clicking on it.
3. Even though the preview is limited to 1000 rows, it takes quite some time for the query to return. Restricting the limit to 50 has no impact on performance.

<details>
<summary>Session Info</summary>

```r
devtools::session_info()
#> output
```
</details>

detule:
Hi - thanks.

Please if you get an opportunity do revert with the back-end, the driver you are using, and size of table ( or maybe view? ).   Also if you can note if there are out-of-the-ordinary characteristics and fields ( VARBINARY / blobs for example ), that would be helpful.

30K rows is not a size where I would have expected an observable delay in fetching 1000 rows from `SELECT *`

PR #525, opened 2022-12-28
odbcPreviewObject: Optimize SELECT against rowLimit

detule:
This closes https://github.com/r-dbi/odbc/issues/502

When previewing large tables, optimize the `SELECT *` query since we are only interested in a limited number of rows.

Testing locally against SQL Server with tables as larges as 70K rows shows limited benefits.  However, when testing against a Snowflake instance with >[ 1B rows tables](https://docs.snowflake.com/en/user-guide/sample-data-tpcds.html), benefit is very tangible.

Note: Current default for the query of the preview format is `SELECT * FROM <table> LIMIT N`, with exceptions for SQL Server, and Oracle.  Tested the `LIMIT N` syntax successfully against MYSQL, PSQL, DB2, SNOWFLAKE.  If there are users of other more esoteric back-ends that do not support the `LIMIT N` suffix, we can add S3 methods for them.

Alternative would have been to leave the default preview query un-optimized as `SELECT * FROM <table>`, and write S3 methods for all of the above that route to the `LIMIT N` implementation.

I have a small bias for the way it is written in this commit ( assume `LIMIT N` works for other back-ends and address as needed ), but can go either way.

Thanks


R/Viewer.R
```diff
@@ -238,7 +238,39 @@ odbcPreviewObject.OdbcConnection <- function(connection, rowLimit, table = NULL,
     name <- paste(dbQuoteIdentifier(connection, catalog), name, sep = ".")
   }
 
-  dbGetQuery(connection, paste("SELECT * FROM", name), n = rowLimit)
+  dbGetQuery(connection, odbcPreviewQuery(connection, rowLimit, name ),
```
hadley, line 241: Argument alignment; again http://github.com/lionel-/codegrip is likely the easiest way to see what we want.

R/Viewer.R
```diff
@@ -238,7 +238,39 @@ odbcPreviewObject.OdbcConnection <- function(connection, rowLimit, table = NULL,
...
+#' @param connection A connection object, as returned by `dbConnect()`.
+#' @param rowLimit The maximum number of rows to display.
+#' @param name Name of the object to be previewed
+odbcPreviewQuery <- function(connection, rowLimit, name) {
+  UseMethod("odbcPreviewQuery")
+}
+
+#' Common top-N syntax ( MYSQL, PSQL, DB2, SNOWFLAKE, etc )
+#' @rdname odbcPreviewQuery
+odbcPreviewQuery.OdbcConnection <- function(connection, rowLimit, name) {
```
hadley, line 260: Probably worth a couple of tests?

hadley (approved):
Worth a NEWS bullet too, I think.

tests/testthat/test-MySQL.R
```diff
@@ -78,9 +78,10 @@ test_that("MySQL", {
     con <- DBItest:::connect(DBItest:::get_default_context())
     dbWriteTable(con, tblName, data.frame(a = 1:10L))
     on.exit(dbRemoveTable(con, tblName))
+    # There should be no "Pending rows" warning
     expect_warning({
```
hadley, line 82: You can use `expect_no_warning()` now, for a bit more clarity.

Issue #458, opened 2021-06-11
Avoid querying column types from the database in dbWriteTable() and dbAppendTable()


krlmlr:
### Issue Description and Expected Result

`dbWriteTable()` on SQL Server 12.xx fails with BLOB columns **if the table name is given in quotes**. (I spent too much time isolating this.)

I suspect we can't make it work with quotes, this must be some SQL server quirk. Unfortunately, this also means that `dplyr::copy_to()` fails in these cases. The internet equivocally suggests using `INSERT INTO xxx VALUES (CONVERT(varbinary(max), ?))` .

Should we introduce a new argument to `dbWriteTable()` and `dbAppendTable()` that supports custom conversions for the input? Are there other use cases besides this particular problem?

### Database

SQL Server 12.xx (doesn't happen on SQL Server 15.xx on Ubuntu).

### Reproducible Example

``` r
library(DBI)

drv <- odbc::odbc()

uid <- Sys.getenv("...")
pwd <- Sys.getenv("...")
con <- dbConnect(
  drv,
  dsn = "dwh",
  uid = uid, pwd = pwd, encoding = "utf-8"
)

con
#> <OdbcConnection> SV0118
#>   Database: master
#>   Microsoft SQL Server Version: 12.00.6164

data <- data.frame(a = 1:3)
data$b <- blob::as_blob(lapply(lapply(1:3, runif), serialize, connection = NULL))

dbWriteTable(con, "blob_test", data)
identical(dbReadTable(con, "blob_test"), data)
#> [1] TRUE
dbRemoveTable(con, "blob_test")

dbWriteTable(con, SQL("[blob_test]"), data)
#> Error in result_insert_dataframe(rs@ptr, values, batch_rows): nanodbc/nanodbc.cpp:1655: 00000: [FreeTDS][SQL Server]Implicit conversion from data type varchar to varbinary is not allowed. Use the CONVERT function to run this query.  [FreeTDS][SQL Server]Statement(s) could not be prepared.
identical(dbReadTable(con, "blob_test"), data)
#> [1] FALSE
dbRemoveTable(con, "blob_test")
```

<sup>Created on 2021-06-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

detule:
Hey @krlmlr 

Is this FreeTDS specific? Everything below is assuming that the answer to above is YES - feel free to ignore otherwise.

[...]

PR #526, opened 2022-12-28
odbcConnectionColumns: Method ingesting DBI::SQL

detule:
This closes https://github.com/r-dbi/odbc/issues/458

Adds a method to `odbc::odbcConnectionColumns` that ingests `DBI::SQL`.

Currently, in addition to exposing the method so that users can inspect table structure, we use `odbcConnectionsColumns` when writing to tables to work around a deficiency in the `FreeTDS` implementation of the ODBC API for SQL Server.

Similar to, for example, `dbExistsTable`, here too we unquote the identifier before passing it to the ODBC API.  This is necessary for catalog functions that are part of the ODBC API ( like `SQLTables`, `SQLColumns` etc).

Thanks

### DIFF
--- a/R/Connection.R
+++ b/R/Connection.R
@@ -176,6 +176,23 @@ setMethod(
   }
 )
 
+#' @details odbcConnectionColumns is routed through the SQLColumns ODBC
+#'  method.  This function, together with remaining catalog functions (SQLTables,
+#'  etc), by default ( SQL_ATTR_METADATA_ID == false ) expect either ordinary
+#'  arguments (OA) in the case of the catalog, or pattern value arguments (PV)
+#'  in the case of schema/table name.  For these, quoted identifiers do not make
+#'  sense, so we unquote identifiers prior to the call.
+#' @seealso The ODBC documentation on [Arguments to catalog functions](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/arguments-in-catalog-functions?view=sql-server-ver16).
+#' @rdname odbcConnectionColumns
+#' @param conn OdbcConnection
+#' @param name table we wish to get information on
+#' @export
+setMethod(
+  "odbcConnectionColumns", c("OdbcConnection", "SQL"),
+  function(conn, name, ...) {
+    odbcConnectionColumns(conn, dbUnquoteIdentifier(conn, name)[[1]], ...)
+  })
+
 # TODO: show encoding, timezone, bigint mapping
 #' @rdname OdbcConnection
 #' @inheritParams methods::show
--- a/R/Connection.R
+++ b/R/Connection.R
@@ -176,16 +176,14 @@ setMethod(
   }
 )
 
-#' @details odbcConnectionColumns is routed through the SQLColumns ODBC
-#'  method.  This function, together with remaining catalog functions (SQLTables,
-#'  etc), by default ( SQL_ATTR_METADATA_ID == false ) expect either ordinary
-#'  arguments (OA) in the case of the catalog, or pattern value arguments (PV)
-#'  in the case of schema/table name.  For these, quoted identifiers do not make
-#'  sense, so we unquote identifiers prior to the call.
+#' @details `odbcConnectionColumns` is routed through the `SQLColumns` ODBC
+#'  method.  This function, together with remaining catalog functions
+#'  (`SQLTables`, etc), by default ( `SQL_ATTR_METADATA_ID == false` ) expect
+#'  either ordinary arguments (OA) in the case of the catalog, or pattern value
+#'  arguments (PV) in the case of schema/table name.  For these, quoted
+#'  identifiers do not make sense, so we unquote identifiers prior to the call.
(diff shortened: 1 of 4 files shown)


R/Connection.R
```diff
@@ -176,6 +176,23 @@ setMethod(
   }
 )
 
+#' @details odbcConnectionColumns is routed through the SQLColumns ODBC
```
hadley, line 179: Needs `` `odbcConnectionColumns` `` etc

R/db.R
```diff
@@ -91,3 +91,22 @@ setMethod("sqlCreateTable", "DB2/AIX64",
     ))
   }
 )
+
+# Microsoft SQL Server ---------------------------------------------------------
+
+# Simple class prototype to avoid messages about unknown classes from setMethod
+setClass("Microsoft SQL Server", where = class_cache)
```
hadley, line 98: And this class gets correctly overridden if we're actually using the SQL server driver?

tests/testthat/test-SQLServer.R
```diff
@@ -173,6 +173,26 @@ test_that("SQLServer", {
     expect_equal(as.double(values[[6]]), as.double(received[[6]]))
   })
 
+  local({
+    con <- DBItest:::connect(DBItest:::get_default_context())
+    input <- DBI::SQL(c(
```
hadley, line 178: Can you indent long arguments with a single tab? (https://style.tidyverse.org/syntax.html#long-lines). You might want to use https://github.com/lionel-/codegrip to do it for you.

tests/testthat/test-SQLServer.R
```diff
@@ -173,6 +173,26 @@ test_that("SQLServer", {
...
+               "[testdb].[testschema].[testtable]",
+               "[testdb].[testschema].testtable" ))
+    expected <- c(
+                  DBI::Id( table = "testtable" ),
+                  DBI::Id( table = "testtable" ),
+                  DBI::Id( schema = "testschema", table = "testtable" ),
+                  DBI::Id( schema = "testschema", table = "testtable" ),
+                  DBI::Id( catalog = "testdb", schema = "testschema", table = "testtable" ),
+                  DBI::Id( catalog = "testdb", schema = "testschema", table = "testtable" ) )
+    expect_identical( DBI::dbUnquoteIdentifier( con, input ), expected )
```
hadley, line 192: I'm a bit surprised that `dbUnquoteIdentifier` returns an `Id`, but that's a DBI issue, and too late to change. (Although the docs could be better)

tests/testthat/test-SQLServer.R
```diff
@@ -173,6 +173,26 @@ test_that("SQLServer", {
...
 
+  local({
+    con <- DBItest:::connect(DBItest:::get_default_context())
+    input <- DBI::SQL(c(
+               "testtable",
+               "[testtable]",
+               "[testschema].[testtable]",
+               "[testschema].testtable",
+               "[testdb].[testschema].[testtable]",
+               "[testdb].[testschema].testtable" ))
```
hadley, line 184: Maybe it's getting to exotic but do we need a test for a keyword containing a `"` or `[`?

tests/testthat/test-SQLServer.R
detule: Thanks - tried to address ( manually, will look into using codegrip ).

tests/testthat/test-SQLServer.R
detule: Thanks; added a few more.

R/db.R
detule: Good question -
For any back end ( and any driver ) these catalog functions need their arguments to be stripped of the quoted delimiters ( according to the ODBC spec ), so I think this is a good fix.

If we were to somehow get this wrong, however, when writing to tables only `FreeTDS` ( or other drivers that suffer from similar deficiencies ) would be affected, since other drivers can recover gracefully from us no being able to describe the column properties of the target table.

R/Connection.R
detule: Thanks - tried to address.  Let me know if I missed anything

tests/testthat/test-SQLServer.R
```diff
@@ -176,19 +176,25 @@ test_that("SQLServer", {
...
+      "[testdb].[testschema].[testtable]",
+      "[testdb].[testschema].testtable" ))
     expected <- c(
-                  DBI::Id( table = "testtable" ),
-                  DBI::Id( table = "testtable" ),
-                  DBI::Id( schema = "testschema", table = "testtable" ),
-                  DBI::Id( schema = "testschema", table = "testtable" ),
-                  DBI::Id( catalog = "testdb", schema = "testschema", table = "testtable" ),
-                  DBI::Id( catalog = "testdb", schema = "testschema", table = "testtable" ) )
+      DBI::Id( table = "testtable" ),
```
hadley, line 189: And shouldn't have wrapping spaces here.

hadley (approved):

PR #531, opened 2023-01-14
Encode non-ASCII Column Names to UTF-8

### LINKED ISSUE #430: The Encoding of the column names is not UTF-8

### LINKED ISSUE #432: Garbage error message when the encodings of the server and client are different

detule:
Closes #430
Closes #432

This is a rebase of #431 - details there.

Thanks @shrektan.


hadley:
What do we need to do to get this merged?

hadley:
@shrektan could you please take a quick look so we can get this into odbc?

shrektan:
It looks good to me.

shrektan (approved):
looks good to me.

simonpcouch:
From what I can tell in public comments, this PR is ready to merge. cc @hadley

PR #534, opened 2023-01-21
tests: drivers: skip if no data sources

detule:
Fixup unit tests in CI.

Looks like there may be one outstanding warning from coming from `sprintf` being used in a header included from Rcpp.  FWICT, this is fixed on github; probably see it as part of Rcpp's January CRAN release.


hadley (approved):

hadley:
Yeah, that shouldn't cause a problem on CRNA.

Issue #558, opened 2023-04-06
Error: invalid names for slots of class “Microsoft SQL Server”: ptr, quote, info, encoding


skanskan:
Hello.

I'm trying to run this repository
https://github.com/oxford-pharmacoepi/Studyathon_LCPASC

And I have problems when trying to create a connection to a SQL Server with DBI and odbc.

```
library(DBI)
library(odbc)
DBI::dbConnect(odbc(),
+                       Driver = 'sql server',
+                       Server = 'myserver',
+                       UID = 'myuser',
+                       PWD = 'mypassword,
+                       Port = myport
+  )
```

I have also tried with 'dsn' and 'user' = instead of Server and UID. 
And with different values for Driver, such as 'SQL Server', 'ODBC Driver 17 for SQL Server', 'SQL Server Native Client 11.0', 'sqlserver'...

But it keeps producing this error:
```
Found more than one class "Microsoft SQL Server" in cache; using the first, from namespace 'DatabaseConnector'
Also defined by ‘.GlobalEnv’
Error in initialize(value, ...) : 
  invalid names for slots of class “Microsoft SQL Server”: ptr, quote, info, encoding
```
I have tried reinstalling odbc and DBI from cran and also from github, with no luck.

### Database in the server
SQL Server 2017 v14.0

### Drivers installed on my local computer
ODBC Driver 17 for SQL Server
SQL Server 10.0
SQL Server Native Client 11

<details>
<summary>Session Info</summary>

```r
devtools::session_info()

R version 4.1.2 (2021-11-01)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 18363)

Matrix products: default

locale:
[1] LC_COLLATE=Spanish_Spain.1252  LC_CTYPE=Spanish_Spain.1252    LC_MONETARY=Spanish_Spain.1252 LC_NUMERIC=C                  
[5] LC_TIME=Spanish_Spain.1252    
system code page: 65001

- Session info --------------------------------------------------------------------------------------------------------------------------
 setting  value
 version  R version 4.1.2 (2021-11-01)
 os       Windows 10 x64 (build 18363)
 system   x86_64, mingw32
 ui       RStudio
 language (EN)
 collate  Spanish_Spain.1252
 ctype    Spanish_Spain.1252
 tz       Europe/Paris
 date     2023-04-06
 rstudio  2021.09.2+382 Ghost Orchid (desktop)
 pandoc   NA

[...]

PR #559, opened 2023-04-08
class_cache: Parented by package env

detule:
Hi @hadley:

I think this fixes https://github.com/r-dbi/odbc/issues/558

To be perfectly honest, main motivation here is not necessarily to fix, but to protect against what looks to be the issue there ( clobbering S4 object definitions ).  I suspect that the [`new` call ](https://github.com/r-dbi/odbc/blob/main/R/Connection.R#L84)picks up another definition of, for example, `Microsoft SQL Server`.

As best as I can tell there is no downside to having package environment be the parent of the class cache: [`getClassDef`](https://github.com/r-dbi/odbc/blob/main/R/Connection.R#L79) is called with `inherits=FALSE` which I think prevents lookup outside of the specified environment ( enclosing environments etc ).

The upside is that the subsequent `new` call should prefer the odbc S4 definitions over other in the global environment that someone else may have defined.

There are other ways to address this, I think.  I thought about for example, evaluating the `new` call in the `class_cache` environment.  This also works but one needs to do some yoga to ensure `package::methods` is found in the environment as well.

How I tested:
```
> library(odbc)
> conn <- dbConnect(odbc::odbc(), dsn = "mssql_oem_db")
> setClass("Microsoft SQL Server", slots = list(name = 'character'))
> conn <- dbConnect(odbc::odbc(), dsn = "mssql_oem_db")
Error in initialize(value, ...) :
  invalid names for slots of class “Microsoft SQL Server”: ptr, quote, info, encoding
```

### DIFF
--- a/R/Connection.R
+++ b/R/Connection.R
@@ -31,7 +31,7 @@ SUPPORTED_CONNECTION_ATTRIBUTES <-
 #' @name OdbcConnection
 NULL
 
-class_cache <- new.env(parent = emptyenv())
+class_cache <- new.env()
 
 OdbcConnection <- function(
   dsn = NULL,


skanskan:
Hello.

I have tried
setClass("Microsoft SQL Server", slots = list(name = 'character'))
before dbConnect
but I'm still getting the same error.

hadley:
See note in related issue.

Issue #540, opened 2023-02-06
Performance issue of `dbListTables()` and `dbWriteTable()` with `package:odbc` versus `package:ROracle`


ThomasSoeiro:
### Issue Description and Expected Result
`dbListTables()` and `dbWriteTable()` (and maybe other functions) are excessively longer when connected to an Oracle database using `package:odbc` and RStudio's Professional drivers rather than when using `package:ROracle`.

I don't know if this is related to `package:odbc` or to RStudio's Professional drivers. I asked my IT department to [set up Oracle's OEM driver](https://solutions.posit.co/connections/db/databases/oracle/) to test it. I will report the results if they do. I have no idea how to investigate the issue further but would be happy to provide more information. 

### Database
```r
dbGetInfo(conn_odbc_rstudio)
# $dbname
# [1] ""
# 
# $dbms.name
# [1] "Oracle"
# 
# $db.version
# [1] "19.00.0000"
# 
# $username
# [1] ""
# 
# $host
# [1] ""
# 
# $port
# [1] ""
# 
# $sourcename
# [1] "xxx" # removed for privacy
# 
# $servername
# [1] "xxx" # removed for privacy
# 
# $drivername
# [1] "RStudio Oracle ODBC Driver"
# 
# $odbc.version
# [1] "03.52"
# 
# $driver.version
# [1] "2.0.2.1002"
# 
# $odbcdriver.version
# [1] "03.80"
# 
# $supports.transactions
# [1] TRUE
# 
# $getdata.extensions.any_column
# [1] TRUE
# 
# $getdata.extensions.any_order
# [1] TRUE
# 
# attr(,"class")
# [1] "Oracle"      "driver_info" "list"  
```

### Reproducible Example
```r
library(DBI)

con_roracle <- dbConnect(ROracle::Oracle(), dbname = "xxx")
system.time(dbListTables(con_roracle))
# utilisateur     système      écoulé 
#       0.003       0.000       0.116 
system.time(dbWriteTable(con_roracle, "iris", iris))
# utilisateur     système      écoulé 
#       0.001       0.000       0.138 

conn_odbc_rstudio <- dbConnect(odbc::odbc(), dsn = "yyy")
system.time(dbListTables(conn_odbc_rstudio))
# utilisateur     système      écoulé 
#       2.430       1.676     482.881 
system.time(dbWriteTable(conn_odbc_rstudio, "iris", iris))
# utilisateur     système      écoulé 
#       9.451       1.367    3240.433 
[...]

PR #560, opened 2023-04-15
dbExistsTable: Specialize for oracle

detule:
Hi:

This should improve `dbWriteTable` performance when writing to an `Oracle` connection.  ( See: https://github.com/r-dbi/odbc/issues/540 )

Notes:
* Ideally would have liked to insert this query based approach for Oracle in `dbListTables`.  However, for `dbListTables("OdbcConnection"`, we advertise pattern based argument matching which is specific to the `SQLTables`/`connection_sql_tables` API; can bend the query to also do something similar but it's a bit more involved so punted for now.
* Outstanding are connection pane performance issues for `Oracle` which are also related to usage of `connection_sql_tables`.  Have an idea how to mitigate at least some of them ( listing catalogs, schemas, etc ).  Will address subsequently.
* Also in this PR, moved one of the specializations of `dbExistsTable` from `Connection.R` to `Table.R`, together with the remaining specializations.

### DIFF
--- a/R/Connection.R
+++ b/R/Connection.R
@@ -337,17 +337,6 @@ setMethod(
       table_type = table_type)$table_name
   })
 
-#' @rdname OdbcConnection
-#' @inheritParams DBI::dbExistsTable
-#' @export
-setMethod(
-  "dbExistsTable", c("OdbcConnection", "character"),
-  function(conn, name, ...) {
-    stopifnot(length(name) == 1)
-    df <- connection_sql_tables(conn@ptr, table_name = name)
-    NROW(df) > 0
-  })
-
 #' @inherit DBI::dbListFields
 #' @inheritParams DBI::dbListFields
 #' @aliases dbListFields
--- a/R/Table.R
+++ b/R/Table.R
@@ -218,6 +218,18 @@ setMethod(
 #' @inheritParams DBI::dbExistsTable
 #' @export
 setMethod(
+  "dbExistsTable", c("OdbcConnection", "character"),
+  function(conn, name, ...) {
+    stopifnot(length(name) == 1)
+    df <- connection_sql_tables(conn@ptr, table_name = name)
+    NROW(df) > 0
+  })
+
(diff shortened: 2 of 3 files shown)


detule:
Just a note.  Was trawling the ORACLE server side logs, and figured out that our calls to `SQLTables` gets translated to something that looks like:
```
SELECT * FROM
  ( SELECT null table_qualifier, o1.owner table_owner, o1.object_name table_name, decode(o1.owner,'SYS', decode(o1.object_type,'TABLE','SYSTEM TABLE','VIEW', 'SYSTEM VIEW', o1.object_type), 'SYSTEM', decode(o1.object_type,'TABLE','SYSTEM TABLE','VIEW', 'SYSTEM VIEW', o1.object_type), o1.object_type) table_type, null remarks
    FROM all_objects o1 WHERE o1.object_type in ('TABLE', 'VIEW')
    UNION
    SELECT null table_qualifier, s.owner table_owner, s.synonym_name table_name, 'SYNONYM' table_type, null remarks
    FROM all_objects o3, all_synonyms s
    WHERE o3.object_type in ('TABLE','VIEW') AND s.table_owner= o3.owner AND s.table_name = o3.object_name
    UNION
    SELECT null table_qualifier, s1.owner table_owner, s1.synonym_name table_name, 'SYNONYM' table_type, null remarks
    FROM all_synonyms s1 WHERE s1.db_link IS NOT NULL
  ) tables
WHERE 1=1 AND table_owner='SA' ORDER BY 4,2,3
```

This doesn't look all too dis-similar compared to the custom dbExistsTable implementation in this patch:
*  Looks like it includes synonyms.  Probably worth looking into including those if we can make this query performant.  Which brings me to second point.
*  Not sure why no subsetting is done within each of the UNION queries? Only filtering after the union?

detule:
Abandoning in favor of more comprehensive solution [here](https://github.com/r-dbi/odbc/pull/577)

PR #569, opened 2023-04-24
Use CRAN DBI test

hadley:


hadley:
This fixes the build failures which come about from some version conflict in dev DBItest.

detule (approved):

Issue #554, opened 2023-04-01
S4 method cleanup


detule:
Consolidate calls to:

* connection_sql_columns, and
* connection_sql_tables

in S4 methods.  The former already has a an S4 method, however we still call it directly outside the S4 method.  The latter needs an S4 method, and all internal callers ( there are several ) need to be transitioned to call the S4 method.

That way, if a particular DBMS needs an adjusted `connection_sql_tables` call, for example, we can specialize the S4 method and all callers will benefit.

```
$ grep -r --include=\*.R --exclude=RcppExports.R 'connection_sql_tables'
R/db.R:      connection_sql_tables(conn@ptr, ...)$table_name)
R/Connection.R:    connection_sql_tables(conn@ptr,
R/Connection.R:    df <- connection_sql_tables(conn@ptr, table_name = name)
R/Viewer.R:  table_types <- string_values(connection_sql_tables(connection@ptr, "", "", "", "%")[["table_type"]])
R/Viewer.R:  schemas <- string_values(connection_sql_tables(connection@ptr, "", "%", "", "")[["table_schema"]])
R/Viewer.R:  catalogs <- string_values(connection_sql_tables(connection@ptr, "%", "", "", "")[["table_catalog"]])
R/Viewer.R:    catalogs <- string_values(connection_sql_tables(connection@ptr, catalog_name = "%", "", "", NULL)[["table_catalog"]])
R/Viewer.R:    schemas <- string_values(connection_sql_tables(connection@ptr, "", "%", "", NULL)[["table_schema"]])
R/Viewer.R:  objs <- tryCatch(connection_sql_tables(connection@ptr, catalog, schema, name, table_type = type), error = function(e) NULL)
R/Viewer.R:            tables <- connection_sql_tables(connection@ptr)
R/Table.R:    name@name[["table"]] %in% connection_sql_tables(conn@ptr,
```
```
$ grep -r --include=\*.R --exclude=RcppExports.R 'connection_sql_columns'
R/Connection.R:    connection_sql_columns(conn@ptr,
R/Connection.R:    connection_sql_columns(conn@ptr,
R/Viewer.R:  cols <- connection_sql_columns(connection@ptr,
```

detule:
Fixed in https://github.com/r-dbi/odbc/pull/577

PR #577, opened 2023-06-09
API refactor + Oracle fixes

### LINKED ISSUE #575: Oracle: odbcConnectionColumns no-op
```r
> odbcConnectionColumns(conn, DBI::Id(schema="SA", table="TEST_TABLE"))
Error in connection_sql_columns(conn@ptr, table_name = name, catalog_name = catalog_name,  :
   nanodbc/nanodbc.cpp:2809: 00000: [Oracle][ODBC]Invalid SQL data type <-25>.
```

detule:
Hi:

Fixes #575 , #540, #554

Some highlights:
- Previously we had runaway usage of `connection_sql_tables`, which accommodates "magic" parameter ( combinations ).  This makes it difficult to customize parts of its functionality for individual back-ends.  Now some of the magic use cases are split out into `odbcConnectionCatalogs`, `odbcConnectionSchemas`, and `odbcConnectionTableTypes`.

- Specialize some of these for Oracle so that (hopefully) all of these are more performant ( and functional ): listing tables, writing to tables, and Rstudio/Viewer operations.


R/Viewer.R
```diff
@@ -84,7 +84,8 @@ odbcListObjects.OdbcConnection <- function(connection, catalog = NULL, schema =
   # if no catalog was supplied but this database has catalogs, return a list of
   # catalogs
   if (is.null(catalog)) {
-    catalogs <- string_values(connection_sql_tables(connection@ptr, catalog_name = "%", "", "", NULL)[["table_catalog"]])
+    catalogs <- tryCatch(
+      string_values(odbcConnectionCatalogs(connection)), error = function(e) vector(mode = "character"))
```
hadley, line 88:
```suggestion
    catalogs <- tryCatch(
      string_values(odbcConnectionCatalogs(connection)), 
      error = function(err) character()
    )
```

R/Viewer.R
```diff
@@ -346,12 +347,12 @@ odbcConnectionActions.default <- function(connection) {
               tableName <- dbQuoteIdentifier(connection, firstTable$table_name)
 
               # add schema
-              if (!is.null(firstTable$table_schema) && nchar(firstTable$table_schema) > 0) {
+              if (!is.null(firstTable$table_schema) && !is.na(firstTable$table_schema) && nchar(firstTable$table_schema) > 0) {
```
hadley, line 350: This level of type flexibility always makes me a bit nervous. Could we add some type checks higher up so we only need to check for one empty sentinel?

R/db.R
```diff
@@ -14,6 +14,76 @@ setMethod("sqlCreateTable", "Oracle",
...
+  "odbcConnectionTables",
+  c("Oracle", "character"),
+  function(conn, name, catalog_name = NULL, schema_name = NULL, table_type = NULL) {
+
+    qTable <- getSelector("object_name", name)
+    if (is.null(schema_name)) {
+      query <- paste0(" SELECT null AS \"table_catalog\", '", conn@info$username ,"' AS \"table_schema\", object_name AS \"table_name\", object_type AS \"table_type\", null AS \"table_remarks\"",
+                      " FROM user_objects ",
+                      " WHERE 1 = 1 ", qTable,
+                      " AND ( object_type = 'TABLE' OR object_type = 'VIEW' ) ")
```
hadley, line 33:
```suggestion
      query <- paste0(
         " SELECT null AS \"table_catalog\", '", conn@info$username ,"' AS \"table_schema\", object_name AS \"table_name\", object_type AS \"table_type\", null AS \"table_remarks\"",
         " FROM user_objects ",
         " WHERE 1 = 1 ", qTable,
         " AND ( object_type = 'TABLE' OR object_type = 'VIEW' ) "
      )
```

etc

R/Connection.R
```diff
@@ -191,6 +195,184 @@ setMethod(
...
+    connection_sql_tables(conn@ptr,
+      catalog_name = catalog_name,
+      schema_name = schema_name,
+      table_name = name,
+      table_type = table_type)
+
+  }
+)
+
+setMethod(
```
hadley, line 271: Missing `@rdname`?

R/Connection.R
```diff
@@ -191,6 +195,184 @@ setMethod(
...
+#' @param conn OdbcConnection
+#' @param catalog_name Catalog where
+#' we are looking to list schemas.
+#' @rdname odbcConnectionSchemas
+setGeneric(
+  "odbcConnectionSchemas",
+  valueClass = "character",
+  function(conn, catalog_name) {
+    standardGeneric("odbcConnectionSchemas")
+  }
```
hadley, line 332:
```suggestion
  function(conn, catalog_name = NULL) {
    standardGeneric("odbcConnectionSchemas")
  }
```

Does that allow you to avoid the new `missingOrNULL` class union?

hadley (approved):

R/db.R
detule: Done - thanks!

R/Connection.R
detule: Good idea - done / thanks!

R/Viewer.R
detule: Thanks Hadley:

As best as I can tell, the `is.null(...)` that was there to begin-with is not guarding anything; I think only the `is.na` is actually effective.

Kind of hesitant to remove the `is.null` since I am not sure why it was there in the first place.  But at the same time this is not in any critical code-path; we can revisit if remove it breaks something.

Issue #586, opened 2023-07-07
Error with `dbWriteTable()` to an Oracle database: `Not compatible with requested type: [type=character; target=double].`


ThomasSoeiro:
### Issue Description and Expected Result

Since `package:odbc` 1.3.5 we get the following error when trying to write a table from R to an Oracle database: `Erreur : Not compatible with requested type: [type=character; target=double].` The error occurs with both RStudio ODBC driver and with Oracle ODBC driver.

The error was not present with `package:odbc` 1.3.3 on the same system (but there was a performance issue that is now fixed, see #540).

### Database

```r
dbGetInfo(con)
## $dbname
## [1] ""
## 
## $dbms.name
## [1] "Oracle"
## 
## $db.version
## [1] "19.00.0000"
## 
## $username
## [1] ""
## 
## $host
## [1] ""
## 
## $port
## [1] ""
## 
## $sourcename
## [1] "xxx"
## 
## $servername
## [1] "xxx"
## 
## $drivername
## [1] "RStudio Oracle ODBC Driver"
## 
## $odbc.version
## [1] "03.52"
## 
## $driver.version
## [1] "2.0.2.1002"
## 
## $odbcdriver.version
## [1] "03.80"
## 
## $supports.transactions
## [1] TRUE
## 
## $getdata.extensions.any_column
## [1] TRUE
## 
## $getdata.extensions.any_order
## [1] TRUE
## 
## attr(,"class")
## [1] "Oracle"      "driver_info" "list" 
```

### Reproducible Example

```r
library(DBI)
con <- dbConnect(odbc::odbc(), dsn = "xxx")
DF <- data.frame(letters, num = 1:26)
dbWriteTable(con, "DF", DF)
## Erreur : Not compatible with requested type: [type=character; target=double].
dbReadTable(con, "DF")
## [1] letters num    
## <0 lignes> (ou 'row.names' de longueur nulle)
```

<details>
<summary>Session Info</summary>

```r
## R version 4.1.2 (2021-11-01)
## Platform: x86_64-pc-linux-gnu (64-bit)
## Running under: Red Hat Enterprise Linux Server 7.9 (Maipo)
## 
## Matrix products: default
## BLAS/LAPACK: /usr/lib64/libopenblasp-r0.3.3.so
## 
## locale:
[...]

PR #587, opened 2023-07-08
odbc_write_table: describe parameters only if column data is coherent

detule:
Fixes https://github.com/r-dbi/odbc/issues/586

[When writing out ](https://github.com/r-dbi/odbc/commit/fc10a195f2b4558e636d62ff02a48b6569eadd3d)our oracle specific implementation of `odbcConnectionColumns` we snooped on how the OEM driver implements the [`SQLColumns`](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlcolumns-function?view=sql-server-ver16) ODBC API call.  As a result we inherited some issues - for example, the OEM query always returns `0` for the `DATA_TYPE` field in the result ( we return `NA` ).

This is a problem, because if the output of `odbcConnectionColumns/SQLColumns` is available, our `dbWriteTable` implementation will attempt to use it as a performance optimization / to reduce the number of round-trips to the server to describe each of the parameters of the `INSERT` query.

With this patch, we attempt to optimize ( pre-describe the parameters ) only if we have a coherent result from the `odbcConnectionColumns` query.  Coherent is defined as `data_type` is not `NA` and the number of rows ( each row describing an individual parameter ) is equal to the number of parameters in the `INSERT` query.

Beyond this patch, we should snoop on the SIMBA driver as well to see if it has a better / more complete way of querying column types of existing tables, and use this to update the ORACLE specific implementation.

Finally, probably also worth thinking about

* Adding ORACLE unit testing infrastructure.
* Adding MSSQL / FreeTDS unit testing infrastructure as for that DB/Driver pairing the parameter description is not just an optimization, it is a necessity.  It is important we don't disrupt how the parameter description works/is used in that case.


R/Table.R
```diff
@@ -70,7 +70,7 @@ odbc_write_table <-
     fieldDetails <- tryCatch({
       details <- odbcConnectionColumns(conn, name)
       details$param_index <- match(details$name, names(values))
-      details[!is.na(details$param_index), ]
+      details[!is.na(details$param_index) & !is.na(details&data_type), ]
```
hadley, line 73:
```suggestion
      details[!is.na(details$param_index) & !is.na(details$data_type), ]
```
?

hadley (approved):

R/Table.R
detule: Ugh - sorry about that.
Thanks!

hadley: Don't worry, spotting this sort of problem makes me feel useful 😄

Issue #588, opened 2023-07-13
[Teradata] `dbListTables` broken by odbc 1.3.5


But2ene:
### Issue Description and Expected Result

In Teradata method for dbListTables, the argument `table_name` is not renamed into the argument `name` expected by the new function `odbcConnectionTables`.  It was caused by #577.
See code in https://github.com/r-dbi/odbc/blob/main/R/db.R#L108C1-L113C5

### Database

Teradata 17.20

### Reproducible Example

```r
con <- DBI::dbConnect(odbc::odbc(), dsn = "my_teradata_dsn")

DBI::dbListTables(con, table_name = "my_table")
```

The code works with odbc 1.3.4 and fails with odbc 1.3.5
```
Error in .local(conn, name, ...) : 
  argument not used (table_name = "my_table")
```

But2ene:
Thank you for the quick answer, I'll check next week if it works.
Actually, I wish the Teradata `dbListTables` method to be deleted and replaced by a Teradata specific `OdbcConnectionTables` method. There is an ongoing issue with Teradata `dbExistsTable` method, it can not detect temporary tables. Sharing the same code as dbListTables could fix this problem!

PR #589, opened 2023-07-13
db: Teradata: dbListTables forwards arguments correctly

detule:
Fixes https://github.com/r-dbi/odbc/issues/588

### DIFF
--- a/R/db.R
+++ b/R/db.R
@@ -107,9 +107,14 @@ setMethod("sqlCreateTable", "Teradata",
 
 setMethod(
   "dbListTables", "Teradata",
-  function(conn, ...) {
+  function(conn, catalog_name = NULL, schema_name = NULL, table_name = NULL,
+    table_type = NULL, ...) {
     c(dbGetQuery(conn, "HELP VOLATILE TABLE")[["Table SQL Name"]],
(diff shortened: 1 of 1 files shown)


detule:
Tested: I set up a Teradata instance locally and had a chance to verify it works.

R/db.R
```diff
@@ -107,9 +107,14 @@ setMethod("sqlCreateTable", "Teradata",
 
 setMethod(
   "dbListTables", "Teradata",
-  function(conn, ...) {
+  function(conn, catalog_name = NULL, schema_name = NULL, table_name = NULL,
+    table_type = NULL, ...) {
     c(dbGetQuery(conn, "HELP VOLATILE TABLE")[["Table SQL Name"]],
-      odbcConnectionTables(conn, ...)$table_name)
+      odbcConnectionTables(conn,
```
hadley, line 113: Might be worth pulling this out into a variable since it's a bit hard to parse currently.

hadley (approved):

R/db.R
detule: Agreed.

Sounds like we may need to circle back to revisit some of the other S4/Teradata extensions.  I'll revisit the syntax here then.

PR #590, opened 2023-07-19
refacto of odbcConnectionTables for teradata

But2ene:
Alternative fix for #588

### DIFF
--- a/R/db.R
+++ b/R/db.R
@@ -111,23 +111,37 @@ setMethod(
   c("Teradata", "character"),
   function(conn, name, catalog_name = NULL, schema_name = NULL, table_type = NULL) {
 
-    schema_table_list <- connection_sql_tables(
-      conn@ptr,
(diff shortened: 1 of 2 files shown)


detule:
Hi - thanks for the PR:

I like where you are headed with this.  It may need a few touch ups:

* We should make sure to retain the output format of `odbcConnectionTables` as described in the documentation of the generic.
* We should make sure to respect the `name` argument.  So if present, we should subset `temporary_table_list` appropriately.
  * This should also account for when `name` is pattern value ( use utils.R::isPatternValue to check )

  I'll go ahead and merge the fix to the regression for now.

detule:
Hi @But2ene - nice work again with the PR.

I made a small update - does it still work for your use case?

detule:
Ping @But2ene:

Let me know if you've had a chance to test this as it sits.

Thanks!

detule (approved):

PR #591, opened 2023-07-21
Add odbcSetMetadataId

fh-afrachioni:
Adds `odbcSetMetadataId(conn, val)`, as a thin wrapper around
```
SQLSetConnectAttr(&conn, SQL_ATTR_METADATA_ID, (SQLPOINTER)val, 0);
```

As in the [ODBC API docs](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlsetconnectattr-function), this affects the behavior of all catalog functions on the connection, by selecting whether the string argument is treated as an identifier or a search pattern.

This can have substantial performance implications in data warehousing environments with many tables, where "search pattern" queries can get expensive if wildcards are not escaped.


fh-afrachioni:
@hadley Any chance the team would be willing to consider this contribution within the next month or two?

### Some context
We're thinking about what it would take to get `compute()` / `copy = TRUE` behavior working with Snowflake warehouses that include many objects.  One of the barriers we identified when prototyping was the high cost of `odbc_write_table()` confirming the temporary table name isn't already defined, largely due to the unwanted handling of pattern matching (see the "Further" header [here](https://github.com/tidyverse/dbplyr/issues/1099)).

The other barrier was picking a catalog and schema name for the temporary table, when no default is provided by the backend.  While we recognize the [decision](https://github.com/tidyverse/dbplyr/issues/1099#issuecomment-1521840853) in April to not handle the "special case of Snowflake," we have seen renewed interest lately in workflows that rely on these features.

cc @fh-mthomson

detule:
Hi - thanks for the submission.

I am a bit weary of modifying the metadata id argument since for one, [some of our methods](https://github.com/r-dbi/odbc/blob/main/R/Connection.R#L497) are documented as supporting pattern value matching.  In addition, it opens another dimension when it comes to driver behavior and debugging issues.

Are there other approaches we could try?  `odbcConnectionColumns` ( as well as `dbExistsTable` ) is `S4` - presumably we could tailor it, within reason, for `Snowflake`, since it sounds like its performance can be an issue.  We recently did [something similar for Oracle.](https://github.com/r-dbi/odbc/blob/main/R/db.R#L92)

Can you share some examples of the problematic executions from the perspective of `package:odbc`.  Somethiing like
A. This is slow:
```
DBI::dbWriteTable( ... )
odbc::odbcConnectionColumns( ... )
```
B. This executes quickly:
```
DBI::dbWriteTable(...)
odbc::odbcConnectionColumns( ... )
```
We can't use B. because ....

Thanks

hadley:
We could also add a `allow_patterns` argument to `dbListTables()`?

detule:
Yep - good call.  But un/setting the METADATA_ID attribute for the connection would also affect the behavior of `odbcConnectionTables` - not exported, but used widely throughout `package:odbc`, as well as `odbcConnectionColumns` ( exported ).

More generally it affects a number of the [ODBC catalog functions](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/catalog-functions-in-odbc?view=sql-server-ver16).  So, say, if in the future we wanted to start using `SQLProcedures` to list sprocs in a database and show those in the RStudio Connections pane ( alongside tables and views ), or `SQLPrimaryKeys` to be able to highlight the primary key of a table - we would have to take into account behavior with the attribute set as well as without.

Having said that, not really opposed to having some means to manipulate the attribute, but thought it might be worth thinking about other options as well.  In addition to looking into whether we can do some of the catalog/schema extraction in a "`odbcConnectionColumns.Snowflake'" method, I can also look into what it would take to set this attribute on a per-statement basis ( rather than per connection ).

Once I think we understand the problem better, and have some examples we can test with, I can try and mock up what this would look like.

hadley:
I'm not that close to odbc, so it did surprise me that the table names match wild cards by default. It might be too late to change it now, but I think it would generally be less surprising if you had to opt-in to wildcard matching behaviour.

fh-afrachioni:
Thanks for the thoughtful discussion on this!  In case there's any ambiguity, I am not necessarily proposing to modify the default value of `METADATA_ID` set on new connections.  Rather, I would lean toward an argument that `odbc` should respect the specification-specified behavior of any driver, in the spirit of its role as a wrapper to that interface.

I also agree with Hadley that, for R users,
> it would generally be less surprising if you had to opt-in to wildcard matching behaviour.

which does present a conflict, simply because the default in the specification is surprising.

I think adding a well-advertised argument to relevant exported functions might be a decent compromise; setting this at a statement level [seems straightforward enough](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlsetstmtattr-function?view=sql-server-ver16).  This would also enable tools like RStudio and `dbplyr` to specify intended behavior without affecting subsequent user calls.

I don't think this issue is related to Snowflake or motivates Snowflake-specific logic, but presents itself when a sufficiently large number of names in a catalog or database make the performance implications of pattern-matching obvious.

detule:
Thanks for the feedback.

When you get a chance, can you send us the `package::odbc` formulated fast / not fast examples?

Setting the attribute per statement is better than doing it per connection, but it's a bit more invasive into the nanodbc codebase.

I'll look into what it would take to upstream those changes, but also want to explore some other angles - especially if passing a `DBI:Id` is already performant.

Cheers.

detule:
ping @fh-afrachioni - when you get a chance, do try and jot down some fast/not-fast examples.  Those, i think will be very helpful.

Thanks!

fh-afrachioni:
> some fast/not-fast examples

Again, the impact of this issue will depend on the local environment (how many other names do we need to check to confirm the table doesn't exist already?)

I've been testing with the following call to `dbWriteTable()`, which simply writes `cars` to the database as a temporary table at some specified location:
```
DBI::dbWriteTable(
  snowflake_conn,
  DBI::Id(catalog = '<redacted_db>', schema = '<redacted_schema>', table = 'tmpName'),
  cars,
  temporary = TRUE, overwrite = TRUE
)
```

Using the `main` version of `odbc` (or when explicitly setting `odbc::odbcSetMetadataId(snowflake_conn, FALSE)`, the default), this takes about **180 seconds**.

If I call `odbc::odbcSetMetadataId(snowflake_conn, TRUE)` prior to `dbWriteTable()`, this takes about **2.4 seconds**.

With respect to dropping the `_` from temp table names: this does seem to have a modest effect on performance.  But importantly, [pattern matching is also used at the schema level](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/arguments-in-catalog-functions?view=sql-server-ver16) when `SQL_ATTR_METADATA_ID` is `FALSE`.  So if one happens to have a current schema with an underscore in its name, attempting to create a temporary table there using `dbplyr` will cause the entire catalog to be searched.

I plan to do some further thinking about this at the `dbplyr` level in the coming days.  I'm also currently seeing `Invalid use of NULL pointer for catalog name` when trying `copy = TRUE` with the development version of `odbc` (I'll file a separate issue).

detule:
Really appreciate the feedback.

And the example is very helpful - it helps frame the problem: I thought when using a DBI:Id the situation was better than when not - some of the issues referenced above seem to focus on the character - to - Id tranformation and in particular identifying schema etc.

Is that not the case?

fh-afrachioni:
Ok, I've learned more about the error message I was seeing.  When I try something like
```
odbc::odbcSetMetadataId(snowflake_tbl$src$con, TRUE)
left_join(snowflake_tbl, local_tbl, copy = TRUE)
```
I get
```
Invalid use of NULL pointer for schema name.
```

I have concluded this is due to `dbplyr` is calling `dbWriteTable()` without specifying a schema or catalog ([here](https://github.com/tidyverse/dbplyr/blob/93e8bcbb8ab7dbc07a99af676bb1d7758e9c68f1/R/verb-copy-to.R#L80) is where `db_copy_to()` is called, within `copy_to.src_sql()`, where `name` will be a plain string from `unique_table_name()` in the context of `copy = TRUE`).  As it turns out, different behavior is specified for null arguments to catalog functions depending on the state of `METADATA_ID`:

If `METADATA_ID` is `FALSE` (default), the catalog, schema, and table name arguments to `SQLTables` are "pattern value" type, [for which](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/pattern-value-arguments?view=sql-server-ver16):
> Passing a null pointer to a search pattern argument does not constrain the search for that argument; that is, a null pointer and the search pattern % (any characters) are equivalent.

If `METADATA_ID` is `TRUE`, the catalog, schema, and table name arguments to `SQLTables` are "identifier" type, [for which](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/identifier-arguments?view=sql-server-ver16):
> Setting an identifier argument to a null pointer returns SQL_ERROR and SQLSTATE HY009 (Invalid use of null pointer), unless the argument is a catalog name and catalogs are not supported.

I admit this breaks many more things than I realized, especially considering `NULL` is a prevalent default for catalog and schema name.  I suppose we could work to handle the case throughout `odbc`, but this might introduce both cost and risk to a meaningful extent.

My humble proposal:
* Pivot from this change exposing the connection-level flag, to [`SQLSetStmtAttr`](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlsetstmtattr-function?view=sql-server-ver16), enabling us to implement something like `dbExistsTable(..., metadata_id = FALSE)` by setting the flag on a statement before execution (and with attention to null arguments when we do).  Then we can pin it `TRUE` in the call from `dbWriteTable()`, as we never care about pattern matching when confirming a specific name does not already exist.
* Update `dbplyr` to use a fully qualified name when specifying the location of a temporary table.  I'm still unsure what catalog and schema we would chose when the backend has no defaults, but we can pick up that topic in another thread.

detule:
Thanks for thinking through the various corner cases.  I said at the onset I am not opposed to using `SQLSetStmtAttr` - we usually upstream any changes we make to `nanodbc` and try not to depart from the main trunk too much.  We can do that if it proves to be the best option.  I am a little annoyed that we would need to make a few round-trips to the server, and those are not free, but we can cross that bridge if/when we get there.

In the meantime, do you have the ability to test https://github.com/detule/odbc/tree/experiment/snowflake ?
It is just about completely untested, but could provide a foundation for a relatively low cost solution, maybe.

I would expect some performance improvements - but not the full benefit, as we still need to specialize at least `odbcConnectionColumns`.  We can do that if this approach has some merit.

Cheers

detule:
Hey - thanks for iterating!

Good to hear this works.   Some of the other points in order:

* `force_exact` + `odbcConnectionColumns` -- at this point, don't think that would be very effective since the generic `odbcConnectionColumns` doesn't parse this argument.  I do think you are on the right track - we would need to write Snowflake specific method that queries `INFORMATION_SCHEMA.COLUMNS`.  To the extent that we believe `odbcConnectionColumns` is the culprit for the performance issues, i guess not too surprising you are not seeing much of a performance benefit.  I can take a crack at this maybe this weekend unless we think it's pointless ( is querying INFORMATION_SCHEMA.COLUMNS snappy on your data/table/column rich database? )

* Upstreaming to `nanodbc`.  Generally we try and avoid direct calls to the `odbc` API in this package, and rather call wrapper functions from within `nanodbc`.  I guess there could be exceptions to this, but this approach has protected us / worked well for us historically.  Even if we do go with the pull request here, I would ask if it makes sense to fold it into `nandobc`.  At any rate, it's not a big hurdle - we have a good relationship with the maintainers there and they have historically been very responsive to feature requests we have brought up.

* Whether `SQLSetStatrementAttr` roundtrips: In the ODBC spec, one of the return codes is `Connection timeout expired` ( and another is `Communication link failure` ) which made me think the API leaves the door open for the driver to check in with the data source.  No idea if the `Snowflake` driver actually does this.

Thanks again for the thoughtful approach and experimentation!

fh-afrachioni:
Thanks again for the collaboration!  I'm seeing an increased level of interest in use cases that would benefit from this, so I'm excited to spend some time pushing on it.  In order:
* Yes, sorry, I see now we can't call the generic with a Snowflake-specific argument.  Simple queries of the `INFORMATION_SCHEMA` generally do take a few seconds to execute, not what I'd call "snappy" but also better than we're seeing in your experiment (I think due to `LIKE` being used with names that include underscores).  To investigate whether `odbcConnectionColumns()` is "the culprit" here, I did a bit more testing:
  * A: `odbcConnectionColumns(snowflake_conn, 'noUnderscoreTbl', 'noUnderscoreCatalog', 'noUnderscoreSchema', 'noUnderscoreColumn')`
  * B: `odbcConnectionColumns(snowflake_conn, 'underscore_tbl', 'underscore_catalog', 'underscore_schema', 'underscore_column')`
  * C: `odbcConnectionColumns(snowflake_conn, 'underscore\\_tbl', 'underscore\\_catalog', 'underscore\\_schema', 'underscore\\_column')`

  Under the conditions
  * 1: Release version of `odbc` (1.3.5)
  * 2: Version of `odbc` corresponding to this pull request (`fh-afrachioni:add_set_metadata_id`), without any call to `odbcSetMetedataId()`
  * 3: Same as 2, but following a call to `odbcSetMetedataId(snowflake_conn, TRUE)`

  I see
  * `1A: FAST, 1B: SLOW, 1C: FAST`
  * `2A: FAST, 2B: SLOW, 2C: FAST`
  * `3A: FAST, 3B: FAST, 3C: FAST`

  Where `FAST` means "reliably less than three seconds, typically about one second," and `SLOW` means "reliably more than two minutes."  I think we can conclude `odbcConnectionColumns()` _is_ the culprit, and we have two possible mitigations: set `METADATA_ID` on the connection or statement (as we've been discussing), OR escape pattern matching characters.

* Thanks for the information!  I didn't realize there was such a close relationship between these packages.  Sounds like it won't be a strong barrier to upstreaming the change, if we decide to make it.
* Good catch on the return code -- I'm still not sure how much of a performance impact the extra call is (at least compared to calling catalog functions when pattern matching is not needed, without setting `METADATA_ID`).

detule:
Thanks for the detective work @fh-afrachioni - nice job zeroing in on `odbcConnectionColumns`.

Do you mind giving https://github.com/detule/odbc/tree/experiment/snowflake one more try;  wrote a `snowflake` specific `odbcConnectionColumns`.  There are some rudimentary performance logs just to make life a bit easier.

There are a few ways it can probably be made faster; at this point not going for any records - but hopefully we can improve ( dramatically ) on 180 seconds / make writing usable.

fh-afrachioni:
Thanks for logging the timing; this is great!  Here's a typical (non-first*) run, using the same test conditions as above:
```
Using custom Snowflake tables method
Using custom Snowflake columns method
Total time:            3.384053
	[R] work:      0.000115
	Exists:        0.880887
	[R] work:      0.000009
	Remove work:   0.099634
	[R] work:      0.000489
	Create work:   0.183451
	Columns work:  1.163285
	Result work:   0.083483
	Describe work: 0.000101
	Write work:    0.972600
```

*note this takes 15-20 seconds on the first run in a new R session (not so for a new connection within a session), mostly in "create work" with some "Exists".  Non-first runs are reliably within 3 or 4 seconds.  I'm not really sure why the backend needs "warming up," but I think we can live with it.

This also successfully leads to `FAST` execution of tests `A`, `B`, and `C`.

Thanks again for putting together this prototype!  I also wonder what your thoughts are about programmatically escaping `_` and `%` in `odbcConnectionColumns(..., force_exact = TRUE)`, prior to a call to `connection_sql_columns()`: the query to `INFORMATION_SCHEMA` clearly does the job, though I wonder about the complexity of the inlined query from a bug-surface-area point of view.

detule:
Thanks - I don't think we can escape all wildcards in `odbcConnectionColumns` since the invocation in `odbc_write_table` asks for column definitions for all columns in the table, without knowing what the columns are.  Therefore we need to be able to pass a wild card in the column name slot at least.  For what is worth, this is also why setting the `METADATA_ID` attribute whether at the connection or at the statement level is not really a viable solution.

It could be that just escaping the wild cards in the catalog, schema, and table names increases performance significantly ( kind of surprising, but then again I have given up a bit on being surprised with `Snowflake`; how is there not a `primary key` or another index on `INFORMATION_SCHEMA.COLUMNS`?).

Let me know if that's the case.  If it is ( we can bridge this issue just with escaping wild cards in catalog/schema/table name), do you still have an issue in/with `dbplyr` with inferring the current catalog/schema? 

Thanks for your input / looking forward to wrapping this up.

fh-afrachioni:
I think we can leave the `column_name` argument `NULL` (which is what's currently done), as
> [Passing a null pointer to a search pattern argument does not constrain the search for that argument; that is, a null pointer and the search pattern % (any characters) are equivalent.](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/pattern-value-arguments?view=sql-server-ver16)

---

> It could be that just escaping the wild cards in the catalog, schema, and table names increases performance significantly

It *is* significantly faster!  I just confirmed on the release `odbc` that "test `D`" are just as fast as "test `C`" above with escaped `column_name` (at least to the current back-of-envelope degree of rigor):
* D (i): `odbcConnectionColumns(snowflake_conn, 'underscore\\_tbl', 'underscore\\_catalog', 'underscore\\_schema', NULL)`
* D (ii): `odbcConnectionColumns(snowflake_conn, 'underscore\\_tbl', 'underscore\\_catalog', 'underscore\\_schema', '%')`

So I think we can get the performance benefits of avoiding pattern matching without needing to inline the `INFORMATION_SCHEMA` query, which is a bit slower and may introduce some level of risk.

detule:
Can you verify that you are actually getting correct/coherent results when you escape underscores?

```
> nrow(odbc::odbcConnectionColumns(conn, catalog_name='TEST\\_DB', schema_name='TEST\\_SCHEMA', name = "TYPETEST"))
[1] 0
> nrow(odbc::odbcConnectionColumns(conn, catalog_name="TEST_DB", schema_name="TEST_SCHEMA", name = "TYPETEST"))
[1] 26
```

fh-afrachioni:
Yes, I am!  I get the expected result, a table with one row for each column in the named table, with columns including `table_name`, `schema_name`, `field.type`, etc.

Without knowing the specifics of your test database, could you be matching more names than you're targeting?  For example, is `TEST_SCHEMA` getting you `TESTXSCHEMA` or similar, when a literal `TEST_SCHEMA` does not exist?

detule:
We can't escape any underscores in the `catalog_name` argument since, [per the spec ](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/arguments-in-catalog-functions?view=sql-server-ver16) it is _**not**_ treated as a pattern value argument regardless of how the `METADATA_ID` attribute is set ( for `SQLColumns`, different than how it is treated for `SQLTables` ).

I am a bit surprised it works with the underscores escaped in the catalog name on your end; that bit of inconsistency stinks.  At any rate, I'll craft a Snowflake / S4 method that escapes the underscores in the schema and table name arguments and wrap this up.  Before you ask, if someone uses "%" in the schema/table name, I think they deserve a long execution time.

Finally, can you confirm that in your use cases `dbplyr` passes an identifier with the catalog and schema name as well?

```
> DBI::dbExecute(conn, "CREATE DATABASE new_test_db")
[1] 0
> DBI::dbExecute(conn, "USE DATABASE new_test_db")
[1] 0
> DBI::dbExecute(conn, "CREATE SCHEMA new_test_schema")
[1] 0
> DBI::dbExecute(conn, "USE SCHEMA new_test_schema")
[1] 0
> dbWriteTable(conn, value=mtcars, name = "t_example_mtcars", row.names=FALSE)
> odbc:::odbcConnectionTables(conn, catalog_name = "NEW_TEST_DB", schema_name = "NEW_TEST_SCHEMA" )
   table_catalog    table_schema       table_name table_type table_remarks
1   NEW_TEST_DB NEW_TEST_SCHEMA t_example_mtcars      TABLE
> odbc:::odbcConnectionTables(conn, catalog_name = "NEW\\_TEST\\_DB", schema_name = "NEW\\_TEST\\_SCHEMA" )
   table_catalog    table_schema       table_name table_type table_remarks
1   NEW_TEST_DB NEW_TEST_SCHEMA t_example_mtcars      TABLE
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW_TEST_DB", schema_name = "NEW_TEST_SCHEMA", name = "t_example_mtcars" ) )
[1] 11
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW\\_TEST\\_DB", schema_name = "NEW\\_TEST\\_SCHEMA", name = "t\\_example\\_mtcars" ) )
[1] 0
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW\\_TEST\\_DB", schema_name = "NEW\\_TEST\\_SCHEMA", name = "t_example_mtcars" ) )
[1] 0
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW\\_TEST\\_DB", schema_name = "NEW_TEST_SCHEMA", name = "t_example_mtcars" ) )
[1] 0
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW_TEST_DB", schema_name = "NEW_TEST_SCHEMA", name = "t\\_example\\_mtcars" ) )
[1] 11
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW_TEST_DB", schema_name = "NEW_TEST_SCHEMA", name = "%" ) )
[1] 11
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW_TEST_DB", schema_name = "NEW\\_TEST\\_SCHEMA", name = "t\\_example\\_mtcars" ) )
[1] 11
> nrow( odbc::odbcConnectionColumns(conn, catalog_name = "NEW\\_TEST\\_DB", schema_name = "NEW\\_TEST\\_SCHEMA", name = "t\\_example\\_mtcars" ) )
[1] 0
```

fh-afrachioni:
Ah I'm sorry, the non-redacted catalog name I was using did not have any underscores, and in my effort to demonstrate that I was escaping underscores, I added one to the example catalog.  You're right -- catalog is an "ordinary argument" to `SQLColumns` (but not `SQLTables`, curiously).

I can confirm the behavior you're seeing with `dbWriteTable()`, `odbcConnectionTables()`, and `odbcConnectionColumns()` matches what I'm seeing.  I'm not sure what role `dbplyr` has in passing the catalog and schema names though; it looks to me like this "current" behavior is driven by the backend, with `name` in `odbc::odbc_write_table()` passed directly to `dbExistsTable()` and friends as a plain string.

Fully agreed on this by the way
> if someone uses "%" in the schema/table name, I think they deserve a long execution time.

Thanks for taking on this implementation!  I'm excited we're finally near the line here.

---
Eyeing the prize of `copy = TRUE`, it may be relevant that if I try (without first `USE DATABASE` or `USE SCHEMA`)
```
  DBI::dbWriteTable(
    snow_conn,
    value=mtcars,
    name='CAT.REAL\\_SCHEMA.t_example_mtcars',
    row.names=FALSE, temporary=TRUE, overwrite=TRUE
  )
```
(or use the equivalent `DBI::Id`), I get
```
Cannot perform CREATE TEMPTABLE. This session does not have a current database. Call 'USE DATABASE', or use a qualified name. 
```
ostensibly from the call to `sqlCreateTable()`.  I think a subsequent challenge might be to support writing tables to a fully qualified location, and/or thinking about how `dbplyr` specifies where temporary tables should go.  We happen to have workflows here that lead to connections without current schema/catalog being common, though maybe this is not broadly common.

detule:
Thanks - I am worried about scope creep, and what you are describing reads like a separate problem than what we've been trying to tackle here.

Do you mind starting a new issue / happy to iterate on there.

In the meantime, I'll close this PR as I opened https://github.com/r-dbi/odbc/pull/618 with the approach we discussed here.  If you have a chance to try it out, please do. 

Cheers

fh-afrachioni:
Yup, will do!  Sorry to derail; I wanted to zoom out for a moment while we have some momentum.

PR #599, opened 2023-08-31
Modify `odbcDataType.Snowflake` to better reflect Snowflake Data Type…

meztez:
…s documentation

https://docs.snowflake.com/en/sql-reference/data-types

> When choosing the maximum length for a VARCHAR column, consider the following:
> 
>  - Storage: A column consumes storage for only the amount of actual data stored. For example, a 1-character string in a VARCHAR(16777216) column only consumes a single character.
> 
>  - Performance: There is no performance difference between using the full-length VARCHAR declaration VARCHAR(16777216) and a smaller length.

Since it is used to create tables in Snowflake, I do no see why (from above) it should be limited to 255.

Limiting it can introduce truncation bugs on the user end.

I am a firm believer in preventing end user potential bugs.


hadley (approved):

detule:
Thank you!

Issue #584, opened 2023-06-30
MSSQL: Cannot append to temporary tables


jatherrien:
### Issue Description and Expected Result
Unable to append to temporary table due to `dbExistsTable` being blind to temporary tables. Given that the table exists and can be queried from, I expect to be able to append to it.

May be related to #509 ?

### Database
Microsoft SQL Server 16.0.1000.6

### Reproducible Example

```r
library(odbc)
library(DBI)

con <- DBI::dbConnect(odbc::odbc(),
                                Driver = "ODBC Driver 18 for SQL Server",
                                SERVER = "<masked>",
                                Trusted_Connection="yes",
                                TrustServerCertificate="yes",
                                port = 1433)

dbExecute(con, "CREATE TABLE #my_temp (
               id int not null,
               x float not null,
               primary key (id)
               )", immediate=TRUE) # need immediate=TRUE for table to continue to exist
#> [1] 0

# We can verify that the table exists through select
dbGetQuery(con, "SELECT * FROM #my_temp")
#> [1] id x 
#> <0 rows> (or 0-length row.names)

dbAppendTable(con, "#my_temp", data.frame(id=1, x=5.0))
#> Error in dbAppendTable(con, "#my_temp", data.frame(id = 1, x = 5)): dbExistsTable(conn, name) is not TRUE
dbExistsTable(con, "#my_temp")
#> [1] FALSE
dbExistsTable(con, "#my_temp", immediate=TRUE)
#> [1] FALSE
```
It may be worth mentioning that I also experience this same issue on a Linux system as well.

<details>
<summary>Session Info</summary>

```r
devtools::session_info()
─ Session info ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.3.1 (2023-06-16 ucrt)
 os       Windows 11 x64 (build 22621)
 system   x86_64, mingw32
 ui       RStudio
 language (EN)
 collate  English_Canada.utf8
 ctype    English_Canada.utf8
 tz       America/Vancouver
 date     2023-06-30
 rstudio  2023.06.0+421 Mountain Hydrangea (desktop)
 pandoc   3.1.1 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)

─ Packages ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 package     * version date (UTC) lib source
[...]

PR #600, opened 2023-09-01
SQL Server: Index local temp tables

detule:
Fixes https://github.com/r-dbi/odbc/issues/584

With this, `dbExistsTable( conn, "#tempname")` should be reliable.  In turn writing to local temp tables with `dbWriteTable` or `dbAppendTable` should also be improved.

I am tracking additional issues with  temp tables in SQL Server.  Will address separately.

As I was working on this I was reminded that we need to add a DB/workflow using FreeTDS to CI/CD ( mostly a note to myself ).

Thanks!


R/db.R
```diff
@@ -203,19 +215,92 @@ setMethod("sqlCreateTable", "DB2/AIX64",
...
-# both quotation marks as well as square bracket are used interchangeably for
-# delimited identifiers.  See:
-# https://learn.microsoft.com/en-us/sql/relational-databases/databases/database-identifiers?view=sql-server-ver16
-# Therefore strip the brackets first, and then call the DBI method that strips
-# the quotation marks.
-# TODO: the generic implementation in DBI should take a quote char as
-# parameter.
+#' SQL Server specific implementation.
+#'
+#' For SQL Server, conn@quote will return the quotation mark, however
```
hadley, line 255:
```suggestion
#' For SQL Server, `conn@quote` will return the quotation mark, however
```

R/db.R
```diff
@@ -203,19 +215,92 @@ setMethod("sqlCreateTable", "DB2/AIX64",
...
+  args <- list(...)
+  if ( "catalog_name" %in% names(args) ) {
+    catalog_name <- args[["catalog_name"]]
+    if ( !is.null(catalog_name) &&
+        catalog_name != "%" &&
+        length(catalog_name ) > 0 &&
+        catalog_name != "tempdb" ) {
+      return(FALSE)
+    }
+  }
```
hadley: FWIW in cases like this, I think it's better to create a helper function and rely on R to do the dot matching, i.e. something like:

```R
mssq_is_temp <- function(..., name, catalog_name = NULL) {
  if ( !is.null(catalog_name) &&
        catalog_name != "%" &&
        length(catalog_name ) > 0 &&
        catalog_name != "tempdb" ) {
      return(FALSE)
    }
  ...
}

File: R/db.R [hunk 3]
```diff
@@ -203,19 +215,92 @@ setMethod("sqlCreateTable", "DB2/AIX64",
...
   })
+
+#' SQL Server specific implementation.
+#'
+#' Local temp tables are stored as
+#' \code{ [tempdb].[dbo].[#name]________(padding using underscores)[numeric identifier] }
+#'
+#' True if:
+#' - If catalog_name is supplied it must equal "temdb" or "%" ( wildcard )
+#' - Name must start with "#" followd by a non-"#" character
```
hadley, line 281: I'd suggest just writing all these method docs in a bulleted list in the `SQLServer` docs. That makes it a bit easier to see how the final rendered doc will look.

And I think it's fine to leave the `@usage` as is, so the read can more easily see what methods are customised.

detule: Thanks Hadley.

I took a look at the markdown in `SQLServer.md` and fwict it already looks like a bulleted list:

```
  \details{
  True if:
  \itemize{
  \item If catalog_name is supplied it must equal "temdb" or "\%" ( wildcard )
  \item Name must start with "#" followd by a non-"#" character
  }
```
Let me know if you had something else in mind.

RE: usage, I spend some time struggling with this:

```
❯ checking for code/documentation mismatches ... WARNING
  Functions or methods with usage in documentation object 'Microsoft SQL Server-class' but not in code:
    ‘\S4method{isTempTable}{Microsoft SQL Server,character}’
```

My first thought was that it has something to do with the class name having white-space characters but now i am not so sure.

The class and method are tagged like so:

```
  \name{Microsoft SQL Server-class}                                                                                                                                                                                                         
  \alias{Microsoft SQL Server-class}
  \alias{isTempTable,Microsoft SQL Server,character-method}
```
while the usage section:
```
\usage{
  \S4method{isTempTable}{`Microsoft SQL Server`,character}(conn, name, catalog_name = NULL, schema_name = NULL, ...)                                                                     
}
```

Let me know if you have any ideas.

Thanks again!

hadley: If you manually change the alias to the following, does the problem go away?

```
  \alias{isTempTable,`Microsoft SQL Server`,character-method}
```

detule: Good suggestion; alas doesn't work.

hadley: Weird! Definitely not worth looking into further.

hadley (approved):

PR #603, opened 2023-09-15
Update to testthat 3e

hadley:


detule:
Looks good.  I'll try and understand the PostgresSQL failure and revert back here.

hadley:
Oh I think that's because I wasn't paying attention — I think all it needs is changing "numeric" to "double" in that test.

detule (approved):

Issue #613, opened 2023-10-23
RStudio IDE Connection Pane issue when connecting to Databricks SQL Warehouse 


edgararuiz:
### Issue Description and Expected Result

When connecting to a Databricks SQL Warehouse, this is how the RStudio IDE's Connection pane should look. The user is able to expand "catalogs", to see "schemas" (they call it Databases), and then expand to see tables. The following screenshot is how it looks when using [`sparklyr` to connect:](https://spark.rstudio.com/deployment/databricks-spark-connect.html) 

![image](https://github.com/r-dbi/odbc/assets/77294576/1ae2d618-e726-4df6-beaf-aba9684fd21f)

The next screenshot, shows how the Connections pane looks when using `odbc` in the following code:

```r
con <- dbConnect(
  odbc::odbc(),
  Driver = "/Library/simba/spark/lib/libsparkodbc_sb64-universal.dylib",
  Host = "[Databricks URL for the Organizations host]",
  Port = 443,
  AuthMech = 3,
  HTTPPath = "/sql/1.0/warehouses/300bd24ba12adf8e",
  Protocol = "https",
  ThriftTransport = 2,
  SSL = 1,
  UID = "token",
  PWD = Sys.getenv("DATABRICKS_TOKEN")
  )
```
![image](https://github.com/r-dbi/odbc/assets/77294576/37a48d05-380b-4fd4-bc93-a510580f6dfd)

When we specify the `catalog` to use, the Connections pane improves. It shows the schema and tables, but, as seen in the following screenshot, it shows the exact same schema and tables beneath all of the catalogs, instead of only the one that we selected.

```r
con <- dbConnect(
  odbc::odbc(),
  Driver = "/Library/simba/spark/lib/libsparkodbc_sb64-universal.dylib",
  Host = "[Databricks URL for the Organizations host]",
  Port = 443,
  AuthMech = 3,
  HTTPPath = "/sql/1.0/warehouses/300bd24ba12adf8e",
  Protocol = "https",
  ThriftTransport = 2,
  SSL = 1,
  UID = "token",
  PWD = Sys.getenv("DATABRICKS_TOKEN"),
  Catalog = "samples"
  )
```

![image](https://github.com/r-dbi/odbc/assets/77294576/84e74b8f-cd62-487f-a20b-9fb32992e8d1)

### Database
[...]

PR #614, opened 2023-10-26
spark/databricks: out-of-catalog schema enumeration

detule:
Fixes https://github.com/r-dbi/odbc/issues/613

Screenshot 2023-10-25 at 11 16 59 PM


edgararuiz:
Hi, can I control the Catalog it will show me the tables for?  I added `Catalog="main"` to the connection arguments, but it didn't show me any tables, and it still showed the schema/tables for `samples`

```r
library(DBI)
con <- dbConnect(
  odbc::odbc(),
  Driver = "/Library/simba/spark/lib/libsparkodbc_sb64-universal.dylib",
  Host = "[posit url]",
  Port = 443,
  AuthMech = 3,
  HTTPPath = "/sql/1.0/warehouses/300bd24ba12adf8e",
  Protocol = "https",
  ThriftTransport = 2,
  SSL = 1,
  UID = "token",
  PWD = Sys.getenv("DATABRICKS_TOKEN"),
  Catalog = "main"
  )
```


Screenshot 2023-10-26 at 1 57 51 PM

detule:
Hey Edgar:

Thanks for checking it out!  Looks like schemas are now enumerated correctly, but you have stumbled across tables within some schemas being listed inconsistently.

Not entirely sure what's going on - do you have the ability to create another catalog and populate it with some sample data? Trying to figure out if it is a permission issue / `main` being somehow special.  Or maybe something specific to `information_schema` ( since i think `default` is actually empty in this case ).

I don't think I have `CREATE CATALOG` permissions / can't try it myself.

edgararuiz:
Hi @detule , maybe this connection needs a custom `odbcConnectionTables` method? I'm able to access the tables by passing the full path of the catalog/schema

```r
dbGetQuery(con, "show schemas in main")
#>         databaseName
#> 1            default
#> 2 information_schema
dbGetQuery(con, "show tables in main.information_schema")
#>              database                       tableName isTemporary
#> 1  information_schema              catalog_privileges       FALSE
#> 2  information_schema                    catalog_tags       FALSE
#> 3  information_schema                        catalogs       FALSE
#> 4  information_schema               check_constraints       FALSE

```

detule:
Hi Edgar - maybe.

Can we verify that 'main' is the only catalog that is misbehaving?

edgararuiz:
`main` seems to be working fine when accessed via passing the full path in the SQL, that's why I think that it needs a custom `odbcConnectionTables` method

detule:
Thanks Edgar.

Before we go ahead and write custom methods - do you have a way of verifying whether enumerating tables with the existing code works fine away from 'main'?

Trying to figure out if this is an issue with the driver that we can raise with the vendor.

edgararuiz:
Yeah, I tried it with `samples`, and it worked fine:

```r
dbGetQuery(con, "show schemas in samples")
#>   databaseName
#> 1      default
#> 2      nyctaxi
#> 3         tpch
dbGetQuery(con, "show tables in samples.nyctaxi")
#>   database tableName isTemporary
#> 1  nyctaxi     trips       FALSE
```

detule:
Thanks - I did some tresting as well away from `main`: current methods to enumerate schema and tables _seem_ to work fine.  Below, see both the `hive_metastore` and `samples` catalogs, and tables within them listed correctly.

Screenshot 2023-10-28 at 9 58 44 PM

@hadley I think let's:

A.  Merge this PR that fixes enumerating schemas - this is a legitimate issue with our code. 
B.  Not being able to enumerate tables using existing methods - if it is isolated to `main` this seems like maybe an issue with the driver, and we should maybe raise with the vendor.  At any rate, if there is urgency to fix here in the near term, we can maybe write a special case for DataBricks + catalog == `main` in a separate PR.

detule:
Not quite a reprex, but this is the output from the integrated query log:

Screenshot 2023-11-01 at 5 48 39 PM

Lower three queries are all calls to [`SQLTables`](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqltables-function?view=sql-server-ver16) with us varying the TableType parameter ( from `TABLE` to `VIEW` to `SYSTEM VIEW` - driver interprets all as `NULL` ).  We retrieved no results from any of those calls.  The top query is a call to `SQLTables` away from the `main` catalog.

hadley (approved):

PR #615, opened 2023-11-08
Add a helper class for using the Databricks ODBC driver

atheriel:
The official Databricks ODBC driver supports a number of authentication methods but these are difficult to use correctly and require users to pass them manually rather than picking up on ambient credentials as other Databricks SDKs do.

This commit adds a high-level helper class that implements (a subset of) the Databricks [client unified authentication model](https://docs.databricks.com/en/dev-tools/auth.html#databricks-client-unified-authentication) in its `dbConnect()` method and should make connection code that runs on Databricks-aware environments work automatically without manual configuration.

Compare the interface from this commit:

    DBI::dbConnect(
      odbc::databricks(),
      http_path = "sql/protocolv1/o/4425955464597947/1026-023828-vn51jugj"
    )

with filling all of the required parameters out manually:

    DBI::dbConnect(
      odbc::odbc(),
      driver = "/opt/simba/spark/lib/64/libsparkodbc_sb64.so",
      host = gsub("https://", "", Sys.getenv("DATABRICKS_HOST")),
      HTTPPath = "sql/protocolv1/o/4425955464597947/1026-023828-vn51jugj"
      ThriftTransport = 2,
      UserAgentEntry = Sys.getenv("SPARK_CONNECT_USER_AGENT"),
      port = 443,
      Protocol = "https",
      SSL = 1,
      AuthMech = 11,
      Auth_Flow = 0,
      Auth_AccessToken = "<your OAuth token>"
    )


R/databricks.R
```diff
@@ -0,0 +1,129 @@
...
+is_hosted_session <- function() {
+  # If RStudio Server or Posit Workbench is running locally (which is possible,
+  # though unusual), it's not acting as a hosted environment.
+  Sys.getenv("RSTUDIO_PROGRAM_MODE") == "server" &&
+    !grepl("localhost", Sys.getenv("RSTUDIO_HTTP_REFERER"), fixed = TRUE)
+}
+
+# Returns a sensible driver name even if odbc.ini and odbcinst.ini do not
+# contain an entry for the Databricks ODBC driver.
+default_databricks_driver <- function() {
```
hadley, line 118: So good

atheriel:
Updated to use a new S4 class instead of a helper function.

R/databricks.R
```diff
@@ -0,0 +1,159 @@
...
+  # implement a subset of the "Databricks client unified authentication" model.
+  token <- Sys.getenv("DATABRICKS_TOKEN")
+  client_id <- Sys.getenv("DATABRICKS_CLIENT_ID")
+  client_secret <- Sys.getenv("DATABRICKS_CLIENT_SECRET")
+  cli_path <- Sys.getenv("DATABRICKS_CLI_PATH", "databricks")
+
+  user_agent <- paste0("r-odbc/", utils::packageVersion("odbc"))
+  if (nchar(Sys.getenv("SPARK_CONNECT_USER_AGENT")) != 0) {
+    # Respect the existing user-agent if present. Normally we'd append, but the
+    # Databricks ODBC driver does not yet support space-separated entries in
```
bschwedler, line 80: Well that is most unfortunate.

R/databricks.R
```diff
@@ -0,0 +1,159 @@
...
+  # official installers. On Windows we use the official driver name instead.
+  default_path <- ""
+  if (Sys.info()["sysname"] == "Linux") {
+    default_path <- "/opt/simba/spark/lib/64/libsparkodbc_sb64.so"
+  } else if (Sys.info()["sysname"] == "Darwin") {
+    default_path <- "/Library/simba/spark/lib/libsparkodbc_sbu.dylib"
+  }
+
+  if (file.exists(default_path)) default_path else "Simba Spark ODBC Driver"
+}
```
bschwedler, line 159: Yay for sensible defaults.

For linux, would we want to default to `/opt/rstudio-drivers/spark/bin/lib/libsparkodbc_sb64.so` and fallback to the `/opt/simba/spark` location?

atheriel: Is that the expected install location for the Pro driver? That could make sense.

bschwedler: Yes, that is the expected location of the driver. I don't quite understand `bin/lib` in the path, but it follows the convention of where the other drivers are installed.

R/databricks.R
```diff
@@ -0,0 +1,159 @@
...
+#' In particular, the custom `dbConnect()` method for the Databricks ODBC driver
+#' implements a subset of the [Databricks client unified authentication](https://docs.databricks.com/en/dev-tools/auth.html#databricks-client-unified-authentication)
+#' model, with support for personal access tokens, OAuth machine-to-machine
+#' credentials, and OAuth user-to-machine credentials supplied via Posit
+#' Workbench or the Databricks CLI on desktop.
+#'
+#' All of these credentials are detected automatically if present using
+#' [standard environment variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
+#'
+#' @param http_path To query a cluster, use the HTTP Path value found under
```
hadley, line 17: 🤢 Should this be called `httpPath`? 🤢

R/databricks.R
```diff
@@ -0,0 +1,159 @@
...
+  "dbConnect", "DatabricksOdbcDriver",
+  function(drv, http_path, workspace = Sys.getenv("DATABRICKS_HOST"),
+           driver = NULL, ...) {
+    args <- databricks_args(
+      http_path = http_path,
+      workspace = workspace,
+      driver = driver
+    )
+    args <- c(args, ...)
+    inject(callNextMethod(!!!args))
```
hadley, line 57: I have zero trust for `callNextMethod()`, so I'd suggest something like:

```suggestion
    inject(dbConnect(odbc(), !!!args))
```

R/databricks.R
```diff
@@ -0,0 +1,159 @@
...
+                            driver = NULL) {
+  if (nchar(workspace) == 0) {
+    stop("No Databricks workspace URL provided")
+  }
+  hostname <- gsub("https://", "", workspace)
+  driver <- driver %||% default_databricks_driver()
+
+  # Check some standard Databricks environment variables. This is used to
+  # implement a subset of the "Databricks client unified authentication" model.
+  token <- Sys.getenv("DATABRICKS_TOKEN")
```
hadley, line 72: Not necessary for this PR, but I wonder if we might want a `databricksSitRep()` function that would print the env vars and paths in a way that makes debugging/support easier.

R/databricks.R
```diff
@@ -0,0 +1,159 @@
...
+    wb_token <- .rs.api.getDatabricksToken(workspace)
+  }
+
+  if (nchar(token) != 0) {
+    # An explicit PAT takes precedence over everything else.
+    args <- c(args, AuthMech = 3, uid = "token", pwd = token)
+  } else if (nchar(client_id) != 0) {
+    # Next up are explicit OAuth2 M2M credentials.
+    args <- c(
+      args, AuthMech = 11, Auth_Flow = 1, Auth_Client_ID = client_id,
```
hadley, line 109: Style: one arg per line

hadley (approved):

R/databricks.R
atheriel: There are some snake-case arguments for functions in this package already (as well as camel-case and dot-separated ones :roll_eyes:), so it wouldn't be breaking precedent. But I'm happy to change it to camel-case if you wish.

R/databricks.R
atheriel: Fixed.

hadley:
I'm happy to merge once the build falures are resolved.

hadley:
Thanks!

PR #621, opened 2023-11-21
Support multiple defaults for databricks path

bschwedler:
This introduces a vector of known paths where we search for the databricks driver.

The first path for the shared library that exists will be used as the default path.


bschwedler:
@atheriel @hadley @blairj09 

This adds the default path for the pro driver to the paths that are searched when the `Driver` parameter is not specified.

atheriel:
Originally I held off implementing this because I wasn't certain whether this will actually be the final Pro Drivers path, and there was no way of testing it.

R/databricks.R
```diff
@@ -153,14 +153,20 @@ is_hosted_session <- function() {
 # Returns a sensible driver name even if odbc.ini and odbcinst.ini do not
 # contain an entry for the Databricks ODBC driver.
 default_databricks_driver <- function() {
-  # For Linux and macOS we can default to the shared library paths used by the
+  # For Linux and macOS we can default to known shared library paths used by the
   # official installers. On Windows we use the official driver name instead.
-  default_path <- ""
+  default_paths <- c("")
```
atheriel, line 158: Nit: style.

```suggestion
  default_paths <- ""
```

R/databricks.R
```diff
@@ -153,14 +153,20 @@ is_hosted_session <- function() {
...
-    default_path <- "/opt/simba/spark/lib/64/libsparkodbc_sb64.so"
+    default_paths <- c(
+      "/opt/rstudio-drivers/spark/bin/lib/libsparkodbc_sb64.so",
+      "/opt/simba/spark/lib/64/libsparkodbc_sb64.so"
+    )
   } else if (Sys.info()["sysname"] == "Darwin") {
-    default_path <- "/Library/simba/spark/lib/libsparkodbc_sbu.dylib"
+    default_paths <- c(
+      "/Library/simba/spark/lib/libsparkodbc_sbu.dylib"
+    )
```
atheriel, line 167: Nit: style.

```suggestion
    default_paths <- "/Library/simba/spark/lib/libsparkodbc_sbu.dylib"
```

hadley: My driver is in `/Library/simba/spark/lib/libsparkodbc_sb64-universal.dylib`. Should we try that too?

bschwedler: Alternatively, we could take a pattern-matching approach for the filename, and search expected directories for a match: `/libsparkodbc_sb.*\.(dylib|so)$/`

hadley: Yeah, that sounds like a good idea.

R/databricks.R
```diff
@@ -47,10 +47,10 @@ setClass("DatabricksOdbcDriver", contains = "OdbcDriver")
 #' @export
 setMethod(
   "dbConnect", "DatabricksOdbcDriver",
-  function(drv, http_path, workspace = Sys.getenv("DATABRICKS_HOST"),
+  function(drv, HTTPPath, workspace = Sys.getenv("DATABRICKS_HOST"),
```
bschwedler, line 50: I renamed the `http_path` parameter to `HTTPPath` throughout so that it matches the parameter name that would be required when connecting with `odbc::odbc()`.

Hopefully this will prevent confusion when switching between these two methods.

hadley (approved):

R/databricks.R
bschwedler: Changing this broke CI, so I need to address that and add the commit back in.

bschwedler: @atheriel Do you have any concerns about changing this to `HTTPPath`?

hadley: Or `httpPath` if we want to standardise on lowerCamelCase?

atheriel (approved):
This looks good to me, I think we can change the case of the argument in a separate PR if needed.

PR #632, opened 2023-11-29
Don't translate SQL for databricks

hadley:


detule (commented):
This looks great.

Was thinking about this some more: If I am reading the driver doc correctly, apparently without the translation parametrized queries no longer work.  Does that mean that our `dbWriteTable` wouldn't work without the translation layer?  Even if it does, I wonder if it makes sense to provide some mechanism for the user to enable the translation layer ( even if it's turned off by default ).

Nice work again.

hadley:
What docs are you referring to?

detule:
I got the driver from here:

https://www.databricks.com/spark/odbc-drivers-download

There is a `Simba Apache Spark ODBC Connector Install and Configuration Guide.pdf`

hadley:
@detule thanks!

atheriel (approved):
:tada:

Issue #641, opened 2023-12-01
Provide some diagnostics for case-equivalent parameter names


hadley:
ODBC parameter names appear to be case insensitive so we should provide some extra tooling to ensure that (e.g.) this is an error:

```R
DBI::dbConnect(odbc::odbc(), foo = 1, Foo = 1)
```

PR #642, opened 2023-12-01
Databricks polishing

hadley:
* Use lowerCamelCase for argument and parameter names
* Don't document/export the class
* Refactoring to pull out separate concerns into own functions
* Warn if no obvious source of credentials found
* Add new driver name fallback and error if no driver found
* Very mild testing


atheriel (approved):
All great polish, just one suggestion.

R/databricks.R
```diff
@@ -33,105 +33,186 @@
...
+  } else if (Sys.info()["sysname"] == "Darwin") {
+    default_paths <- Sys.glob("/Library/simba/spark/lib/libsparkodbc_sb*.dylib")
+  }
+  default_paths <- default_paths[file.exists(default_paths)]
+
+  if (length(default_paths) > 0) {
+    return(default_paths[1])
+  }
+
+  fallback <- "Simba Spark ODBC Driver"
```
hadley, line 133: @bschwedler does this logic look right to you?

R/databricks.R
```diff
@@ -33,105 +33,191 @@
...
+databricks_default_driver <- function() {
+  default_paths <- databricks_default_driver_paths()
+  if (length(default_paths) > 0) {
+    return(default_paths[1])
+  }
+
+  fallback <- "Simba Spark ODBC Driver"
+  if (fallback %in% odbcListDrivers()$name) {
+    return(fallback)
+  }
```
bschwedler, line 127:
```suggestion
  fallbacks <- c("Databricks", "Simba Spark ODBC Driver")
  fallbacks <- intersect(fallbacks, odbcListDrivers()$name)
  if (len(fallbacks) > 1) {
    return(fallbacks[1])
  }
```

If we want to cover the majority of cases, I would add `Databricks` to a list of fallbacks to look for:

* `Databricks`
    * Driver name from the sample `odbcinst.ini` file shipped with the Posit drivers (for Linux/macOS)
    * Driver name as written by Desktop Pro on [Linux/macOS](https://github.com/rstudio/rstudio/blob/d9c0b090d49752fe60e7a2ea4be3123cc3feeb6c/src/cpp/session/modules/SessionConnectionsInstaller.R#L436-L441) and [Windows](https://github.com/rstudio/rstudio/blob/d9c0b090d49752fe60e7a2ea4be3123cc3feeb6c/src/cpp/session/modules/SessionConnectionsInstaller.R#L482-L486)
* `Simba Spark ODBC Driver` is the name of the driver on Windows when installed from the [Databricks download](https://www.databricks.com/spark/odbc-drivers-download)

The driver name is dependent on how the user has configured the system so it may not be the Posit driver.

R/databricks.R
```diff
@@ -14,7 +14,7 @@
 #' [standard environment variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
 #'
 #' @inheritParams DBI::dbConnect
-#' @param HTTPPath To query a cluster, use the HTTP Path value found under
+#' @param httpPath To query a cluster, use the HTTP Path value found under
```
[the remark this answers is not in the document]
hadley: Yes, odbc itself is case-insensitive, and I think we should try to move consistently to once case style in all our documentation and examples. I've filed #641 to ensure that we give better diagnostics when confronted with multiple arguments that differ only in case.

PR #655, opened 2023-12-07
Document dbListTables and dbListFields together

### LINKED ISSUE #645: `dbListTables<OdbcConnection>` help-file shouldn't pull in description of generics
In [`dbListTables<OdbcConnection>`'s help-page](https://r-dbi.github.io/odbc/reference/dbListTables-OdbcConnection-method.html), the "methods in other packages" subsection doesn't make sense.

hadley:
And inherit less from DBI in order to avoid confusing text. Mostly taken from #618. Fixes #645


R/Connection.R
```diff
@@ -494,45 +494,57 @@ setMethod(
...
-#' @inherit DBI::dbListTables
-#' @param catalog_name The name of the catalog to return, the default returns all catalogs.
-#' @param schema_name The name of the schema to return, the default returns all schemas.
-#' @param table_name The name of the table to return, the default returns all tables.
+#' List remote tables and fields for an ODBC connection
+#'
+#' `dbListTables()` provides names of remote tables accessible through this
+#' connection; `dbListFields()` provides names of columns within a table.
+#'
+#' @inheritParams DBI::dbListTables
```
simonpcouch, line 502:
```suggestion
#' @inheritParams DBI::dbListFields
```

Seeing:

```
❯ checking Rd \usage sections ... WARNING
  Undocumented arguments in documentation object 'dbListTables,OdbcConnection-method'
    'name'
```

in GHA. `?DBI::dbListFields` documents both `conn` and `name` and looks like the place where that `@param`'s docs were inherited from previously.

R/Connection.R
```diff
@@ -494,45 +494,57 @@ setMethod(
...
+#' @param catalog_name,schema_name,table_name Catalog, schema, and table names.
+#'
+#'   Use `%` as a wildcard to match any name; use `_` to any any character.
 #' @param table_type The type of the table to return, the default returns all table types.
-#' @aliases dbListTables
```
simonpcouch, line 502: Noting that this change, and the analogous one for the `dbListFields` method, means that with only odbc loaded (i.e. not having called `library(DBI)`) `?dbListTables` will now head to the generic's docs in DBI rather than the "pick the place where you want to go to" docs:

Screenshot 2023-12-07 at 2 28 13 PM

This sounds good to me, since DBI's docs link to the odbc method's docs early on, but is this a change we want to make generally?

simonpcouch (commented):
🐧

R/Connection.R
hadley: Good catch, thanks!

hadley: Oh actually, I had this further down; I had just forgotten to re-document 😬

PR #656, opened 2023-12-07
Automatically quote when generating the connection string

### LINKED ISSUE #616: Password authentication error with Amazon Redshift if certain characters are present in the password
```
library(odbc)
  con <- dbConnect(odbc::odbc(), 
  driver = "Redshift", 
  server = abc.redhsift.amazonaws.com, 
  database = db_name, 
  uid = abc,
  pwd = password,
  port = 1234)
```  
returns
**Error: nanodbc/nanodbc.cpp:1135: 0000: [RStudio][AmazonRedshift] (10) Error occured while trying to connect: [SQLState 28000] connection to server "abc.redhsift.amazonaws.com" , port 1234 failed: FATAL: password authentication failed for user "abc".**

I do not see this error when I try connecting to Redshift using python. My observation is that the whenever the Redshift password contains the characters **[]{}(),;?*=!@** it fails with the above error. The Redshift documentation [here](https://docs.aws.amazon.com/redshift/latest/mgmt/configure-odbc-connection.html#:~:text=following%20special%20characters%3A-,%5B%5D%7B%7D()%2C%3B%3F*%3D!%40,-If%20you%20use) suggests to wrap the password inside curly braces **{}** whenever these characters are present but wrapping them inside curly braces also does not work with the odbc.

### Database
Amazon Redshift

hadley:
Fixes #616

@atheriel no need for a detailed review; but can you see any obvious unintended consequences from this change?

### DIFF
--- a/R/Connection.R
+++ b/R/Connection.R
@@ -55,7 +55,7 @@ OdbcConnection <- function(
   stopifnot(all(has_names(attributes)))
   stopifnot(all(names(attributes) %in% SUPPORTED_CONNECTION_ATTRIBUTES))
 
-  connection_string <- paste0(.connection_string, paste(collapse = ";", sep = "=", names(args), args))
+  connection_string <- paste0(.connection_string, connection_string(args))
 
   bigint <- bigint_mappings()[match.arg(bigint, names(bigint_mappings()))]
 
@@ -82,6 +82,15 @@ OdbcConnection <- function(
       contains = "OdbcConnection", where = class_cache)
   }
   res <- new(info$dbms.name, ptr = ptr, quote = quote, info = info, encoding = encoding)
+}
+
+build_connection_string <- function(args) {
+  needs_escape <- grepl("[{}(),;?*=!@]", args) &
+    !grepl("^\\{.*\\}$", args) &
+    !vapply(args, inherits, "AsIs", FUN.VALUE = logical(1))
+
+  args[needs_escape] <- paste0("{", args[needs_escape], "}")
+  paste(names(args), args, sep = "=", collapse = ";")
 }
 
 #' @rdname OdbcConnection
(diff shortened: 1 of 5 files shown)


R/Connection.R
```diff
@@ -55,7 +55,7 @@ OdbcConnection <- function(
   stopifnot(all(has_names(attributes)))
   stopifnot(all(names(attributes) %in% SUPPORTED_CONNECTION_ATTRIBUTES))
 
-  connection_string <- paste0(.connection_string, paste(collapse = ";", sep = "=", names(args), args))
+  connection_string <- paste0(.connection_string, connection_string(args))
```
simonpcouch, line 58:
```suggestion
  connection_string <- paste0(.connection_string, build_connection_string(args))
```

Will spend some time to make sure I've vetted the new helper well, but first thing:

tests/testthat/test-Connection.R
```diff
@@ -0,0 +1,13 @@
...
+  expect_equal(build_connection_string(list()), "")
+  expect_equal(build_connection_string(list(foo = "1")), "foo=1")
+  expect_equal(build_connection_string(list(foo = "1", bar = "2")), "foo=1;bar=2")
+})
+
+test_that("build_connection_string automatically escapes if needed", {
+  expect_equal(build_connection_string(list(foo = "*")), "foo={*}")
+  # Already wrapped
+  expect_equal(build_connection_string(list(foo = "{*}")), "foo={*}")
+  #
```
simonpcouch, line 11:
```suggestion
  # Respects AsIs
```

simonpcouch (approved):
Looks good to me once we're calling the correct helper, though I'm a little sketched out that that didn't show up in any CI🤔😵‍💫

atheriel (approved):
I don't think so. LGTM.

PR #658, opened 2023-12-07
Automatically set immediate to TRUE if not using parameters

### LINKED ISSUE #633: Automatically set `immediate` in `dbGetQuery()`
To something like `immediate = !is.null(params)`. This should give a small performance boost.

hadley:
Fixes #633


hadley:
Need to look into the two failures with SQL Server both with error "Error: nanodbc/nanodbc.cpp:1708: 00000: [Microsoft][ODBC Driver 17 for SQL Server]Connection is busy with results for another command". They're coming from test "send_query_only_one_result_set" and "send_query_immediate". Need to read the docs for https://dbitest.r-dbi.org/reference/spec_result_send_query#specification-for-the-immediate-argument-1.

hadley:
Minimal reprex:

```R
con <- DBI::dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA", pwd = "BoopBop123")
res1 <- dbSendQuery(con, "SELECT 1", immediate = TRUE)
res2 <- dbSendQuery(con, "SELECT 2", immediate = TRUE)
dbClearResult(res1)
dbClearResult(res2)
```

src/odbc_result.cpp
```diff
@@ -16,6 +16,8 @@ odbc_result::odbc_result(
       bound_(false),
       output_encoder_(Iconv(c_->encoding(), "UTF-8")) {
 
+  c_->cancel_current_result(false);
```
hadley, line 19: Previously we were only cancelling the results _after_ performing the query which doesn't work when `immediate = TRUE`. I think cancelling before makes more sense, and it's a bit nicer to explicit about it, rather than assuming you remember that `set_current_result` will cancel if needed.

src/odbc_connection.cpp
```diff
@@ -4,18 +4,24 @@
 
 namespace odbc {
 
-void odbc_connection::set_current_result(odbc_result* r) {
-  if (r == current_result_) {
+void odbc_connection::cancel_current_result(bool quiet) {
```
hadley, line 7: My C++ is extremely rusty, but I think this is simple enough that I didn't mess it up. (Famous last words!)

detule (approved):
Nice work, looks good!

PR #662, opened 2023-12-08
Add a helper function for connecting to Snowflake.

atheriel:
This commit introduces a new `odbc::snowflake()` function intended to help setting up connections to Snowflake and handling auth correctly on platforms that provide Snowflake-native OAuth credentials. It generally follows the pattern established with `odbc::databricks()`.

We've been testing this internally on Snowflake's container platform, and it works as expected.


atheriel:
Tests are currently failing because we now have both a `Snowflake.Rd` and a `snowflake.Rd` file. I'm not sure what you'd suggest in this case, but happy to fix what's needed.

simonpcouch:
Ah, interesting. From my perspective, I'd say we transition existing `@rdname Snowflake`s to `@rdname driver-Snowflake` and allow `?snowflake` to head to docs on the wrapper implemented in this PR.

atheriel:
@simonpcouch Done.

hadley:
Since we just added snowflake to our CI, should we add a couple of very basic tests?

atheriel:
I've added a number of unit tests (and fixed some issues in the process!), but due to the design of the current CI secrets -- which supply entire connection strings -- I can't see an easy way to pass real credentials through the new `odbc::snowflake()` function.

hadley:
@simonpcouch could you please take a look at how we're managing secrets to see if we could facilitate their use here?

R/driver-snowflake.R
```diff
@@ -35,7 +35,7 @@ getCatalogSchema <- function(conn, catalog_name = NULL, schema_name = NULL) {
 #' to aid with performance, as the SQLColumns method is more performant
 #' when restricted to a particular DB/schema.
 #' @inheritParams DBI::dbListFields
-#' @rdname Snowflake
+#' @rdname driver-Snowflake
```
hadley, line 38: I'd be tempted to document them alltogether, adding a new section that details the snowflake method variations. But might be easiest for me to do this after the PR is merged.

R/driver-snowflake.R
```diff
@@ -87,3 +87,212 @@ setMethod("odbcDataType", "Snowflake",
     )
   }
 )
+
+#' Helper for Connecting to Snowflake via ODBC
```
hadley, line 91: Should be sentence case

R/driver-snowflake.R
```diff
@@ -87,3 +87,212 @@ setMethod("odbcDataType", "Snowflake",
...
+  function(drv,
+           account = Sys.getenv("SNOWFLAKE_ACCOUNT"),
+           driver = NULL,
+           warehouse = NULL,
+           database = NULL,
+           schema = NULL,
+           uid = NULL,
+           pwd = NULL,
+           ...) {
+    args <- snowflake_args(
```
hadley, line 159: Need to add basic input validation here

R/driver-snowflake.R
```diff
@@ -87,3 +87,212 @@ setMethod("odbcDataType", "Snowflake",
...
+  all <- utils::modifyList(c(args, auth), list(...))
+
+  # Respect the Snowflake Partner environment variable, if present.
+  if (is.null(all$application) && nchar(Sys.getenv("SF_PARTNER")) != 0) {
+    all$application <- Sys.getenv("SF_PARTNER")
+  }
+
+  arg_names <- tolower(names(all))
+  if (!"authenticator" %in% arg_names && !all(c("uid", "pwd") %in% arg_names)) {
+    warn(
```
hadley, line 193: Should this be a warning or an error?

simonpcouch:
Just pushed a quick demo of using the secret in tests--fine by me to have both the PWD embedded in a connection string and the PWD itself in the `GITHUB_ENV`. The remaining connection information is in [`.github/odbc/odbc.ini`](https://github.com/r-dbi/odbc/blob/62c7432b143cffc154492dbf45cba4b4551aa715/.github/odbc/odbc.ini#L46).

simonpcouch:
In the above, "ambient" credentials are those set by default in the Posit Workbench build hosted at dev palm ptd:

```r
grep("SNOWFLAKE", names(Sys.getenv()), value = TRUE)
#> [1] "SNOWFLAKE_ACCOUNT" "SNOWFLAKE_HOME"    "SNOWFLAKE_ROLE"  
```

If this is just a config issue, we should detect the conditions when this error will be raised and raise a more informative one before dispatching to `dbConnect.OdbcConnection()`.

R/driver-snowflake.R
atheriel: I'm emulating the `odbc::databricks()` call here, which just issues a warning. I'm open to changing both, though.

R/driver-snowflake.R
atheriel: I don't think we do that in `odbc::databricks()` either but could add it; do you have suggestions?

R/driver-snowflake.R
atheriel: Fixed.

R/driver-snowflake.R
simonpcouch: See [these lines](https://github.com/r-dbi/odbc/blob/2310dae3b12e53564e0baa495463a6dd41fddcd9/R/driver-databricks.R#L68-L76)! Those are recent changes.

atheriel:
This is ready for re-review:

* Added additional tests, including a smoke test to make sure that `odbc::snowflake()` can be used to connect to the real Snowflake account in our CI system.
* I've fixed the Quarto rendering issue by removing the call to in-process RStudio APIs; instead we parse the Workbench-generated file directly, which should be a lot more portable and future-proof.
* CI failures seem to be related to code I haven't changed, but I may be mistaken.

R/driver-snowflake.R
```diff
@@ -87,3 +87,238 @@ setMethod("odbcDataType", "Snowflake",
...
+        "Both `uid` and `pwd` must be specified to authenticate.",
+        i = "Or leave both unset to use ambient Snowflake credentials."
+      ),
+      call = quote(DBI::dbConnect())
+    )
+  }
+
+  # Manual override of the authentication method.
+  if (!is.null(authenticator)) {
+    list()
```
simonpcouch, line 276:
```suggestion
    return(list())
```

As-is, this line doesn't early return. I believe the suggested is the intent?

R/driver-snowflake.R
```diff
@@ -87,3 +87,238 @@ setMethod("odbcDataType", "Snowflake",
...
+# Reads Posit Workbench-managed Snowflake credentials from a
+# $SNOWFLAKE_HOME/connections.toml file, as used by the Snowflake Connector for
+# Python implementation. The file will look as follows:
+#
+# [workbench]
+# account = "account-id"
+# token = "token"
+# authenticator = "oauth"
+workbench_snowflake_token <- function(account,
+                                      sf_home = Sys.getenv("SNOWFLAKE_HOME")) {
```
simonpcouch, line 309:
```suggestion
workbench_snowflake_token <- function(account, sf_home) {
```

This default made me think that we might need to check its value, but this is actually always set by the calling function

R/driver-snowflake.R
atheriel: Done!

R/driver-snowflake.R
atheriel: Done!

tests/testthat/test-driver-snowflake.R
```diff
@@ -1,3 +1,154 @@
...
+test_that("Workbench-managed credentials are detected correctly", {
+  # Emulate the connections.toml file written by Workbench.
+  sf_home <- tempfile("posit-workbench")
+  dir.create(sf_home)
+  writeLines(
+    '[workbench]
+account = "testorg-test_account"
+token = "token"
+authenticator = "oauth"
+',
```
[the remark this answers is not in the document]
atheriel: Done!

atheriel:
The warning about missing docs is unrelated to this PR but I fixed it anyway to try and make CI happy :)

garrettgman:
cc @skaltman

R/driver-snowflake.R
hadley: I think we should make it an error for both. If you don't want to do in this PR, feel free to file an issue.

hadley (approved):

simonpcouch (approved):
The GHA failure:

```
── Failed tests ────────────────────────────────────────────────────────────────
Error ('test-driver-snowflake.R:2:3'): can connect to snowflake
Error in `test_con("SNOWFLAKE")`: ! ODBC failed with error 00000 from [Snowflake][Snowflake].
*  Password not found.
i From 'nanodbc/nanodbc.cpp:1147'.
```

...is due to the relevant repository secret not being available and is fine to ignore. Game to merge with your thumbs-up, @atheriel—just let us know whether you want to address the warning -> error conversion review comment in this PR or file a separate issue to make that happen.

Issue #661, opened 2023-12-08
Look into ambiguous dispatch


hadley:
```
Note: method with signature 'Microsoft SQL Server#character' chosen for function 'dbExistsTable',
 target signature 'Microsoft SQL Server#SQL'.
 "OdbcConnection#SQL" would also be valid
Note: method with signature 'Microsoft SQL Server#character' chosen for function 'isTempTable',
 target signature 'Microsoft SQL Server#SQL'.
 "OdbcConnection#SQL" would also be valid
```

From https://github.com/r-dbi/odbc/actions/runs/7133592650/job/19426605756?pr=658#step:12:16

hadley:
Fix this by moving to single dispatch, something like this:

```R
setMethod(
  "dbExistsTable", c("OdbcConnection", "ANY"),
  function(conn, name, ...) {
    
    if (is(name, "SQL")) {
      # converts to an Id object
      name <- dbUnquoteIdentifier(conn, name)[[1]]
    }
    
    if (is(name, "Id")) {
      df <- odbcConnectionTables(
        conn,
        name = id_field(name, "table"),
        catalog_name = id_field(name, "catalog"),
        schema_name = id_field(name, "schema")
      )
    } if (is.character(name) && length(name) == 1) {
      df <- odbcConnectionTables(conn, name = name, ..., exact = TRUE)
    } else {
      abort("`name` must be a string, Id(), SQL() object.")
    }
    
    NROW(df) > 0
  }
)
```

Also need to think through how this impacts other methods.

PR #664, opened 2023-12-11
Missing methods

hadley:
Part of #661


hadley:
Better to fix upstream

hadley:
Worth doing a quick fix before we fully patch.

R/db.R
```diff
@@ -365,6 +376,20 @@ setMethod(
...
+#' @rdname SQLServer
+#' @usage NULL
+setMethod(
+  "dbExistsTable", c("Microsoft SQL Server", "SQL"),
+  function(conn, name, ...) {
+    dbExistsTable(
+      conn,
+      name = id_field(name, "table"),
+      catalog_name = id_field(name, "catalog"),
+      schema_name = id_field(name, "schema")
```
hadley, line 388: This is ID not SQL

detule (approved):
LGTM for a near term solution.  I like where your ideas for a longer term solution.

PR #669, opened 2023-12-12
Look in two places for class def

hadley:
Now need to look in both the package (for classes we know about) and in the class cache (for dynamically created classes).


detule:
Ugh, not very pretty.

Now that the genie is out of the bottle, do we need `class_cache`, or can we do away with it altogether.

detule (approved):

hadley:
I think we still need it for the case where we've never seen the class before. We could just give those connections the `OdbcConnection` class, but I think there might be some value in having a meaningful class name.

PR #671, opened 2023-12-13
Add missing ... to odbcConnectionColumns()

hadley:
Fixes `Error: Expecting a single string value: [type=logical; extent=1]` when using the RStudio connections pane. https://github.com/rstudio/rstudio-pro/issues/5581

### DIFF
--- a/R/Connection.R
+++ b/R/Connection.R
@@ -189,7 +189,7 @@ setGeneric(
 setMethod(
   "odbcConnectionColumns",
   c("OdbcConnection", "Id"),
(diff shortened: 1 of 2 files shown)


hadley:
I'm not sure why S4 tooling doesn't automatically flag this as a problem 😞

bschwedler (approved):
LGTM

hadley:
Similarly, merging this so it can get into our internal QA. @detule please let me know if you have any concerns (but I think this is a straightforward change to fix a mistake we made a while back)

detule:
LGTM - sorry for the delay.

Issue #624, opened 2023-11-25
MacOS/Sonoma: Crash when building `odbc` from source


detule:
[Other users](https://github.com/r-dbi/odbc/issues/624#issue-2010813484) have [reported this as well](https://github.com/r-dbi/odbc/issues/611), but I am running into this for the first time.

```
> install.packages("odbc", type = "source")
trying URL 'https://mirror.las.iastate.edu/CRAN/src/contrib/odbc_1.3.5.tar.gz'
Content type 'application/x-gzip' length 348241 bytes (340 KB)
==================================================
downloaded 340 KB

* installing *source* package ‘odbc’ ...
** package ‘odbc’ successfully unpacked and MD5 sums checked
** using staged installation
Found odbc_config cflags and libs!
PKG_CFLAGS=-DHAVE_UNISTD_H -DHAVE_PWD_H -DHAVE_SYS_TYPES_H -DHAVE_LONG_LONG -DSIZEOF_LONG_INT=8 -I/opt/R/x86_64/include
PKG_LIBS=-L/opt/R/x86_64/lib -lodbc
** libs
using C++ compiler: ‘Apple clang version 15.0.0 (clang-1500.0.40.1)’
using SDK: ‘MacOSX14.0.sdk’
(cd cctz && \
make libcctz.a CC="clang -arch x86_64" CXX="clang++ -arch x86_64 -std=gnu++17" CXXFLAGS="-falign-functions=64 -Wall -g -O2 " AR="ar" ARFLAGS=rv CXXPICFLAGS="-fPIC")
clang++ -arch x86_64 -std=gnu++17 -falign-functions=64 -Wall -g -O2  -Wall -Iinclude -std=c++11 -pthread  -O -fPIC -MMD  -c -o time_zone_format.o src/time_zone_format.cc
clang++ -arch x86_64 -std=gnu++17 -falign-functions=64 -Wall -g -O2  -Wall -Iinclude -std=c++11 -pthread  -O -fPIC -MMD  -c -o time_zone_if.o src/time_zone_if.cc
clang++ -arch x86_64 -std=gnu++17 -falign-functions=64 -Wall -g -O2  -Wall -Iinclude -std=c++11 -pthread  -O -fPIC -MMD  -c -o time_zone_impl.o src/time_zone_impl.cc
[...]

PR #677, opened 2023-12-13
Use SQLUINTEGER in get_info()

gaborcsardi:
Instead of `unsigned short`, which could be allocated or aligned differently by different compilers.

There are other places in nanodbc that use `unsigned short` in the protocol, so potentially those are broken as well.

I am not very familiar with the ODBC toolchain, so it is possible that this is not the best fix, but nevertheless this is the step that causes the memory error and the crash, so this needs to change.

For #624.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -75,10 +75,10 @@ std::string get_info_or_empty(connection_ptr const& p, short type) {
 
 // [[Rcpp::export]]
 Rcpp::List connection_info(connection_ptr const& p) {
-  SQLINTEGER getdata_ext;
+  SQLUINTEGER getdata_ext;
   try {
     getdata_ext =
-        (*p)->connection()->get_info<SQLINTEGER>(SQL_GETDATA_EXTENSIONS);
+        (*p)->connection()->get_info<SQLUINTEGER>(SQL_GETDATA_EXTENSIONS);
   } catch (const nanodbc::database_error& c) {
     getdata_ext = 0;
   }
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -75,10 +75,10 @@ std::string get_info_or_empty(connection_ptr const& p, short type) {
 
 // [[Rcpp::export]]
 Rcpp::List connection_info(connection_ptr const& p) {
-  unsigned short getdata_ext;
+  SQLINTEGER getdata_ext;
   try {
     getdata_ext =
-        (*p)->connection()->get_info<unsigned short>(SQL_GETDATA_EXTENSIONS);
+        (*p)->connection()->get_info<SQLINTEGER>(SQL_GETDATA_EXTENSIONS);
   } catch (const nanodbc::database_error& c) {
     getdata_ext = 0;
   }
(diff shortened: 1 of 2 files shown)


detule (approved):
Looks great / nice find.

hadley:
@detule can you upstream this problem to nanodbc?

detule:
Hey @hadley @gaborcsardi 

I don't really think this is an issue with `nanodbc`.  More of a user error on our end I think.

Also the `get_info` template is already specialized for 32- and 64-bit storage types - we probably don't need the  `nanodbc.cpp` change; I suspect the compiler will manage to connect with the appropriate one.

gaborcsardi:
@detule Yes, actually, according to the docs at https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetinfo-function?view=sql-server-ver16 it should be `SQLUINTEGER`, so I changed it to that. Then indeed to need for another template instance. For `SQLINTEGER` you would need one, because that is a signed type.

I saw `SQLINTEGER` in some ODBC code or docs which I can't find now, but that was wrong, it should be `SQLUINTEGER`. 

TBH, if the ODBC protocol requires `SQL*` types, that's what nanodbc should be using instead of `unsigned short` etc. But yeah, now no nanodbc changes are needed now.

gaborcsardi:
Actually, we still need a new template instance for `SQLUINTEGER`, because it is a type that is not covered on Windows (https://github.com/r-dbi/odbc/actions/runs/7206871978/job/19632556465):
```
g++ -std=gnu++17 -shared -s -static-libgcc -o odbc.dll tmp.def odbc_result.o connection.o nanodbc.o result.o odbc_connection.o RcppExports.o Iconv.o utils.o -lodbc32 -Lcctz -lcctz -LC:/rtools43/x86_64-w64-mingw32.static.posix/lib/x64 -LC:/rtools43/x86_64-w64-mingw32.static.posix/lib -LC:/R/bin/x64 -lR
C:\rtools43\x86_64-w64-mingw32.static.posix\bin/ld.exe: connection.o:connection.cpp:(.text+0x25ef): undefined reference to `unsigned long nanodbc::connection::get_info<unsigned long>(short) const'
collect2.exe: error: ld returned 1 exit status
```

It is better to use the official types, anyway.

gaborcsardi:
Well, adding an instance for `SQLUINTEGER` breaks on Linux now. :( The compiler complains about instantiating with the same type twice. Let's try to only have instances for the proper SQL types now.

gaborcsardi:
> Let's try to only have instances for the proper SQL types now.

That seems to work on all platforms. But then maybe this is something to report upstream.

hadley:
I'm going to merge this so we can double check it passes our QA. I'll leave it up to you both if there's any additional work to do in odbc or upstream.

mloskot:
Folks, I have commented on PR received from @gaborcsardi in https://github.com/nanodbc/nanodbc/pull/398#issuecomment-1859158360 Please, share your thoughts.

p.s. Big thanks to @gaborcsardi and @detule and other members of the r-odbc community for contributing to nanodbc!

PR #680, opened 2023-12-14
add helper to locate configuration files

### LINKED ISSUE #565: Improve setup documentation
By creating some vignettes instead of trying to put everything in the readme

simonpcouch:
Closes #565.☃️


simonpcouch:
On my machine:

``` r
library(odbc)

odbcListConfig()
#>                  name                       location
#> 1             DRIVERS /opt/homebrew/etc/odbcinst.ini
#> 2 SYSTEM DATA SOURCES     /opt/homebrew/etc/odbc.ini
#> 3   USER DATA SOURCES    /Users/simoncouch/.odbc.ini
```

<sup>Created on 2023-12-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/Connection.R
```diff
@@ -712,6 +712,56 @@ setMethod(
     invisible(TRUE)
   })
 
+#' List Locations of ODBC Configuration Files
```
hadley, line 715: Should be sentence case

R/Connection.R
```diff
@@ -712,6 +712,56 @@ setMethod(
...
+#' On MacOS and Linux, odbc uses the unixODBC driver manager to manage
+#' information about driver and data sources. This helper returns the filepaths
+#' where the driver manager will look for that information.
+#'
+#' @details
+#' This function is a wrapper around the command line call `odbcinst -j`.
+#'
+#' Windows now uses the ODBC Data Source Administrator application to manage
+#' drivers and thus does not use `.ini` files; this function will return a
+#' 0-row data frame on Windows.
```
hadley, line 727: I think you could shrink this up a bit and put it all in the description.

R/Connection.R
```diff
@@ -712,6 +712,56 @@ setMethod(
...
+#' information on the contents of `odbcinst.ini` and `odbc.ini` files,
+#' respectively.
+#'
+#' Learn more about unixODBC and the `odbcinst` utility
+#' [here](https://www.unixodbc.org/odbcinst.html).
+#'
+#' @examplesIf FALSE
+#' configs <- odbcListConfig()
+#'
+#' file.edit(configs$location[1])
```
hadley, line 740: My main question is if this should just be a named character vector with names "drivers", "system_dsn", "user_dsn", or similar.

hadley (approved):
Looks good!

R/Connection.R
simonpcouch: You got it—[680b502](https://github.com/r-dbi/odbc/pull/680/commits/680b50228c201b284d3e93be45b0b2cdcd09cfbb).

R/Connection.R
simonpcouch: You got it—[680b502](https://github.com/r-dbi/odbc/pull/680/commits/680b50228c201b284d3e93be45b0b2cdcd09cfbb).

R/Connection.R
simonpcouch: [680b502](https://github.com/r-dbi/odbc/pull/680/commits/680b50228c201b284d3e93be45b0b2cdcd09cfbb)!

R/Connection.R
simonpcouch: Heard. Had thought about that but I wasn't sure whether folks might end up with multiple user `odbc.ini` files with different levels of permissions/scope. Trying to find some documentation that explicitly states that there is only ever one user-level `odbc.ini` file but I'll run with it if we're confident that that's the case.

simonpcouch: Okay, from the source, looks like there can only ever be one active user-level `odbc.ini`. Glad to make this change if you feel we ought to!

hadley: Yeah, lets do it.

simonpcouch: [6689544](https://github.com/r-dbi/odbc/pull/680/commits/6689544ebfb54fe6d8ee6034030e377ba583b85c)!

R/Connection.R
```diff
@@ -712,6 +712,53 @@ setMethod(
...
+#' configs <- odbcListConfig()
+#'
+#' file.edit(configs[1])
+#' @export
+odbcListConfig <- function() {
+  if (identical(.Platform$OS.type, "windows")) {
+    return(character(0))
+  }
+
+  if (identical(unname(Sys.which("odbcinst")), "")) {
```
hadley, line 745: I think it'd be worth extracting this out to a new `has_odbc()` function. That'll make the skips easier and you can also add a snapshot test of this error by mocking `has_odbc()` to return `FALSE`.

R/Connection.R
```diff
@@ -712,6 +712,53 @@ setMethod(
...
+    stop(
+      c("The unixODBC driver manager is not available. ",
+        "Please install and try again.")
+    )
+  }
+
+  res <- system("odbcinst -j", intern = TRUE)
+  res <- res[grepl("\\.ini", res)]
+  res <- strsplit(res, "\\:")
+  res <- vapply(res, `[[`, character(1), 2)
```
hadley, line 755: I think it's worth considering a quick check of the output here, maybe something like this:

```R
if (!identical(length(res), c(2, 2, 2)) {
  abort("Failed to parse output from odbcinst", internal = TRUE)
}
```

That'll ensure the user gets a reasonable error if `odbcinst` gives unexpected results. (And the `internal = TRUE` ensures they get a prompt to file an issue).

Then you can add a snapshot test for that error by mocking `system()`.

hadley (approved):
Looking good! Just a couple of last thoughts about testing this code; let me know if you haven't used `local_mocked_bindings()` before and I can give you some pointers.

R/Connection.R
simonpcouch: [275b69c](https://github.com/r-dbi/odbc/pull/680/commits/275b69c3740b695d8b96fb6e2d265d81553c1fa6)!

R/Connection.R
simonpcouch: [275b69c](https://github.com/r-dbi/odbc/pull/680/commits/275b69c3740b695d8b96fb6e2d265d81553c1fa6).

simonpcouch:
Testing updates made in [275b69c](https://github.com/r-dbi/odbc/pull/680/commits/275b69c3740b695d8b96fb6e2d265d81553c1fa6). Muuuch better!

I believe [39fb93b](https://github.com/r-dbi/odbc/pull/680/commits/39fb93b407b663804016915d943078c11f17863b) is unrelated to this PR, but saw this output on `devtools::test()`.

R/Connection.R
simonpcouch: My only question here is whether we should run a similar check instead on the output of `system("odbcinst -j", intern = TRUE)`? Based on it's source code I'm skeptical that odbcinst will ever return something that's _almost_ parsable for us, but some issue in the communication with `system()` seems plausible.

hadley: I don't think it's worth spending too much more time contemplating this as it's an unlikely failure mode; we just want to make sure we get something diagnostic in the unlikely event that the output does change.

hadley:
BTW I'm pretty sure `lengths()` is now old enough that we can use it

Also please hold of on merging this PR for a bit.

hadley:
Good to go with the merge. My bad for the mix-up with the merge conflict 'cause of my file reorg PR

simonpcouch:
All good! Glad we made that renaming happen.

tests/testthat/test-odbc-config.R
```diff
@@ -0,0 +1,64 @@
...
+  skip_on_os(c("windows", "solaris"))
+
+  local_mocked_bindings(
+    has_odbc = function() {FALSE}
+  )
+
+  expect_snapshot(
+    error = TRUE,
+    odbcListConfig()
+  )
```
hadley, line 41:
```suggestion
  local_mocked_bindings(has_odbc = function() FALSE)
  expect_snapshot(odbcListConfig(), error = TRUE)
```

I don't have particularly strong feelings, but I think compressing this into a few lines more accurately conveys that these are pretty simple tests.

tests/testthat/test-odbc-config.R
```diff
@@ -0,0 +1,64 @@
...
+test_that("odbcListConfig returns an empty vector on Windows", {
+  skip_on_os(c("mac", "linux", "solaris"))
+
+  res <- odbcListConfig()
+
+  expect_equal(res, character(0))
+})
+
+test_that("odbcListConfig errors informatively without unixODBC", {
+  skip_on_os(c("windows", "solaris"))
```
hadley, line 22: FWIW I'd be tempted to make a `is_window()` function so that you could mock it and then drop the skips.

hadley (approved):
Two tiny last testing things; they're not really important in and of themselves but I think they're good habits to build going forward.

tests/testthat/test-odbc-config.R
simonpcouch: Sure! [12c965f](https://github.com/r-dbi/odbc/pull/680/commits/12c965f6e6d214de690a88f26d7c1d4d39e4041f).

tests/testthat/test-odbc-config.R
simonpcouch: Sure thing. In these tests I always skip on Solaris. Should `is_windows()` on Solaris return true? Or figure out how to support Solaris and test on those systems as well?

hadley: Once you've mocked the right functions, it shouldn't matter which OS it's actually running on.

And I'm reasonably certain we don't generally need to worry about solaris any more?

simonpcouch: [99938dd](https://github.com/r-dbi/odbc/pull/680/commits/99938dd10532cd679075dece04ff61bdfb29ecfb) and [e40215e](https://github.com/r-dbi/odbc/pull/680/commits/e40215ef3a190088577d121b8a584891f4bcd938) should get us most of the way there, though I'm not sure how we'd mock through that first test where we do need odbcinst output.

Issue #480, opened 2021-12-15
Avoid checking existence of table in call to dbAppendTable()


krlmlr:
### Issue Description and Expected Result

`dbExistsTable()` is brittle, success or failure to write data to a table should be reported by the database. Related: #465.

As part of this change, `dbWriteTable()` should be calling `dbAppendTable()`, not the other way round.

### Database

All.

### Reproducible Example

N/A. Relevant code: https://github.com/r-dbi/odbc/blob/main/R/Table.R#L134 .

arthur-holtz:
Greetings -- my company uses DBI/ODBC for an internal tool and I believe the function `dbExistsTable` is incorrectly throwing an error on our end when we call `dbAppendTable`. When I strip out all our custom code so it's only using DBI's functions, I can fetch data from a table but not write to it.

Is the plan to remove the call to `dbExistsTable` in the definition of `dbAppendTable`? If so, I think that would solve our problem; if not, should I create a new issue? I'd be happy to provide more details if it would help.

krlmlr:
Thanks. That's the plan, we don't have a timeline though.

krlmlr:
For now, please work around with `dbWriteTable(append = TRUE)` .

arthur-holtz:
Good to know, thank you for the info!

For what it's worth, I get the same error when using `dbWriteTable(append = TRUE)`. The only way I've managed to get it working is by overriding the definitions of 2 functions in the library (details below).

I imagine this isn't an ideal fix since I'm completely bypassing some checks that probably exist for good reason, but it's a working proof of concept!
1. [Comment out this](https://github.com/r-dbi/odbc/blob/main/R/Table.R#L134) in `dbAppendTable`, which you mentioned above.
2. [...and this](https://github.com/r-dbi/odbc/blob/004bc3f31f218b3c83085566512984ab835c36b8/R/Table.R#L65-L68) in `odbc_write_table`, which is attempting to create a table that already exists because `dbExistsTable` incorrectly returns `FALSE`.

detule:
Hi there, Arthur.

dbExistsTable is indeed a special snowflake, but usually can be massaged into working.  Having said that, and i havent thought about this too much, but rather than commenting out lines, you should be able to extend the S4 methods to give you the behavior you are looking for specific to your back-end.  Even if it's a hack like always return true.


Which database are you having issues with?

arthur-holtz:
[...]

PR #691, opened 2023-12-19
dbAppendTable: do not check for existence twice

detule:
Hi - 

This is a performance enhancement when using `dbAppendTable`.

We have seen recently that calls to `dbExistsTable` can be costly.  Avoid calling it twice when using the `dbAppendTable` API.


detule:
Not in love with the argument name or style.

Feedback welcome!

hadley:
What if we just remove `stopifnot(dbExistsTable(conn, name))` from `dbAppendTable()`? If the user calls it incorrectly on a table that doesn't exist, they'll now get an error from the database rather than R, but I think that's fine. We might also be able to remove the explicit `dbExistsTable()` check in `odbc_write_table()` in a similar way, but I think that would require a bit more analysis. (And maybe a little wrapping with `withCallingHandlers()` to ensure the user is adequately informed about what's happening).

We might instead (or also?) consider moving the append part of `odbc_write_table()` into `dbAppendTable()` so that `dbWriteTable()` becomes mostly a wrapper around `dbExists()`/`dbRemoveTable()` + `dbCreateTable()` + `dbAppendTable()`.

detule:
Hey Hadley, Simon:

I like the idea of letting the DB report the error - made the change, thanks.
I vote to tackle the dbWrite->dbAppend, rather than dbAppend->dbWrite task / refactor in a separate ticket - though happy to also tackle here.

hadley (approved):
I think it's fine to do the bigger change in separate PR, but you might want to note that this one fixes #480, and if you don't start work on the next change right away, make an issue so we don't forget about it.

simonpcouch (approved):
I'm on board!🙂

hadley:
I'll merge this just so it's in the next release.

detule:
Sorry about that - figured out why I went with the new argument to `odbc_write_table` to begin with - it's because `dbWriteTable`, per the DBI spec, should create the table on-the-fly if it does not exist ( unlike `dbAppendTable` where the expectation is that it fails ).

Given that the release is out the door, i might spend a bit of time to try and fit this within the `dbAppendTable`/`dbWriteTable` refactor we discussed.

hadley:
No problems; hopefully the fix will fall in a fairly straightforward way once `dbWriteTable()` is implemented in terms of `dbCreateTable()` and `dbAppendTable()`.

detule:
Thanks for the patience / gave it a shot.  Let me know if anything looks suspect.
Will re-run pipeline once upstream issues have been resolved.

R/dbi-table.R
```diff
@@ -152,21 +117,71 @@ setMethod("dbWriteTable", c("OdbcConnection", "SQL", "data.frame"),
...
+#' chained together from needlessly iterating across the columns of
+#' the data.frame.  It is unlikely that users would ever need to set ( or
+#' interogate ) this attribute directly.
 #' @inheritParams DBI::dbReadTable
 #' @export
 setMethod("sqlData", "OdbcConnection",
   function(con, value, row.names = NA, ...) {
+
+    chk <- attr(value, ".odbc.transformed", exact = TRUE)
+    if ( !is.null(chk) && chk == TRUE ) return(value)
```
hadley, line 183:
```suggestion
    if (isTRUE(chk)) return(value)
```

R/dbi-table.R
```diff
@@ -152,21 +117,71 @@ setMethod("dbWriteTable", c("OdbcConnection", "SQL", "data.frame"),
...
+      )
+    }
+
     invisible(NA_real_)
-  }
-)
+  })
 
 #' @rdname DBI-methods
+#' @details If the `value` argument has an attribute `.odbc.transformed` that
```
hadley, line 171: IMO since this is purely for internal use, you could omit from the docs.

R/dbi-table.R
```diff
@@ -36,10 +36,11 @@ odbc_write_table <- function(conn,
                              overwrite = FALSE,
                              append = FALSE,
                              temporary = FALSE,
-                             row.names = NA,
+                             row.names = NULL,
```
hadley, line 39: I think this will fix #737

simonpcouch:
Closes #737.

R/dbi-table.R
detule: Thanks, done.

R/dbi-table.R
detule: Thanks, done.

NEWS.md
```diff
@@ -35,9 +37,6 @@
 * SQL Server: correctly enumerate schemas across databases in connections pane
   (@detule, #527).
 
-* SQL Server: now uses column type `"BIGINT"` integer64 objects
```
hadley, line 38: @detule I think you deleted this accidentally?

detule: Ugh - sorry about that! Will fix tonight, unless someone beats me to it.

hadley: Let me do it since I'm polishing news for release.

R/dbi-table.R
```diff
@@ -152,21 +117,65 @@ setMethod("dbWriteTable", c("OdbcConnection", "SQL", "data.frame"),
...
+      if (!is.null(fieldDetails) && nrow(fieldDetails) == nparam) {
+        result_describe_parameters(rs@ptr, fieldDetails)
+      }
+
+      values <- sqlData(conn, row.names = row.names, value[, , drop = FALSE])
+      if (is.na(batch_rows)) {
+        batch_rows <- NROW(value)
+        if (batch_rows == 0) {
+          batch_rows <- 1
+        }
```
krlmlr, line 158: Not sure how this condition can ever be `TRUE`, with the check in line 137 in this PR.

R/dbi-table.R
```diff
@@ -152,21 +117,65 @@ setMethod("dbWriteTable", c("OdbcConnection", "SQL", "data.frame"),
...
+    fieldDetails <- tryCatch({
+      details <- odbcConnectionColumns(conn, name, exact = TRUE)
+      details$param_index <- match(details$name, colnames(value))
+      details[!is.na(details$param_index) & !is.na(details$data_type), ]
+    },
+    error = function(e) {
+      return(NULL)
+    })
+
+    if (nrow(value) > 0) {
```
krlmlr, line 137: Perhaps worth rephrasing as

```r
out <- invisible(NA_real_)

if (nrow(value) == 0) {
  return(out)
}
```

to avoid the indent?

krlmlr (commented):
Thanks, great! Do we still need the custom `dbWriteTable()` implementation?

PR #694, opened 2023-12-19
remove `driver_info` class from `dbGetInfo()` output

### LINKED ISSUE #667: Eliminate `driver_info` class
It looks vestigial and I don't think it does anything any more.

simonpcouch:
Closes #667.

Introduced in https://github.com/r-dbi/odbc/commit/9585993b6eeb42621eaebd6029dd9cfe63535400. It didn't seem to have had any methods attached to it then, either.


R/dbi-connection.R
```diff
@@ -220,7 +220,7 @@ setMethod("dbListFields", c("OdbcConnection", "character"),
 setMethod("dbGetInfo", "OdbcConnection",
   function(dbObj, ...) {
     info <- connection_info(dbObj@ptr)
-    structure(info, class = c(info$dbms.name, "driver_info", "list"))
```
hadley, line 223: Since it's now just a "list", you can yeet the class attribute 😄 (and same below)

simonpcouch: So the `info$dbms.name` class is unneeded as well?

hadley: Oh hmmm, that's a good question. But I don't think so since that's usually an S4 class and here we're making an S3 class.

hadley: I double checked this by looking at the `NAMESPACE` to see what S3 generics we have methods for. As far as I can tell, the only S3 generic is `odbcDataType()`, and that's always called on the connection object.

hadley (approved):
This PR slaps.

simonpcouch:
😂😂😂

PR #695, opened 2023-12-19
Improve databricks support for manual auth

### LINKED ISSUE #690: Passing `pwd` as an argument does not work in `databricks()`
Then passing `pwd` as an argument to `databricks()` I get the following error:

```
Error in .local(drv, ...) : 
  formal argument "pwd" matched by multiple actual arguments
```

Here is a reprex:

```r
library(DBI)

con <- dbConnect(
  odbc::databricks(), 
  pwd = Sys.getenv("DIFFERENT_TOKEN"),
  httpPath = "/sql/1.0/warehouses/300bd24ba12adf8e"
  )
```

This occurs regardless if DATABRICKS_TOKEN is set or not

hadley:
Fixes #690

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,11 @@
 # odbc (development version)
 
+* `databricks()` now works with manually supplied `pwd` and `uid` (#690).
+
 * Added a function `odbcListConfig()` to help locate configuration files on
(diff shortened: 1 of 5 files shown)


atheriel (approved):
LGTM!

PR #697, opened 2023-12-19
Better way of building args object

### LINKED ISSUE #696: Driver Keyword syntax error with new version of odbc with SQL SERVER
got this error suddenly with new odbc with either freeTDS and Microsoft drivers on linux

```
conn <- odbc::dbConnect(odbc::odbc(), dsn = "MSSQLTEST;usr=***;pwd=***")
Error: nanodbc/nanodbc.cpp:1135: 00000
[unixODBC][Driver Manager]DRIVER keyword syntax error 
```

my odbc file is

```
[MSSQLTEST]
Driver = /usr/lib64/libtdsodbc.so.0
Description = Connection to SQL server
Trace           = Yes
TraceFile       = /tmp/odbctest.trace
Server = ***
Database = ***
Port = ***
User = ***
Password = ***
```

Could not find any change that would change the connection process

Also, I can connect if I provide all the arguments instead of dsn.
It also workss fine with dsn for MYSQL and POSTGRE

hadley:
Fixes #696

### DIFF
--- a/R/odbc-connection.R
+++ b/R/odbc-connection.R
@@ -68,7 +68,7 @@ OdbcConnection <- function(
 
 build_connection_string <- function(.string = NULL, ...) {
   args <- compact(list(...))
-  # check_args(args)
+  check_args(args)
 
   needs_escape <- grepl("[{}(),;?*=!@]", args) &
     !grepl("^\\{.*\\}$", args) &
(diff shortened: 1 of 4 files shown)


simonpcouch (approved):
This looks solid.

Only thing that feels worth bringing up is that we don't do any automatic escaping for fields in `.connection_string`. I take it this is on purpose as we assume folks know what they're doing when they pass that argument?

hadley:
Right. And I think automatic escaping is also impossible because to parse that string into name-value pairs that could be escaped, we'd have to assume that it's already escaped correctly 😄

PR #703, opened 2023-12-20
add notes on snowflake configuration

simonpcouch:
❄️❄️❄️

Marking as draft as I'm still troubleshooting an SSL error.


simonpcouch (commented):
Most of this seems a little on the boundary of this vignette and `setup.Rmd`, but `develop.Rmd` seems to dig a bit more into the oddities of specific DBMS/driver interactions so I've placed it here.

There goes a day.😵‍💫

hadley (approved):
⛄ Thanks for figuring this out! ⛄

PR #722, opened 2024-01-11
Fix DBItest failures

hadley:
R-devel failure is https://github.com/RcppCore/Rcpp/issues/1290; nothing we can do about it, but Rcpp release should fix soon.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -53,7 +53,9 @@ parse_size <- function(x) {
   )
 }
 
-id_field <- function(id, field, default = NULL) {
+id_field <- function(id, field = c("catalog", "schema", "table")) {
+  arg_match(field)
+
   if (length(id@name) == 1) {
     switch(field,
       catalog = NULL,
--- a/R/utils.R
+++ b/R/utils.R
@@ -54,10 +54,26 @@ parse_size <- function(x) {
 }
 
 id_field <- function(id, field, default = NULL) {
-  if (field %in% names(id@name)) {
-    id@name[[field]]
+  if (length(id@name) == 1) {
+    switch(field,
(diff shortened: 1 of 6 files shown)


tests/testthat/test-driver-sql-server.R
```diff
@@ -58,6 +58,7 @@ test_that("SQLServer", {
     "list_fields_wrong_table",
     "list_fields_quoted",
     "list_fields_object",
+    "list_objects_features",
```
hadley, line 61: I'm not sure why this worked before, but it was skipped on all other backends.

R/utils.R
```diff
@@ -53,11 +53,31 @@ parse_size <- function(x) {
...
 
-id_field <- function(id, field, default = NULL) {
-  if (field %in% names(id@name)) {
-    id@name[[field]]
+id_field <- function(id,
+                     field = c("catalog", "schema", "table"),
+                     error_call = caller_env()) {
+  arg_match(field, error_call = error_call)
+
+  if (length(id@name) == 1) {
```
hadley, line 61: This is the main fix; we can no longer assume that the `Id()` object is named, so we instead rely on position.

tests/testthat/test-driver-sql-server.R
detule: Yep - looks like fixed by @krlmlr on `DBItest::main`.  Maybe we bring back once that one makes its way to cran

detule (approved):
Looks good - thanks!

krlmlr (commented):
The newly skipped tests might work better with dev DBItest, soon on CRAN. It's worth revisiting them then.

PR #725, opened 2024-01-12
Refactor database tests into individual test blocks

hadley:
* Add `test_con()` helper and move individual database tests into own blocks
* Add `local_table()` helper

Fixes #634

PR #747, opened 2024-01-19
include temp tables in `dbListTables()` output for MSSQL

### LINKED ISSUE #509: MSQL: dbListTables does not return temporary tables

`odbc::dbListTables` does not return temporary tables unless `catalog_name` parameter is provided. DBI specification and documentation suggests that temporary tables should be returned when default parameters are used.  

### Database

Microsoft SQL Server Version: 13.00.7016

### Reproducible Example

```r
library(odbc)
library(DBI)
con <- dbConnect(odbc::odbc(), dsn = "MicrosoftSQL")

DBI::dbExecute(con, 
               "CREATE TABLE ##global_temp (
                  colA varchar(10)
               )")

#output does not include ##global_temp table
odbc::dbListTables(con)

#output includes ##global_temp table
odbc::dbListTables(con, catalog_name = "tempdb")

```
<details>
<summary>Session Info</summary>

```r
devtools::session_info()
─ Session info ──────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.2.1 (2022-06-23 ucrt)
 os       Windows 10 x64 (build 19044)
 system   x86_64, mingw32
 ui       RStudio
 language (EN)
 collate  English_Australia.utf8
 ctype    English_Australia.utf8
 tz       Australia/Sydney
 date     2022-09-14
 rstudio  2022.02.3+492 Prairie Trillium (desktop)
 pandoc   NA

─ Packages ──────────────────────────────────────────────────────────────────────────────────────────
 package     * version date (UTC) lib source
 bit           4.0.4   2020-08-04 [1] CRAN (R 4.1.3)
 bit64         4.0.5   2020-08-30 [1] CRAN (R 4.1.3)
 blob          1.2.3   2022-04-10 [1] CRAN (R 4.1.3)
 brio          1.1.3   2021-11-30 [1] CRAN (R 4.1.3)
 cachem        1.0.6   2021-08-19 [1] CRAN (R 4.1.3)
 callr         3.7.0   2021-04-20 [1] CRAN (R 4.1.3)
 cli           3.3.0   2022-04-25 [1] CRAN (R 4.1.3)
 clipr         0.8.0   2022-02-22 [1] CRAN (R 4.1.3)
 crayon        1.5.1   2022-03-26 [1] CRAN (R 4.1.3)
 DBI         * 1.1.3   2022-06-18 [1] CRAN (R 4.1.3)
 desc          1.4.1   2022-03-06 [1] CRAN (R 4.1.3)
 devtools    * 2.4.3   2021-11-30 [1] CRAN (R 4.1.3)
 dplyr       * 1.0.9   2022-04-28 [1] CRAN (R 4.1.3)
 ellipsis      0.3.2   2021-04-29 [1] CRAN (R 4.1.3)
 fansi         1.0.3   2022-03-24 [1] CRAN (R 4.1.3)
 fastmap       1.1.0   2021-01-25 [1] CRAN (R 4.1.3)
 fs            1.5.2   2021-12-08 [1] CRAN (R 4.1.3)
 generics      0.1.2   2022-01-31 [1] CRAN (R 4.1.3)
 glue          1.6.2   2022-02-24 [1] CRAN (R 4.1.3)
 hms           1.1.1   2021-09-26 [1] CRAN (R 4.1.3)
 lifecycle     1.0.1   2021-09-24 [1] CRAN (R 4.1.3)
 magrittr      2.0.3   2022-03-30 [1] CRAN (R 4.1.3)
 memoise       2.0.1   2021-11-26 [1] CRAN (R 4.1.3)
 odbc          1.3.3   2021-11-30 [1] CRAN (R 4.1.3)
 pillar        1.7.0   2022-02-01 [1] CRAN (R 4.1.3)
 pkgbuild      1.3.1   2021-12-20 [1] CRAN (R 4.1.3)
 pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.1.3)
 pkgload       1.2.4   2021-11-30 [1] CRAN (R 4.1.3)
 prettyunits   1.1.1   2020-01-24 [1] CRAN (R 4.1.3)
 processx      3.5.3   2022-03-25 [1] CRAN (R 4.1.3)
 ps            1.7.0   2022-04-23 [1] CRAN (R 4.1.3)
 purrr         0.3.4   2020-04-17 [1] CRAN (R 4.1.3)
 R6            2.5.1   2021-08-19 [1] CRAN (R 4.1.3)
 Rcpp          1.0.8.3 2022-03-17 [1] CRAN (R 4.1.3)
 readr       * 2.1.2   2022-01-30 [1] CRAN (R 4.1.3)
 remotes       2.4.2   2021-11-30 [1] CRAN (R 4.1.3)
 rlang         1.0.2   2022-03-04 [1] CRAN (R 4.1.3)
 rprojroot     2.0.3   2022-04-02 [1] CRAN (R 4.1.3)
 rstudioapi    0.13    2020-11-12 [1] CRAN (R 4.1.3)
 sessioninfo   1.2.2   2021-12-06 [1] CRAN (R 4.1.3)
 testthat      3.1.4   2022-04-26 [1] CRAN (R 4.1.3)
 tibble        3.1.7   2022-05-03 [1] CRAN (R 4.1.3)
 tidyselect    1.1.2   2022-02-21 [1] CRAN (R 4.1.3)
 tzdb          0.3.0   2022-03-28 [1] CRAN (R 4.1.3)
 usethis     * 2.1.6   2022-05-25 [1] CRAN (R 4.1.3)
 utf8          1.2.2   2021-07-24 [1] CRAN (R 4.1.3)
 vctrs         0.4.1   2022-04-13 [1] CRAN (R 4.1.3)
 withr         2.5.0   2022-03-03 [1] CRAN (R 4.1.3)
─────────────────────────────────────────────────────────────────────────────────────────────────────

```
</details>

simonpcouch:
Closes #509.

``` r
library(DBI)
library(odbc)

con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA", 
                 pwd = Sys.getenv("sqlServerPass"))

dbExecute(con, 
               "CREATE TABLE ##global_temp (
                  colA varchar(10)
               )")
#> [1] 0


"##global_temp" %in% odbc::dbListTables(con)
#> [1] TRUE

dbDisconnect(con)
```

<sup>Created on 2024-01-19 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


R/driver-sql-server.R
```diff
@@ -81,6 +81,46 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
...
+           table_type = NULL,
+           ...) {
+    res <- callNextMethod(
+      conn = conn,
+      catalog_name = catalog_name,
+      schema_name = schema_name,
+      table_name = table_name,
+      table_type = table_type,
+      ...
+    )
```
hadley, line 105:
```suggestion
    res <- callNextMethod()
```

R/driver-sql-server.R
```diff
@@ -81,6 +81,46 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
...
+
+    if (is.null(catalog_name)) {
+      res_temp <- callNextMethod(
+        conn = conn,
+        catalog_name = "tempdb",
+        schema_name = schema_name,
+        table_name = table_name,
+        table_type = table_type,
+        ...
+      )
```
hadley, line 115:
```suggestion
      res_temp <- callNextMethod(
        catalog_name = "tempdb",
      )
```

I am no `callNextMethod()` expert but I believe the main way it works is MAGIC and you only need to supply arguments that are different to the caller.

And I think you might also need `schema_name = "dbo"` (Reading between the lines from https://stackoverflow.com/questions/8311959/specifying-schema-for-temporary-tables and my recollection of how `dbExistsTable()` was previously implemented)

hadley (commented):
Turn your code in the comment into a test? Take at look at #725 for the direction in which tests are moving.

simonpcouch:
It seems like `dbListTables()` will also, by default, exclude entries in [the `msdb` catalog](https://learn.microsoft.com/en-us/sql/relational-databases/databases/msdb-database?view=sql-server-ver16). I haven't "fixed" that in this PR.

R/driver-sql-server.R
simonpcouch: Woah😮

Works well for the first call, `callNextMethod()`, but writing `callNextMethod(catalog_name = "tempdb")` for the second results in:

```
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'conn' in selecting a method for function 'odbcConnectionTables': argument "conn" is missing, with no default
```

Seems like it needs the arguments preceding any (in position) that are manually set; adding `conn = con` and leaving the rest unspecified seems to work but will confirm.

------

re: dbo, I did notice that `tempdb` has analogous `INFORMATION_SCHEMA` and `sys` schemas to those in `master` (or any non-default catalog). We currently include tables in those schemas from those other databases by default; should we exclude them for `tempdb` by setting `schema_name = "dbo"`?

R/driver-sql-server.R
simonpcouch: 🪄 [7d6478f](https://github.com/r-dbi/odbc/pull/747/commits/7d6478fa941ba403b3986ef353fc1f4d365284c2)

simonpcouch:
Sounds good on adding that test! When that PR is merged, we can add the following chunk [here](https://github.com/r-dbi/odbc/blob/070378e17c3534f3acbc4fc56618cb6e3fbddb54/tests/testthat/test-driver-sql-server.R#L309-L310):

``` r
# ensure we include temp tables in dbListTables output (#509)
expect_true(locTblName %in% dbListTables(con))
```

If you'd prefer we not wait, could also just add the following in the meantime:

``` r
test_that("temp tables are included in dbListTables() output (#509)", {
  con <- DBItest:::connect(DBItest:::get_default_context())
  dbWriteTable(con, "##global_temp", mtcars)
  on.exit({
    dbRemoveTable(con, "##global_temp")
  })
  expect_true("##global_temp" %in% odbc::dbListTables(con))
})
```

R/driver-sql-server.R
```diff
@@ -81,6 +81,32 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
...
+  function(conn,
+           catalog_name = NULL,
+           schema_name = NULL,
+           table_name = NULL,
+           table_type = NULL,
+           ...) {
+    res <- callNextMethod()
+
+    if (is.null(catalog_name)) {
+      res_temp <- callNextMethod(conn = conn, catalog_name = "tempdb")
```
hadley, line 101:
```suggestion
      res_temp <- callNextMethod(catalog_name = "tempdb")
```

R/driver-sql-server.R
```diff
@@ -81,6 +81,32 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
...
+setMethod("dbListTables", "Microsoft SQL Server",
+  function(conn,
+           catalog_name = NULL,
+           schema_name = NULL,
+           table_name = NULL,
+           table_type = NULL,
+           ...) {
+    res <- callNextMethod()
+
+    if (is.null(catalog_name)) {
```
hadley, line 100:
```suggestion
    if (is.null(catalog_name) && is.null(schema_name)) {
```

Since IIUC temp tables live in their own schema too

hadley (approved):

R/driver-sql-server.R
simonpcouch: Any thoughts on https://github.com/r-dbi/odbc/pull/747#discussion_r1459638726? I see the error noted there with this change.

R/driver-sql-server.R
simonpcouch: Some thoughts here in https://github.com/r-dbi/odbc/pull/747#discussion_r1459638726—I'm on board if we indeed want to exclude those other schemas though.

R/driver-sql-server.R
hadley: Sorry, somehow missed this and every other comment 😬 

I am super surprised that you get that error because the docs say "the arguments have no effect on selecting the next method.", but they clearly do! But maybe `callNextMethod()` is confused because we've already called it once? I wish I understood what was happening here better, but it's probably not worth the time to explore, and if your code works, then go with it.

I do think we should set `schema_name = "dbo"` too.

R/driver-sql-server.R
hadley: I think we should do both maybe? I'm not 100% sure what people would expect here, but given that temporary tables exist in their own catalog + schema, I think if you request a specific catalog/schema, we should respect that.

NEWS.md
```diff
@@ -21,6 +24,7 @@
 * On MacOS and Linux, the package will now automatically set the `ODBCSYSINI`
   environmental variable when using the unixODBC driver manager. `ODBCSYSINI`
   will not be changed if it exists already (@simonpcouch, #709).
+>>>>>>> main
```
hadley, line 27: 🚨

Issue #699, opened 2023-12-19
`odbcConnectionTableTypes()` is not exported


simonpcouch:
`odbcConnectionTableTypes()` has a man-page:

https://github.com/r-dbi/odbc/blob/95e8b69833fabcbc78888a917e83cb0e9af78ddb/R/odbc-connection.R#L427-L429

and some exported functions refer to it in theirs:

https://github.com/r-dbi/odbc/blob/95e8b69833fabcbc78888a917e83cb0e9af78ddb/R/odbc-connection.R#L297

but the function isn't exported.

Is this on purpose?

hadley:
Looks like `odbcConnectionSchemas`, `odbcConnectionCatalogs`, `odbcConnectionTables` are in the same situation.

But `odbcConnectionActions()`, `odbcConnectionColumns()`, and `odbcConnectionIcon()` are exported.

Many of these are only used for the connections pane.

detule:
IMO: I think we leave them unexported, unless there is a need.

`odbcConnectionSchemas/Catalogs/Tables/Types` are all routed to the same odbc API endpoint and differ only in the magic parameters that are used.  It used to be the case that whenever we needed to interact with it we used to always have to call `connection_sql_tables`.  I created the separate S4 methods so as to make it easier to create back-end specific implementations for individual use cases ( listing tables, types, schemas, etc ), without worrying if doing that will jack up all the different ways that `connection_sql_tables` gets called for that back-end.

If I could go back, I would probably un-export `odbcConnectionColumns`, but cat is out of the bag there.

hadley:
We could deprecate `odbcConnectionColumns()`, pointing to `dbListFields()`, and then remove it in a couple of years. I think that's worth doing.

And then we should remove the docs for the other functions, or at least minimise them, and prevent them from becoming `.Rd` topics.

detule:
Sounds good on deprecating  the `odbcConnectionColumns` export with enough of a headsup, as long as we continue to carry it around internally.  In particular, we need its extended output to handle writing with some drivers/back-ends.

simonpcouch:
re: deprecating `odbcConnectionColumns()`, I can imagine two approaches:

1) Use `lifecycle::deprecate_soft()` directly and just ignore the warnings as developers of the package for the meantime until `odbcConnectionColumns()` can be unexported, or
[...]

PR #749, opened 2024-01-22
remove docs on unexported `odbcConnection*()` functions

simonpcouch:
Related to #699, but does not close.


R/driver-oracle.R
```diff
@@ -31,15 +31,11 @@ setMethod("sqlCreateTable", "Oracle",
...
-#' @rdname Oracle
-#' @description
-#' ## `odbcConnectionTables()`
-#'
-#' Query, rather than use `SQLTables` ODBC API for performance reasons.
-#' Main functional difference between the implementation of `SQLTables` in the
-#' OEM driver and the query below is that the OEM implementation also looks
-#' through the synonyms. Given the performance reports, we sacrifice the
-#' synonym look-through for better execution time.
+# Query, rather than use `SQLTables` ODBC API for performance reasons.
```
hadley, line 34: I think you could keep these in the existing docs but reframe slightly to be more about the implementation details.

hadley (approved):

R/driver-oracle.R
simonpcouch: You got it! 643ef41.

PR #750, opened 2024-01-22
deprecate `odbcConnectionColumns()`

### LINKED ISSUE #699: `odbcConnectionTableTypes()` is not exported

simonpcouch:
Closes #699.

Note that this PR doesn't touch the `odbcConnectionActions` and `odbcConnectionIcon` functions, which are also exported.


R/odbc-connection.R
```diff
@@ -140,11 +140,15 @@ NULL
 
 #' odbcConnectionColumns
 #'
+#' @description
+#' `r lifecycle::badge("deprecated")`
```
hadley, line 144: I'd suggest you recommend `dbListFields()` here too.

R/odbc-connection.R
```diff
@@ -176,21 +180,33 @@ NULL
 #'
 #' @rdname odbcConnectionColumns
```
hadley, line 181: I think you could `@keywords internal` and make sure it's not include in the pkgdown ref too.

hadley (approved):
This approach looks good to me. I think you could also do `odbcConnectionActions()` and `odbcConnectionIcon()` in this PR too if you wanted.

R/odbc-connection.R
simonpcouch: Sounds good! 5096a4b.

R/odbc-connection.R
simonpcouch: You got it! 9abd87a.

simonpcouch:
In deprecating `odbcConnectionActions()` in 127387a I didn't use the temporary underscore pattern as we make no "internal" usage of the function. It doesn't look me to like this is called anywhere in the IDE, so it seems fine to deprecate as usual.

simonpcouch:
Same goes for [odbcConnectionIcon()](https://github.com/r-dbi/odbc/pull/750/commits/73aa7e252913a7e9784dd18f6b83e8d6f301ea44).

hadley:
Aren't they both called in `on_connection_opened()`?

simonpcouch:
Ah, I missed that somehow. Back to it.🐢

simonpcouch:
Transitioned `odbcConnectionActions()` and `odbcConnectionIcon()` from S3 methods, as they only have a default method and our change to the underscore naming would break dispatch to any methods supplied by other packages anyway.

hadley (approved):

PR #753, opened 2024-01-26
automatically configure spark on apple

### LINKED ISSUE #651: Look into apple + spark drive config issue
https://solutions.posit.co/connections/db/databases/databricks/#troubleshooting-apple-macos-users

Can we fix this automatically?

simonpcouch:
Some machinery to close #651.

Opening as draft as I've yet to integrate this into any functionality yet. Would `configure_spark()` be best situated as:

1) An exit handler on `dbConnect<OdbcDriver>()` (or `dbConnect<DatabricksOdbcDriver>()`?) that greps for an error message like the one noted in https://solutions.posit.co/connections/db/databases/databricks/#troubleshooting-apple-macos-users and
	a) automatically runs the `configure_spark()` function
	 b) points the user to `configure_spark()` in a rethrown error
2) A helper that's always run inside of one of those `dbConnect()` methods
3) Something else?

I also imagine some sort of argument or env var that users could set to override `configure_unixodbc_spark(unixodbc_install)` and `configure_unixodbc_spark(spark_config)`, but waited to implement pending a decision on the above.

Moving on to tests for now.


hadley:
I think we'd run this from inside `databricks()`

R/utils.R
```diff
@@ -165,3 +169,123 @@ random_name <- function(prefix = "") {
...
+      )
+    }
+  )
+
+  locate_install_unixodbc()
+}
+
+install_unixodbc_libs <- function() {
+  source("https://mac.R-project.org/bin/install.R")
+  install.libs("unixodbc")
```
simonpcouch, line 239: I personally wasn't able to successfully run these lines on either my work computer (M1, existing unixodbc install) or personal computer (Intel, no existing unixodbc install). Both failed with errors:

```
various/file/paths: Can't restore time
```

I _think_ this approach would also require some ODBCSYSINI magic, as well?

R/utils.R
```diff
@@ -165,3 +169,122 @@ random_name <- function(prefix = "") {
...
+}
+
+locate_install_unixodbc <- function() {
+  unixodbc_prefix <- tryCatch(
+    system("odbc_config --lib-prefix", intern = TRUE),
+    error = function(e) character(0)
+  )
+  if (!identical(unixodbc_prefix, character(0))) {
+    return(paste0(unixodbc_prefix, "/libodbcinst.dylib"))
+  }
```
hadley, line 205:
```suggestion
  tryCatch(
    {
      unixodbc_prefix <- system("odbc_config --lib-prefix", intern = TRUE)
      return(paste0(unixodbc_prefix, "/libodbcinst.dylib"))
    },
    error = function(e) {}
  )
```

Maybe?

R/utils.R
hadley: I think it's fine to just tell the user what they need. 

But maybe checking for `unixODBC` should happen `onLoad()` since this will affect all connection?

tests/testthat/test-utils.R
```diff
@@ -54,3 +54,117 @@ test_that("getSelector", {
   expect_equal(getSelector("mykey", "%", exact = TRUE), " AND mykey LIKE '%'")
   expect_equal(getSelector("mykey", "%", exact = FALSE), " AND mykey LIKE '%'")
 })
+
+test_that("configure_spark() returns early on windows", {
+  local_mocked_bindings(is_macos = function() {FALSE})
+
+  res <- configure_spark()
+
+  expect_equal(res, NULL)
```
hadley, line 63:
```suggestion
  expect_equal(configure_spark(), NULL)
```

Because this test isn't that important, I'd keep it as compact as possible.

R/utils.R
simonpcouch: Heard. Just transitioned from trying to install unixODBC to erroring if not installed in 60cc4da5.

Agreed that checking for unixODBC on load is reasonable. Should that happen in this PR or a separate one?

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # odbc (development version)
 
+* `databricks()` will now automatically configure the needed driver and driver
+  manager (#651).
```
hadley, line 4:
```suggestion
  manager on macOS (#651).
```

?

R/utils.R
```diff
@@ -165,3 +169,106 @@ random_name <- function(prefix = "") {
...
+  unixodbc_install <- locate_install_unixodbc()
+  if (identical(unixodbc_install, character(0))) {
+    abort(
+      c(
+        "Unable to locate the unixODBC driver manager.",
+        i = "Please install unixODBC using Homebrew with `brew install unixodbc`."
+      ),
+      call = call
+    )
+  }
```
hadley, line 188: Should this go into the generic `dbConnect()` method? It will affect all drivers, not just databricks.

R/utils.R
```diff
@@ -165,3 +169,106 @@ random_name <- function(prefix = "") {
...
 }
+
+# apple + spark drive config (#651) --------------------------------------------
+configure_spark <- function(call = caller_env()) {
+  if (!is_macos()) {
+    return(invisible())
+  }
+
+  unixodbc_install <- locate_install_unixodbc()
+  if (identical(unixodbc_install, character(0))) {
```
hadley, line 180:
```suggestion
  if (length(unixodbc_install) == 0) {
```

I think that makes the interface a bit clearer to me.

R/utils.R
```diff
@@ -165,3 +169,106 @@ random_name <- function(prefix = "") {
...
+  )
+
+  odbcinstlib_lines <- grepl("^ODBCInstLib=", spark_lines)
+  odbcinstlib <- spark_lines[odbcinstlib_lines]
+  odbcinstlib_config <- paste0("ODBCInstLib=", unixodbc_install)
+  if (identical(odbcinstlib, character(0))) {
+    spark_lines <- c(spark_lines, odbcinstlib_config)
+  } else {
+    spark_lines[odbcinstlib_lines] <- odbcinstlib_config
+  }
```
hadley, line 260: I found this a little hard to reason about, and since you're repeating the code in the next block, I'd suggest pulling it out into a separate function like `replace_or_append(lines, pattern, replacement)`

R/utils.R
```diff
@@ -165,3 +169,106 @@ random_name <- function(prefix = "") {
...
+  manager_encoding_lines <- grepl("^DriverManagerEncoding=", spark_lines)
+  manager_encoding <- spark_lines[manager_encoding_lines]
+  manager_encoding_config <- "DriverManagerEncoding=UTF-16"
+  if (identical(manager_encoding, character(0))) {
+    spark_lines <- c(spark_lines, manager_encoding_config)
+  } else {
+    spark_lines[manager_encoding_lines] <- manager_encoding_config
+  }
+
+  writeLines(spark_lines, spark_config)
```
hadley, line 271: It feels a little weird to do this everytime, even if we don't actually change anything. But I'm not sure it's worth bothering.

R/utils.R
```diff
@@ -165,3 +169,106 @@ random_name <- function(prefix = "") {
...
+  if (identical(spark_config, character(0))) {
+    abort(
+      c(
+        "Unable to locate the needed spark ODBC driver.",
+        i = "Please install the needed driver from https://www.databricks.com/spark/odbc-drivers-download."
+      ),
+      call = call
+    )
+  }
+
```
hadley, line 200: We might also want to check that `spark_config` is writeable.

R/utils.R
```diff
@@ -165,3 +169,106 @@ random_name <- function(prefix = "") {
...
+      c(
+        "Unable to locate the unixODBC driver manager.",
+        i = "Please install unixODBC using Homebrew with `brew install unixodbc`."
+      ),
+      call = call
+    )
+  }
+
+  spark_config <- locate_config_spark()
+  if (identical(spark_config, character(0))) {
```
simonpcouch, line 191:
```suggestion
  if (length(spark_config) == 0) {
```

R/utils.R
simonpcouch: I had wondered the same thing. The only reason I didn't is because this check is both relatively invasive and ad-hoc, but I'm game to move it over if you are.

R/utils.R
simonpcouch: I feel that. When resolving the comment that we ought to check if the file is writeable, I went ahead and added this in as well so that we don't bother checking whether writeable + trying to write if there wouldn't be any changes to make anyway.

hadley (approved):
Looks great!

PR #756, opened 2024-01-30
transition `odbcDataType()` to S4

### LINKED ISSUE #701: Switch `odbcDataType` to S4
For consistency. And consider documentation for consistency with other generics.

simonpcouch:
Closes #701.🐙


DESCRIPTION
```diff
@@ -49,6 +49,7 @@ RoxygenNote: 7.3.0
 SystemRequirements: GNU make, An ODBC3 driver manager and drivers.
 Collate: 
     'RcppExports.R'
+    'aaa-odbc-data-type.R'
```
simonpcouch, line 52: 😞

I wasn't able to find any other places where we needed a workaround to define generics before methods. We could also just delete this file and place its contents in `odbc-connection.R`?

R/aaa-odbc-data-type.R
```diff
@@ -15,61 +18,51 @@
...
-#' ```
-#' con <- dbConnect(odbc::odbc(), "FooConnection")
-#' dbGetInfo(con)$dbms.name
-#' #> [1] "foo"
+#' a new method. The object type for your method will be the database name
+#' retrieved by `dbGetInfo(con)$dbms.name`. Use the documentation provided with
+#' your database to determine appropriate values for each R data type.
 #'
-#' `odbcDataType.foo <- function(con, obj, ...) {
-#'   switch_type(obj,
```
simonpcouch, line 35: This example used this unexported `switch_type()` helper. If we wanted to keep this example, we could export `switch_type()`, but us never having received an issue related to this makes me wonder if folks ever used this code.

hadley: Yeah, I don't think there's a compelling reason to keep including this example.

hadley (approved):
This PR slaps. Fact. 🚫🧢.

PR #760, opened 2024-02-14
snowflake: improve performance on write

### LINKED ISSUE #759: Default to arbitrary database if not specified when creating temp tables on a warehouse (Snowflake / Databricks)
For warehouses like Snowflake / Databricks, could `odbc` reliably assume a reasonable default database? Or do users _always_ need to specify _some_ `catalog` value they have (read) access to, even though the user doesn't necessarily care _where_ the temp is saved?

To reproduce: 

1. First, create a connection:
```r
snowflake_connection_string <- paste(
  'database=', # IMPORTANT: intentionally left blank to enable cross-catalog joins from a single connection object
  'Driver=/opt/rstudio-drivers/snowflake/bin/lib/libsnowflakeodbc_sb64.so',
  'Locale=en-US', 'Server=<my_server>',
  'UID=<my_user>', 'Role=<my_role>', 'Warehouse=<my_warehouse>',
  'AUTHENTICATOR=OAUTH', 'TOKEN=<my_token>'
  sep = ';'
  )

conn <- odbc::dbConnect(
  odbc::odbc(),
  .connection_string = snowflake_connection_string
)
```

2. Then, trying to write a simple temp table:
```r
odbc::dbWriteTable(conn, "mtcars", mtcars, temporary = TRUE)
```
Errors:
```
Error: nanodbc/nanodbc.cpp:1710: 00000
Cannot perform CREATE TABLE. This session does not have a current database. Call 'USE DATABASE', or use a qualified name. 
```

### Mitigations (manual)

1. Set the database at the connection level:
```
odbc::dbExecute(conn, "USE DATABASE <ARBITRARY_DATABASE>")
odbc::dbExecute(conn, "USE SCHEMA PUBLIC")
```
then the `odbc::dbWriteTable()` call above **works**, but is _very_ slow (covered by https://github.com/r-dbi/odbc/issues/480)

2.  Qualify the table name (fast, ~2s):
```
tbl_ref <- DBI::Id(
  catalog = "<ARBITRARY_DATABASE>",
  schema  = "PUBLIC",
  table   = "mtcars"
)
odbc::dbWriteTable(conn, tbl_ref, mtcars, temporary = TRUE)
```

### Alternatives considered

we tried leveraging the global Snowflake database:
```
odbc::dbExecute(conn, "USE DATABASE SNOWFLAKE")
odbc::dbExecute(conn, "USE SCHEMA INFORMATION_SCHEMA")
```

But, errors with:
```
Error in eval(e, env) : nanodbc/nanodbc.cpp:1710: 00000
SQL execution error: Creating table on shared database 'SNOWFLAKE' is not allowed. 
```

<details>
<summary>Database</summary>

```r
> DBI::dbGetInfo(conn)
$dbname
[1] ""

$dbms.name
[1] "Snowflake"

$db.version
[1] "8.5.2"

$username
[1] ""

$host
[1] ""

$port
[1] ""

$sourcename
[1] ""

$servername
[1] "Snowflake"

$drivername
[1] "libsnowflakeodbc_sb64.so"

$odbc.version
[1] "03.52"

$driver.version
[1] "2.25.5.1019"

$odbcdriver.version
[1] "03.80"

$supports.transactions
[1] TRUE

$getdata.extensions.any_column
[1] TRUE

$getdata.extensions.any_order
[1] TRUE
```
</details>

detule:
Snowflake: use current DB/schema, if none are provided, in `odbcConnectionColumns` and when checking for existence in `dbWriteTable`.  

Fixes #759.


meztez:
Do you think this should also be done for the `odbcConnectionTables` method?

dbExistsTable ends up having the same problem when listing tables from all DATABASE / SCHEMA instead of the one that will be used by dbWrite Table?

detule:
Thanks @meztez 

I thought about this - but i think we need to be a bit careful.  We should probably interpret `dbExistsTable(conn, "tblname")` to mean "does `tblname` exist anywhere", not just in the current db/schema.

The method has an ellipsis; one approach is to add a `package::odbc` specific argument - something like `useCurrentSchemaIfUnspecified` that defaults to `FALSE`.  We would set it to TRUE in `dbWriteTable`.  What do you think?

meztez:
@detule Totally make senses. The issue is only the found part of the dbWriteTable method. In any other use case, your response is on point. Not sure the best way to do the found check any faster.

hadley:
@detule hmmmm, my gut feeling would have been that it only searches the current catalog/schema. We should try and figure out the underlying principle here.

detule:
Hey @hadley - copy.  Though as currently implemented it searches everywhere.  I wonder then, if there is some ambiguity in terms of how to interpret it, if it makes sense to side with "current precedent" so as to not break any code folks might have out there in the wild.

This is further, I think, muddled by some anecdotal evidence - for example with SQL Server when you create a temporary table it is *not* in your current catalog.  Rather, it is housed in `tempdb.dbo`.  So - if trying to write to a temp table and only checking for existence in the current catalog, you might never find it.

With this said, I wonder if it makes sense to merge the optimization in this PR since according to #759 it goes a way towards improving performance.  Happy to tackle `dbExistsTable` as a follow up.

detule:
@hadley

Let me know if I missed the mark on this one.

It allows us to, down the line and if needed, also change the default for `dbExistsTableForWrite` to something like what you suggested - checking for output of `SELECT *`.

R/driver-snowflake.R
```diff
@@ -5,6 +5,60 @@ NULL
...
+    }
+    callNextMethod(conn = conn, name = name, ..., catalog_name = catalog_name,
+      schema_name = schema_name, column_name = column_name, exact = exact)
+  }
+)
+
+#' If the catalog, or the schema arguments are NULL, attempt to infer
+#' by querying for CURRENT_DATABASE() and CURRENT_SCHEMA().  We do this
+#' to aid with performance, as the SQLTables method is more performant
+#' when restricted to a particular DB/schema.
```
simonpcouch, line 43: In `?Snowflake` currently, this description is repeated twice, and this block is interpreted as the page title. I would remove the above 4 lines

R/driver-snowflake.R
```diff
@@ -5,6 +5,60 @@ NULL
 #' @rdname DBI-classes
 setClass("Snowflake", contains = "OdbcConnection")
 
+#' @description
```
simonpcouch, line 8:
```suggestion
#' Connecting to Snowflake via ODBC
#' 
#' @description
```

R/odbc-connection.R
```diff
@@ -460,3 +460,56 @@ setMethod("odbcConnectionTableTypes", "OdbcConnection",
...
+# to optimize this search.
+setGeneric(
+  "dbExistsTableForWrite",
+  valueClass = "logical",
+  function(conn, name, ...) {
+    standardGeneric("dbExistsTableForWrite")
+  }
+)
+
+#' @rdname OdbcConnection
```
simonpcouch, line 488: Since this is our own generic and this help-file is described as

> Implementations of pure virtual functions defined in the DBI package for OdbcConnection objects.

I might just `@noRd` these methods.

R/driver-snowflake.R
```diff
@@ -5,6 +5,60 @@ NULL
...
+           name,
+           ...,
+           catalog_name = NULL,
+           schema_name = NULL,
+           column_name = NULL,
+           exact = FALSE) {
+    if(is.null(catalog_name) || is.null(schema_name)) {
+      res <- dbGetQuery(conn, "SELECT CURRENT_DATABASE() AS CAT, CURRENT_SCHEMA() AS SCH")
+      if(is.null(catalog_name) && !is.na(res$CAT[1])) {
+        catalog_name <- res$CAT[1]
```
simonpcouch, line 29: From what I can tell in docs and experimenting locally, `CAT` and `SCH` will always be length one. Any reason why we're doing this `[1]`?

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # odbc (development version)
 
+* Snowflake: Improved performance on write (#760).
```
simonpcouch, line 3: By quite a bit!🎉

R/driver-snowflake.R
detule: Done - thanks!

NEWS.md
detule: Thanks Simon!

R/driver-snowflake.R
```diff
@@ -5,6 +5,58 @@ NULL
...
+      if(is.null(catalog_name) && !is.na(res$CAT)) {
+        catalog_name <- res$CAT
+      }
+      if(is.null(schema_name) && !is.na(res$SCH)) {
+        schema_name <- res$SCH
+      }
+    }
+    callNextMethod(conn = conn, name = name, ..., catalog_name = catalog_name,
+      schema_name = schema_name, column_name = column_name, exact = exact)
+  }
```
hadley, line 39: I think you only need to supply the arguments that have changed?

```suggestion
    callNextMethod(catalog_name = catalog_name, schema_name = schema_name)
  }
```

R/driver-snowflake.R
```diff
@@ -5,6 +5,58 @@ NULL
...
+setMethod("odbcConnectionColumns_", c("Snowflake", "character"),
+  function(conn,
+           name,
+           ...,
+           catalog_name = NULL,
+           schema_name = NULL,
+           column_name = NULL,
+           exact = FALSE) {
+    if(is.null(catalog_name) || is.null(schema_name)) {
+      res <- dbGetQuery(conn, "SELECT CURRENT_DATABASE() AS CAT, CURRENT_SCHEMA() AS SCH")
```
hadley, line 29: I think it's probably worth pulling this out into a function that returns a list where the catalog and schema are `NULL` when not set/found. I think that will make the rest of the code a bit easier to understand/reason about, and then will make it straightforward to extend if we decide that's the function that should become generic.

R/driver-snowflake.R
```diff
@@ -5,6 +5,58 @@ NULL
...
+      }
+    }
+    callNextMethod(conn = conn, name = name, ..., catalog_name = catalog_name,
+      schema_name = schema_name, column_name = column_name, exact = exact)
+  }
+)
+
+#' @rdname Snowflake
+setMethod("dbExistsTableForWrite", c("Snowflake", "character"),
+  function(conn, name, ...) {
```
hadley, line 44: I think schema and catalog should be the explicit arguments here since you're modifying them, and I'm pretty sure that should allow you to drop the `do.call`.

R/odbc-connection.R
```diff
@@ -460,3 +460,56 @@ setMethod("odbcConnectionTableTypes", "OdbcConnection",
...
+setGeneric(
+  "dbExistsTableForWrite",
+  valueClass = "logical",
+  function(conn, name, ...) {
+    standardGeneric("dbExistsTableForWrite")
+  }
+)
+
+#' @noRd
+#' @inheritParams DBI::dbExistsTable
```
hadley, line 489: Since these two roxygen tags just cancel each other out, I think you could just delete them 😄

hadley (approved):
With regards to the `SELECT *`, I was thinking of that more of a general principle to guide whether or not a table exists, rather than an implementation idea.

It really feels like, particularly for snowflake and databricks, that assuming any catalog/schema makes everything really slow, and I'm not convinced that that makes sense as a default, since it doesn't correspond to the mental model you will have built up writing SQL queries.  Does that make sense?

R/driver-snowflake.R
detule: Hey Hadley -  good thought, but I don't think this works ( or i screwed something up when trying ).

R/driver-snowflake.R
detule: Good idea.  Done.
Thanks

R/driver-snowflake.R
detule: Hey - that's also a good thought!

But unless you feel strongly about this, I would err on the side of leaving the signature of `dbExistsTableForWrite` the same as `dbExistsTable` - at least for now, with the current default.

detule:
> With regards to the `SELECT *`, I was thinking of that more of a general principle to guide whether or not a table exists, rather than an implementation idea.
> 
> It really feels like, particularly for snowflake and databricks, that assuming any catalog/schema makes everything really slow, and I'm not convinced that that makes sense as a default, since it doesn't correspond to the mental model you will have built up writing SQL queries. Does that make sense?

I think I am with you, and in particular when writing.  We should assume that the identifier passed is within the scope of the `CURRENT` catalog/schema, or a temporary table.  There is some nuance with temporary tables.  I think we are safe with what we have done here with `Snowflake` since temporary tables are always created in the current catalog/schema.  But for other back-ends this may not be the case.

I plan on following up for `databricks` - i am not even sure if we are creating temporary tables correctly there when using `dbWriteTable( ..., temporary = TRUE )`.

R/odbc-connection.R
detule: Good call. Done.

Thanks

hadley:
Ah yeah, I keep forgetting about temporary tables, and how they live in different places in different databases. I think that's a nice place where the `SELECT *` idea works (since presumably if you create `CREATE TEMPORARY TABLE foo` then `SELECT * FROM foo` will work), but it may be hard to translate that idea into code that uses the odbc interface.

R/driver-snowflake.R
hadley: I think we can leave the signature of the generic the same, but it's ok for the signature of method to be different?

R/driver-snowflake.R
hadley: Let me take a look.

hadley: Oh yeah, you're correct. I had misunderstood the docs for `callNextMethod()`:

``` r
setClass("A")

1. setClass("B", contains = "A")

setGeneric("foo", function(x, a, b) standardGeneric("foo"))
#> [1] "foo"
setMethod("foo", "A", function(x, a, b) {
  c(a = a, b = b)
})
setMethod("foo", "B", function(x, a, b) {
  callNextMethod()
})
foo(new("B"), a = 1, b = 1)
#> a b 
#> 1 1

setMethod("foo", "B", function(x, a, b) {
  callNextMethod(b = 2)
})
foo(new("B"), a = 1, b = 1)
#> Error in .nextMethod(b = 2): argument "a" is missing, with no default
```

<sup>Created on 2024-03-05 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

R/driver-snowflake.R
detule: Done - let me know if that's not what you had in mind.

R/driver-snowflake.R
```diff
@@ -5,6 +5,66 @@ NULL
...
+    catSchema <- getCatalogSchema(conn, catalog_name, schema_name)
+
+    callNextMethod(conn = conn, name = name, ..., catalog_name = catSchema$catalog_name,
+      schema_name = catSchema$schema_name, column_name = column_name, exact = exact)
+  }
+)
+
+#' @rdname Snowflake
+setMethod("dbExistsTableForWrite", c("Snowflake", "character"),
+  function(conn, name, ...,
```
hadley, line 57: Yeah, this looks good!

hadley (approved):

PR #773, opened 2024-03-07
include synonyms in `odbcListObjects()` output

### LINKED ISSUE #221: Have synonyms show-up for Microsoft SQL Server `dbListTables`
*@brshallo commented on May 7, 2018, 5:21 PM UTC:*

I regularly connect to data marts provisioned for me on Microsoft SQL Server.

When connecting to Microsoft SQL Server, the RStudio "Connections" tab shows "tables" and "views" but does not show "synonyms". ("Synonyms" also do not show-up when running `DBI::dbListTables(con)`)

If I want to view the synonyms in R I will typically do something like:

```
library(tidyverse)
xtypes <- c("V", "SN", "U")
tables_access <- tbl(con, "sysobjects") %>% 
  select(name, xtype) %>% 
  collect() %>%
  filter(str_detect(xtype, "V|SN|U"))

View(tables_access)

```

Though this is less convenient. The ability to connect and query synonyms is unaffected. Only the ability to preview tables via `DBI::dbListTables` / the Connections tab, and only for Microsoft SQL Server.

Right now I end-up usually using Tableau's data previewer in parallel (which I believe is using the same driver but defaults to show synonyms as well as views and tables)...

If it does not make sense to have synonyms show-up by default in the package, do you know what change to options i can make that would make 'synonyms' show-up in my 'Connections' pane in RStudio when connecting to Databases on Microsoft SQL Server?

### System details

```
RStudio Edition : Desktop
RStudio Version :  1.1.282
OS Version      :  Windows 10 Enterprise, x64-based processor
R Version       :  R version 3.4.2 (2017-09-28)

```

### Packages:

```
dbi pacakge version: 0.8
driver: "SQL Server"

```

The same thing occurs whether I'm connecting via windows authentication or with a specific username and password with odbc. The lack of synonyms showing-up when running `DBI::dbListTables` also occurs whether I'm using odbc or RJDBC for connecting.

*This issue was moved by [krlmlr](https://github.com/krlmlr) from [r-dbi/DBI#244](https://github.com/r-dbi/DBI/issues/244).*

simonpcouch:
Closes #221.

This PR modifies/adds SQL Server methods for `dbListTables()`, `dbExistsTable()`, and `odbcListObjects()` that include synonyms in output. This also means that synonyms will show up in the Connections pane.

~~Note that this PR makes no changes to `dbListTables()`, which is noted in the original issue.~~

[EDITs: update PR scope]

### DIFF
--- a/R/driver-sql-server.R
+++ b/R/driver-sql-server.R
@@ -260,6 +260,10 @@ synonyms_query <- function(conn, catalog_name = NULL, schema_name = NULL) {
   has_catalog <- !is.null(catalog_name)
   has_schema <- !is.null(schema_name)
 
+  if (!has_catalog & !has_schema) {
+    return(paste0(res, " WHERE ", filter_is_table))
+  }
+
   if (has_catalog & has_schema) {
     res <- paste0(res, " WHERE DB.name = '", catalog_name, "' AND Sch.name = '", schema_name, "'")
   } else if (has_catalog) {
@@ -268,5 +272,7 @@ synonyms_query <- function(conn, catalog_name = NULL, schema_name = NULL) {
     res <- paste0(res, " WHERE Sch.name = '", schema_name, "'")
   }
 
-  paste0(res, " AND OBJECTPROPERTY(Object_ID(Syn.base_object_name), 'IsTable') = 1;")
+  paste0(res, " AND ", filter_is_table)
 }
+
+filter_is_table <- "OBJECTPROPERTY(Object_ID(Syn.base_object_name), 'IsTable') = 1;"
--- a/R/driver-sql-server.R
+++ b/R/driver-sql-server.R
@@ -76,8 +76,12 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
       return(!is.na(dbGetQuery(conn, query)[[1]]))
     }
     df <- odbcConnectionTables(conn, name = name, ...)
+    if (NROW(df) > 0) {
+      return(TRUE)
+    }
+
     synonyms <- dbGetQuery(conn, synonyms_query(conn, ...))
-    NROW(df) > 0 || name %in% synonyms$name
+    name %in% synonyms$name
   }
 )
 
@@ -255,9 +259,6 @@ synonyms_query <- function(conn, catalog_name = NULL, schema_name = NULL) {
 
   has_catalog <- !is.null(catalog_name)
   has_schema <- !is.null(schema_name)
-  if (!has_catalog & !has_schema) {
-    return(paste0(res, filter_is_table))
-  }
 
   if (has_catalog & has_schema) {
     res <- paste0(res, " WHERE DB.name = '", catalog_name, "' AND Sch.name = '", schema_name, "'")
@@ -267,7 +268,5 @@ synonyms_query <- function(conn, catalog_name = NULL, schema_name = NULL) {
     res <- paste0(res, " WHERE Sch.name = '", schema_name, "'")
   }
 
-  paste0(res, filter_is_table)
+  paste0(res, " AND OBJECTPROPERTY(Object_ID(Syn.base_object_name), 'IsTable') = 1;")
 }
-
-filter_is_table <- " AND OBJECTPROPERTY(Object_ID(Syn.base_object_name), 'IsTable') = 1;"
--- a/R/driver-sql-server.R
+++ b/R/driver-sql-server.R
@@ -73,11 +73,11 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
     stopifnot(length(name) == 1)
     if (isTempTable(conn, name, ...)) {
       query <- paste0("SELECT OBJECT_ID('tempdb..", name, "')")
-      !is.na(dbGetQuery(conn, query)[[1]])
-    } else {
-      df <- odbcConnectionTables(conn, name = name, ...)
-      NROW(df) > 0
+      return(!is.na(dbGetQuery(conn, query)[[1]]))
     }
+    df <- odbcConnectionTables(conn, name = name, ...)
+    synonyms <- dbGetQuery(conn, synonyms_query(conn, ...))
+    NROW(df) > 0 || name %in% synonyms$name
   }
 )
 
--- a/R/driver-sql-server.R
+++ b/R/driver-sql-server.R
@@ -204,3 +204,56 @@ setMethod("odbcDataType", "Microsoft SQL Server",
     )
   }
 )
+
+#' @rdname SQLServer
+#' @description
+#' ## `odbcListObjects()`
+#'
+#' This method makes tables that are synonyms visible in the Connections pane.
+# See (#221).
+#' @usage NULL
+#' @export
+`odbcListObjects.Microsoft SQL Server` <- function(connection,
+                                                   catalog = NULL,
+                                                   schema = NULL,
+                                                   name = NULL,
+                                                   type = NULL,
+                                                   ...) {
+  objects <- NextMethod(
+    object = connection,
+    catalog = catalog,
+    schema = schema,
+    name = name,
+    type = type,
+    ...
+  )
+
+  if (is.null(catalog) | is.null(schema)) {
+    return(objects)
+  }
+
+  synonyms <- dbGetQuery(
+    connection,
+    "SELECT
+       catalog = DB.name,
+       [schema] = Sch.name,
+       name   = Syn.name
+     FROM sys.synonyms          AS Syn
+       INNER JOIN sys.schemas AS Sch
+         ON Sch.schema_id = Syn.schema_id
+       INNER JOIN sys.databases AS DB
+         ON Sch.principal_id = DB.database_id
+     WHERE DB.name = ? AND Sch.name = ?
+       AND OBJECTPROPERTY(Object_ID(Syn.base_object_name), 'IsTable') = 1;",
+    params = list(catalog, schema)
+  )
+
+  if (!is.null(name)) {
+    synonyms <- synonyms[synonyms$name == name, , drop = FALSE]
+  }
+
+  rbind(
+    objects,
+    data.frame(name = synonyms$name, type = rep("table", length(synonyms$name)))
+  )
+}
--- a/R/driver-sql-server.R
+++ b/R/driver-sql-server.R
@@ -234,6 +234,10 @@ setMethod("odbcDataType", "Microsoft SQL Server",
     ...
   )
 
+  if (!any(objects$type == "table")) {
+    return(objects)
+  }
+
   synonyms <- dbGetQuery(connection, synonyms_query(connection, catalog, schema))
 
   if (!is.null(name)) {
--- a/R/driver-sql-server.R
+++ b/R/driver-sql-server.R
@@ -234,7 +234,7 @@ setMethod("odbcDataType", "Microsoft SQL Server",
     ...
   )
 
-  if (!any(objects$type == "table")) {
+  if (!any(objects$type == "table") && nrow(objects) > 0) {
     return(objects)
   }
 
--- a/R/driver-sql-server.R
+++ b/R/driver-sql-server.R
@@ -107,15 +107,7 @@ setMethod("dbListTables", "Microsoft SQL Server",
       res <- c(res, res_temp)
     }
 
-    if (is.null(catalog_name) | is.null(schema_name)) {
-      return(res)
-    }
-
-    synonyms <- dbGetQuery(
-      conn,
-      synonyms_query,
-      params = list(catalog_name, schema_name)
-    )
+    synonyms <- dbGetQuery(conn, synonyms_query(conn, catalog_name, schema_name))
 
     c(res, synonyms$name)
   }
@@ -238,15 +230,7 @@ setMethod("odbcDataType", "Microsoft SQL Server",
     ...
   )
 
-  if (is.null(catalog) | is.null(schema)) {
-    return(objects)
-  }
-
-  synonyms <- dbGetQuery(
-    connection,
-    synonyms_query,
-    params = list(catalog, schema)
-  )
+  synonyms <- dbGetQuery(connection, synonyms_query(connection, catalog, schema))
 
   if (!is.null(name)) {
     synonyms <- synonyms[synonyms$name == name, , drop = FALSE]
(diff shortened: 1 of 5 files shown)


R/driver-sql-server.R
```diff
@@ -204,3 +204,56 @@ setMethod("odbcDataType", "Microsoft SQL Server",
...
+       catalog = DB.name,
+       [schema] = Sch.name,
+       name   = Syn.name
+     FROM sys.synonyms          AS Syn
+       INNER JOIN sys.schemas AS Sch
+         ON Sch.schema_id = Syn.schema_id
+       INNER JOIN sys.databases AS DB
+         ON Sch.principal_id = DB.database_id
+     WHERE DB.name = ? AND Sch.name = ?
+       AND OBJECTPROPERTY(Object_ID(Syn.base_object_name), 'IsTable') = 1;",
```
simonpcouch, line 247: For now, I've opted to only include synonyms whose base object is a table, and refer to that synonym as a "table" in the output. This means that synonyms (that are tables) will be previewable through the Connections pane, but also feels a bit hacky. From what I can tell, the drop-down based in the pane based on `odbcListFields()` also doesn't work (but doesn't cause errors/crashes).

Instead, we could opt not to include this table and return the `type` as `synonym`. This means that there would be entries in the Connections pane but no previews.

ThomasSoeiro:
@simonpcouch 
Hi,
I have the same issue with an Oracle database. Should I open a new issue or can it be taken care of here?
Thanks!

simonpcouch:
@ThomasSoeiro A separate issue is good! :)

hadley:
Hmmm, I think I'd be inclined to include these in `dbListTables()`, using our principle that if you can do `SELECT foo` then `dbExistsTable(con, "foo")` should be true, and `"foo"` should be included in the output of `dbListTables()`.

simonpcouch:
Synonyms are now supported with `dbListTables()`. Still need to figure out how to go about `dbExistsTable()`. 

<details>
  <summary>With `main`</summary>
  
  ``` r
library(DBI)
library(odbc)
con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA",
                 pwd = Sys.getenv("sqlServerPass"))

dbExecute(con, "create schema odbc")
#> [1] 0
dbExecute(con, "create table odbc.test (x int)")
#> [1] 0

# confirm that we can find the table:
odbcListObjects(con, catalog = "master", schema = "odbc")
#>   name  type
#> 1 test table
dbListTables(con, catalog = "master", schema = "odbc")
#> [1] "test"

# make a synonym and show that it can't be found:
dbExecute(con, "create synonym odbc.test2 for odbc.test")
#> [1] 0
odbcListObjects(con, catalog = "master", schema = "odbc")
#>   name  type
#> 1 test table
dbListTables(con, catalog = "master", schema = "odbc")
#> [1] "test"
```
</details>

<details>
  <summary>With this PR</summary>
  
With this PR:

``` r
library(DBI)
library(odbc)
con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA",
                 pwd = Sys.getenv("sqlServerPass"))

dbExecute(con, "create schema odbc")
#> [1] 0
dbExecute(con, "create table odbc.test (x int)")
#> [1] 0

# confirm that we can find the table:
odbcListObjects(con, catalog = "master", schema = "odbc")
#>   name  type
#> 1 test table
dbListTables(con, catalog = "master", schema = "odbc")
#> [1] "test"

# make a synonym and show that it CAN be found:
dbExecute(con, "create synonym odbc.test2 for odbc.test")
#> [1] 0
odbcListObjects(con, catalog = "master", schema = "odbc")
#>    name  type
#> 1  test table
#> 2 test2 table
dbListTables(con, catalog = "master", schema = "odbc")
#> [1] "test"  "test2"
```
</details>

R/driver-sql-server.R
```diff
@@ -73,11 +73,11 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
...
       query <- paste0("SELECT OBJECT_ID('tempdb..", name, "')")
-      !is.na(dbGetQuery(conn, query)[[1]])
-    } else {
-      df <- odbcConnectionTables(conn, name = name, ...)
-      NROW(df) > 0
+      return(!is.na(dbGetQuery(conn, query)[[1]]))
     }
+    df <- odbcConnectionTables(conn, name = name, ...)
+    synonyms <- dbGetQuery(conn, synonyms_query(conn, ...))
+    NROW(df) > 0 || name %in% synonyms$name
```
hadley, line 80:
```suggestion
    df <- odbcConnectionTables(conn, name = name, ...)
    if (NROW(df) > 0) {
       return(TRUE)
    )
    
    synonyms <- dbGetQuery(conn, synonyms_query(conn, ...))
    name %in% synonyms$name
```

Maybe?

R/driver-sql-server.R
```diff
@@ -73,11 +73,11 @@ setMethod("dbExistsTable", c("Microsoft SQL Server", "character"),
     stopifnot(length(name) == 1)
     if (isTempTable(conn, name, ...)) {
       query <- paste0("SELECT OBJECT_ID('tempdb..", name, "')")
```
hadley, line 75: I wonder if it's worth checking to see if this works for synonyms too? (without the `tempdb..`). If so, that would substantially simplify this function, and maybe make it simpler?

R/driver-sql-server.R
```diff
@@ -204,3 +206,68 @@ setMethod("odbcDataType", "Microsoft SQL Server",
...
+            INNER JOIN sys.schemas AS Sch
+              ON Sch.schema_id = Syn.schema_id
+            INNER JOIN sys.databases AS DB
+              ON Sch.principal_id = DB.database_id"
+
+  has_catalog <- !is.null(catalog_name)
+  has_schema <- !is.null(schema_name)
+  if (!has_catalog & !has_schema) {
+    return(paste0(res, filter_is_table))
+  }
```
hadley, line 260:
```suggestion
```

I think you can just remove this?

R/driver-sql-server.R
```diff
@@ -204,3 +206,68 @@ setMethod("odbcDataType", "Microsoft SQL Server",
...
+
+  if (has_catalog & has_schema) {
+    res <- paste0(res, " WHERE DB.name = '", catalog_name, "' AND Sch.name = '", schema_name, "'")
+  } else if (has_catalog) {
+    res <- paste0(res, " WHERE DB.name = '", catalog_name, "'")
+  } else if (has_schema) {
+    res <- paste0(res, " WHERE Sch.name = '", schema_name, "'")
+  }
+
+  paste0(res, filter_is_table)
```
hadley, line 270: And then you could inline `filter_is_table` here.,

R/driver-sql-server.R
hadley: @simonpcouch Did you see this comment?

simonpcouch: I did! Just haven't made a moment to come back to this PR to address this comment and run some quick benchmarks to see how it affects performance. Will re-request review then, likely tomorrow!

hadley: Oh sorry, I though you had re-requested a review, but I must've been misreading my notifications.

simonpcouch: Was almost able to make this work with:

``` r
    dots <- list(...)
    name_loc <- paste0(
      c(dots[["catalog_name"]], dots[["schema_name"]], name),
      collapse = "."
    )
    return(!is.na(dbGetQuery(conn, paste0("SELECT OBJECT_ID('", name_loc, "')"))[[1]]))
```

The only issue is that `OBJECT_ID()` seems to only be able to work with fully qualified names. i.e. 

``` r
library(DBI)
library(odbc)
con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA",
                 pwd = Sys.getenv("sqlServerPass"))

dbExecute(con, "create schema odbc")
#> [1] 0
dbExecute(con, "create table odbc.test (x int)")
#> [1] 0
dbExecute(con, "create synonym odbc.test2 for odbc.test")
#> [1] 0

# when fully qualified, works fine:
dbExistsTable(con, SQL("master.odbc.test2"))
#> [1] TRUE

# but not without schema/catalog specified
"test2" %in% dbListTables(con)
#> [1] TRUE
dbExistsTable(con, "test2")
#> [1] FALSE
```

<sup>Created on 2024-04-10 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Specifically, this triggered [this DBItest test](https://github.com/r-dbi/DBItest/blob/adc3f7f69de7addb96554f40ea581d0ac7c71279/R/spec-sql-exists-table.R#L92).

R/driver-sql-server.R
```diff
@@ -204,3 +210,63 @@ setMethod("odbcDataType", "Microsoft SQL Server",
...
+    objects,
+    data.frame(name = synonyms$name, type = rep("table", length(synonyms$name)))
+  )
+}
+
+synonyms_query <- function(conn, catalog_name = NULL, schema_name = NULL) {
+  res <- "SELECT
+            catalog = DB.name,
+            [schema] = Sch.name,
+            name   = Syn.name
```
hadley, line 253: You might be able to simplify this query by using the `DB_ID()` and `SCHEMA_ID()` functions to convert schema/database names directly to IDs.

R/driver-sql-server.R
hadley: Hmmmm, is `dbListTables()` correct here? i.e. `SELECT * FROM test2` won't work (IIUC) because it's not in the current schema. So should it actually be listed?

simonpcouch:
Here are those benchmarks with the PR as-is:

Elapsed times to the 3 affected functions with 0 to 10000 synonyms present. One column of plots gives CRAN timings and the other giving dev, showing a drastic increase in timings for the dev package.

..where the values to the left of `x = 1` are `x = 0` (with a small shift for `log(0)`). Timings for `dbExistsTable()` are for a table that doesn't exist, so the `synonyms_query` kicks in. Generally, doesn't look great🏄

Will look into your most recent comments.


<details>
  <summary>Source</summary>

 ```r
  library(tidyverse)
  
  # run the following, once with cran odbc and once with dev ------------------------------
  library(DBI)
  library(odbc)
  con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA",
                   pwd = Sys.getenv("sqlServerPass"))
  
  dbExecute(con, "create schema odbc")
  dbExecute(con, "create table odbc.test (x int)")
  
  n_synonyms <- c(0, 10^c(0:4))
  timings <- 
    data.frame(
      n = numeric(0), 
      odbcList = numeric(0), 
      dbList = numeric(0), 
      dbExists = numeric(0)
    )
  
  for (n_synonym in n_synonyms) {
    for (n in seq_len(n_synonym)) {
      dbExecute(
        con, 
        SQL(paste0("create synonym odbc.test", n, " for odbc.test"))
      )
    }
    
    for (i in 1:3) {
      timings <- rbind(
        timings,
        data.frame(
          n = n_synonym, 
          odbcList = system.time(odbcListObjects(con))[["elapsed"]], 
          dbList = system.time(dbListTables(con))[["elapsed"]], 
          dbExists = system.time(dbExistsTable(con, "test0"))[["elapsed"]],
          version = ifelse(packageVersion("odbc") == "1.4.2", "cran", "dev")
        )
      )
    }

    for (n in seq_len(n_synonym)) {
      dbExecute(
        con, 
        SQL(paste0("drop synonym odbc.test", n))
      )
    }
  }
  
  # write to the respective .csv
  # write_csv(timings, file = "dev_timings.csv")
  # write_csv(timings, file = "cran_timings.csv")


  # then, reading in and plotting --------------------------------------------------------------
  dev_timings <- read_csv("dev_timings.csv")
  cran_timings <- read_csv("cran_timings.csv")
  
  
  bind_rows(dev_timings, cran_timings) %>%
    pivot_longer(cols = c(odbcList, dbList, dbExists), names_to = "fn") %>%
    mutate(n = n + .1) %>%
    ggplot() +
    geom_point(aes(x = n, y = value)) +
    facet_grid(rows = vars(fn), cols = vars(version)) +
    scale_x_log10() +
    labs(
      x = "Number of Synonyms", 
      y = "Time",
      caption = "Only 'default' objects + 1 0-row table, 3 calls per (n, fn) pair"
    )
```
</details>

File: R/driver-sql-server.R (hunk 3 above)
Comment (simonpcouch) [reply R/driver-sql-server.R:75]: :thinkies: The DBI docs for the generic read:

> dbListTables() returns a character vector that enumerates all tables and views in the database.

so... I think it's correct that `dbListTables()` would list `test2`. 

---

Some statements:

1) DBI docs: _dbListTables() returns a character vector that enumerates all tables and views in the database._
2) DBItest, linked above: _For all tables listed by `dbListTables()`, `dbExistsTable()` should return `TRUE`._
3) Us, on slack: _maybe the principle is that if `SELECT * FROM foo` works  then `dbExistsTable(con, "foo")` should return TRUE?_

So `dbListTables()` and `dbExistsTable()` should agree, and `dbListTables()` seems to be doing the right thing. The connection to `SELECT * FROM foo` doesn't seem to hold up, at least for SQL Server.

FWIW, with this PR:

``` r
library(DBI)
library(odbc)
con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA",
                 pwd = Sys.getenv("sqlServerPass"))


"test" %in% dbListTables(con)
#> [1] TRUE
dbExistsTable(con, "test")
#> [1] TRUE
dbGetQuery(con, "SELECT * FROM test")
#> Error in eval(expr, envir, enclos): nanodbc/nanodbc.cpp:1711: 00000
#> [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Invalid object name 'test'.

"test2" %in% dbListTables(con)
#> [1] TRUE
dbExistsTable(con, "test2")
#> [1] TRUE
dbGetQuery(con, "SELECT * FROM test2")
#> Error in eval(expr, envir, enclos): nanodbc/nanodbc.cpp:1711: 00000
#> [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Invalid object name 'test2'.
```

<sup>Created on 2024-04-10 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

With CRAN, both `"test2"` calls are `FALSE`.

R/driver-sql-server.R
simonpcouch: Ah, went ahead and removed this  at the time but just ran into why it was there. If `!has_catalog & !has_schema` then the filter has to start with `WHERE` rather than `AND`. I'll go ahead and add a test there so it's more obvious in the future. da86cc41.

simonpcouch:
Updated benchmarks:

Screenshot 2024-04-11 at 2 00 03 PM

Definitely better than before but not negligible.

R/driver-sql-server.R
simonpcouch: Was able to rework using `SCHEMA_NAME()` and respecting `synonyms(catalog_name)`, so no more joins and a good bit simpler now.

R/driver-sql-server.R
hadley: I'm pretty sure that the DBI docs are underspecified here — I think they really should read a "character vector that enumerates all tables and views in the database _for the active schema/catalog_.". 

So maybe we're blowing out the scope of this PR, but I think we should narrow down this behaviour because I think that we should be returning `FALSE` for the `test` calls too.

simonpcouch:
Let's table this PR until at least after the upcoming release. Will mark as draft for now.

Issue #714, opened 2024-01-01
performance - actions taken on establishing a connection


detule:
Currently when determining the structure of the object tree for the connection, [we call](https://github.com/r-dbi/odbc/blob/main/R/connection-pane.R#L34) odbcConnectionCatalogs/Schemas just to verify if there is a non-trivial result.  For example, `if length(odbcConnectionSchemas(conn)) then <include schemas in object tree>`.

We have seen instances where these catalog calls can be expensive.  At the same, this information should be available by interrogating the`SQL_CATALOG_NAME` and `SQL_SCHEMA_USAGE` info types in the `SQLGetInfo` API call.  Driver should be able to satisfy these calls without a round-trip to data source.

Issue #606, opened 2023-09-21
Performance issue to connect to an Oracle database using Oracle's OEM driver


ThomasSoeiro:
### Issue Description and Expected Result

When using `package:odbc` with Oracle's OEM driver, the connection is excessively long to establish. Once established, the Connection Pane seems to load objects forever and the session is excessively slow in the meantime:

![image](https://github.com/r-dbi/odbc/assets/124079058/c45cd745-e84b-4611-b12e-f1e60ce33aae)

When `dbConnect()` is wrapped in e.g. `invisible()` or `system.time()`, the connection establishes much quicker and the session is not slow. However, the Connection Pane is not populated by the content of the database.

This was previously discussed on rstudio/rstudio#12699.

### Database

```r
dbGetInfo(con)
# $dbname
# [1] ""
# 
# $dbms.name
# [1] "Oracle"
# 
# $db.version
# [1] "19.00.0000"
# 
# $username
# [1] ""
# 
# $host
# [1] ""
# 
# $port
# [1] ""
# 
# $sourcename
# [1] "OracleODBC-12g"
# 
# $servername
# [1] "IPIAMPR2.WORLD"
# 
# $drivername
# [1] "SQORA32.DLL"
# 
# $odbc.version
# [1] "03.52"
# 
# $driver.version
# [1] "12.01.0020"
# 
# $odbcdriver.version
# [1] "03.52"
# 
# $supports.transactions
# [1] TRUE
# 
# $getdata.extensions.any_column
# [1] TRUE
# 
# $getdata.extensions.any_order
# [1] TRUE
# 
# attr(,"class")
# [1] "Oracle"      "driver_info" "list" 
```

### Reproducible Example

This is slow:
```r
library(DBI)

start_time <- Sys.time()
con <- dbConnect(drv = odbc::odbc(), dsn = "<DSN using Oracle's OEM driver>")
Sys.time() - start_time
# Time difference of 33.26858 mins
```

A workaround:
```r
library(DBI)

# or invisible() instead of system.time()
system.time(
  con <- dbConnect(drv = odbc::odbc(), dsn = "<DSN using Oracle's OEM driver>")
)
# utilisateur     système      écoulé 
#       0.073       0.033      20.351 
```

<details>
<summary>Session Info</summary>

```r
sessionInfo()
# R version 4.1.2 (2021-11-01)
# Platform: x86_64-pc-linux-gnu (64-bit)
[...]

PR #777, opened 2024-03-17
connections-pane: infer hierarchy more efficienty

detule:
This should lighten the usage of the ( sometimes expensive ) `SQLTables` API call to infer the DB hierarchy when connecting.

Over time, and if/when needed, we could also make the DB hierarchy static and encode the per-back-end structure in the package.

[Documentation on the attributes](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlgetinfo-function?view=sql-server-ver16).

Closes https://github.com/r-dbi/odbc/issues/714 and hopefully fixes https://github.com/r-dbi/odbc/issues/606.

### DIFF
--- a/R/connection-pane.R
+++ b/R/connection-pane.R
@@ -44,15 +44,13 @@ odbcListObjectTypes.default <- function(connection) {
     obj_types <- c(obj_types, list(view = list(contains = "data")))
   }
 
-  # check for multiple schema or a named schema
-  schemas <- string_values(odbcConnectionSchemas(connection))
-  if (length(schemas) > 0) {
+  # check for schema support
+  if (conn@info$supports.schema) {
     obj_types <- list(schema = list(contains = obj_types))
   }
 
   # check for multiple catalogs
-  catalogs <- string_values(odbcConnectionCatalogs(connection))
-  if (length(catalogs) > 0) {
+  if (conn@info$supports.catalogs) {
     obj_types <- list(catalog = list(contains = obj_types))
   }
 
--- a/R/connection-pane.R
+++ b/R/connection-pane.R
@@ -45,7 +45,7 @@ odbcListObjectTypes.default <- function(connection) {
   }
 
   # check for schema support
-  if (conn@info$supports.schema) {
+  if (connection@info$supports.schema) {
     obj_types <- list(schema = list(contains = obj_types))
   }
 
--- a/R/connection-pane.R
+++ b/R/connection-pane.R
@@ -50,7 +50,7 @@ odbcListObjectTypes.default <- function(connection) {
   }
 
   # check for multiple catalogs
-  if (conn@info$supports.catalogs) {
+  if (connection@info$supports.catalogs) {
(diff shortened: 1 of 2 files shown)


R/connection-pane.R
```diff
@@ -44,15 +44,13 @@ odbcListObjectTypes.default <- function(connection) {
     obj_types <- c(obj_types, list(view = list(contains = "data")))
   }
 
-  # check for multiple schema or a named schema
-  schemas <- string_values(odbcConnectionSchemas(connection))
-  if (length(schemas) > 0) {
+  # check for schema support
+  if (conn@info$supports.schema) {
```
simonpcouch, line 48:
```suggestion
  if (connection@info$supports.schema) {
```

R/connection-pane.R
```diff
@@ -44,15 +44,13 @@ odbcListObjectTypes.default <- function(connection) {
...
     obj_types <- list(schema = list(contains = obj_types))
   }
 
   # check for multiple catalogs
-  catalogs <- string_values(odbcConnectionCatalogs(connection))
-  if (length(catalogs) > 0) {
+  if (conn@info$supports.catalogs) {
```
simonpcouch, line 53:
```suggestion
  if (connection@info$supports.catalogs) {
```

src/connection.cpp
```diff
@@ -76,11 +76,15 @@ std::string get_info_or_empty(connection_ptr const& p, short type) {
 // [[Rcpp::export]]
 Rcpp::List connection_info(connection_ptr const& p) {
   SQLUINTEGER getdata_ext;
+  SQLUINTEGER owner_usage;
   try {
     getdata_ext =
         (*p)->connection()->get_info<SQLUINTEGER>(SQL_GETDATA_EXTENSIONS);
+    owner_usage =
+        (*p)->connection()->get_info<SQLUINTEGER>(SQL_OWNER_USAGE);
```
simonpcouch, line 84: I see on the linked Microsoft docs that `SQL_OWNER_USAGE` was deprecated in favor of `SQL_SCHEMA_USAGE` in ODBC 3.x, released in 1995. As someone relatively new to using ODBC, a question: do we generally aim to support all ODBC versions or are we better off to use the newer keyword in anticipation of the older one being fully deprecated? From what I can tell, the deprecation cycle for ODBC is loooong, so fine by me to use the older keyword as we are here if we anticipate it will be supported for a long time.

R/connection-pane.R
detule: Thanks!

src/connection.cpp
detule: Hey Simon - I went back on forth on this one, but your comment nudged me in the direction of `SCHEMA` in favor of `OWNER`.

Thanks!

simonpcouch (approved):
Looking good!🙂

simonpcouch:
Thumbs-up from me if you wanna give these changes a shoutout in NEWS.

Issue #628, opened 2023-11-28
use rlang type checkers


simonpcouch:
`usethis::use_standalone("r-lib/rlang", "types-check")` in favor of of home-grown argument checkers.

Issue #627, opened 2023-11-28
Use cli errors


simonpcouch:

jan-swissre:
Is this related to this issue?
```
Error in "fun(..., .envir = .envir)" : 
  ! Could not parse cli `{}` expression: `6E291E8D-0D4A-464C-…`.
Caused by error in `parse(text = code, keep.source = FALSE)`:
! <text>:1:6: unexpected symbol
1: 6E291E8D
```
Do you have any idea if/why it comes, from cli?

`odbc 1.5.0`

PR #781, opened 2024-04-05
check inputs with rlang type checkers

simonpcouch:
Related to #628 but does not close.

This PR introduces type checks for functions and their methods listed in "DBI Methods" in the function reference. Opening this PR just for that subset as this is already quite a large PR and will allow us to correct patterns early on.

To-do, either in a separate PR or here after one round of review (opinions?): functions in "ODBC configuration", "Database-specific helpers".

Transitioning away from base warnings and errors (see #627) where they're related to type checks but trying not to touch them generally.

### DIFF
--- a/R/dbi-connection.R
+++ b/R/dbi-connection.R
@@ -191,10 +191,10 @@ setMethod("dbListTables", "OdbcConnection",
            table_name = NULL,
            table_type = NULL,
            ...) {
-    check_string(catalog_name, allow_na = TRUE, allow_null = TRUE)
-    check_string(schema_name, allow_na = TRUE, allow_null = TRUE)
-    check_string(table_name, allow_na = TRUE, allow_null = TRUE)
-    check_string(table_type, allow_na = TRUE, allow_null = TRUE)
+    check_string(catalog_name, allow_null = TRUE)
+    check_string(schema_name, allow_null = TRUE)
+    check_string(table_name, allow_null = TRUE)
+    check_string(table_type, allow_null = TRUE)
 
     tables <- odbcConnectionTables(
       conn,
--- a/R/dbi-connection.R
+++ b/R/dbi-connection.R
@@ -191,6 +191,11 @@ setMethod("dbListTables", "OdbcConnection",
            table_name = NULL,
            table_type = NULL,
            ...) {
+    check_string(catalog_name, allow_na = TRUE, allow_null = TRUE)
+    check_string(schema_name, allow_na = TRUE, allow_null = TRUE)
+    check_string(table_name, allow_na = TRUE, allow_null = TRUE)
+    check_string(table_type, allow_na = TRUE, allow_null = TRUE)
+
     tables <- odbcConnectionTables(
       conn,
       name = table_name,
--- a/R/dbi-driver.R
+++ b/R/dbi-driver.R
@@ -168,18 +168,18 @@ setMethod("dbConnect", "OdbcDriver",
       dbms.name = NULL,
       attributes = NULL,
       .connection_string = NULL) {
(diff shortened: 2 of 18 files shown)


R/dbi-connection.R
```diff
@@ -191,6 +191,11 @@ setMethod("dbListTables", "OdbcConnection",
            table_name = NULL,
            table_type = NULL,
            ...) {
+    check_string(catalog_name, allow_na = TRUE, allow_null = TRUE)
```
simonpcouch, line 194: Generally, w.r.t. `allow_na`, I tried to err on the side of letting things slip by that I ought not to rather than than mistakenly erroring for input that's okay.

hadley: Hmmm, I think I'd tend to trend in the opposite direction, especially for these functions where we're explicitly using `NULL` as the "default" option.

R/dbi-driver.R
```diff
@@ -168,6 +168,19 @@ setMethod("dbConnect", "OdbcDriver",
       dbms.name = NULL,
       attributes = NULL,
       .connection_string = NULL) {
+    check_string(dsn, allow_na = TRUE, allow_null = TRUE)
+    check_string(timezone, allow_na = TRUE, allow_null = TRUE)
+    check_string(timezone_out, allow_na = TRUE, allow_null = TRUE)
+    check_string(encoding, allow_na = TRUE, allow_null = TRUE)
+    check_character(bigint, allow_na = TRUE, allow_null = TRUE)
```
hadley, line 175: This can be `arg_match()`

R/dbi-table.R
```diff
@@ -40,26 +40,32 @@ odbc_write_table <- function(conn,
...
                              batch_rows = getOption("odbc.batch_rows", NA),
                              ...) {
-  # Sanity checks: begin
-  stopifnot(
-    rlang::is_scalar_logical(overwrite) && !is.na(overwrite),
-    rlang::is_scalar_logical(append) && !is.na(append),
-    rlang::is_scalar_logical(temporary) && !is.na(temporary),
-    rlang::is_null(field.types) || (rlang::is_named(field.types))
-  )
+  call <- call2("dbWriteTable")
```
hadley, line 43: Does `caller_env()` not work here?

R/utils.R
```diff
@@ -165,3 +165,25 @@ random_name <- function(prefix = "") {
   name <- paste0(sample(vals, 10, replace = TRUE), collapse = "")
   paste0(prefix, "odbc_", name)
 }
+
+# check helpers for common odbc arguments --------------------------------------
+check_row.names <- function(row.names, call = caller_env()) {
+  if (!inherits_any(row.names, c("logical", "character", "NULL")) ||
+      length(row.names) > 1) {
```
hadley, line 172: I found this a bit confusing. I'd suggesting something more like:

```R
if (is.null(row.names) || is_scalar_logical() || is_string()) {
   return()
}

cli::cli_abort(...)
```

R/utils.R
```diff
@@ -165,3 +165,25 @@ random_name <- function(prefix = "") {
...
+    cli::cli_abort(
+      "{.arg row.names} must be {.code NULL}, {.code TRUE}, {.code FALSE}, \\
+       {.code NA}, or a single string, not {.obj_type_friendly {row.names}}.",
+      call = call
+    )
+  }
+}
+
+check_field.types <- function(field.types, call = caller_env()) {
+  if (!(is_null(field.types) || (is_named(field.types)))) {
```
hadley, line 182: Is `field.types` supposed to be a character or list? Or either?

simonpcouch: Not documented as one way or another, and in historical issues / code I see both used interchangeably.🥴

PR #789, opened 2024-04-16
rethrow database errors

simonpcouch:
A follow-up to #643. At the time, parsing the message any further was agreed to be more painful than it was worth. Following up on PRs since then that refine the appearance of errors, this PR proposes rethrowing `nanodbc::database_error`s with cli formatting.

Before this PR:

``` r
library(DBI)
library(odbc)
con <- dbConnect(odbc(), dsn = "sillyhead")
#> Error: nanodbc/nanodbc.cpp:1138: 00000
#> [unixODBC][Driver Manager]Data source name not found and no default driver specified

con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA", pwd = Sys.getenv("sqlServerPass"))

dbExecute(con, "SELECT * FROM boopbopbopbeep")
#> Error in eval(expr, envir, enclos): nanodbc/nanodbc.cpp:1710: 00000
#> [Microsoft][ODBC Driver 18 for SQL Server][SQL Server]Invalid object name 'boopbopbopbeep'.
```

With this PR:

``` r
library(DBI)
library(odbc)
con <- dbConnect(odbc(), dsn = "sillyhead")
#> Error in `dbConnect()`:
#> ℹ At 'nanodbc/nanodbc.cpp:1147', error 00000 from [unixODBC][Driver
#>   Manager].
#> ✖ Data source name not found and no default driver specified

con <- dbConnect(odbc(), dsn = "MicrosoftSQLServer", uid = "SA", pwd = Sys.getenv("sqlServerPass"))

dbExecute(con, "SELECT * FROM boopbopbopbeep")
#> Error in `dbExecute()`:
#> ℹ At 'nanodbc/nanodbc.cpp:1719', error 00000 from [Microsoft][ODBC Driver
#>   18 for SQL Server][SQL Server].
#> ✖ Invalid object name 'boopbopbopbeep'.
```

<sup>Created on 2024-04-16 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

In the RStudio terminal, this looks like:

Screenshot 2024-04-16 at 3 08 30 PM

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -154,6 +154,53 @@ random_name <- function(prefix = "") {
   paste0(prefix, "odbc_", name)
 }
 
+# error handling ---------------------------------------------------------------
+rethrow_database_error <- function(msg, call = trace_back()$call[[1]]) {
+  tryCatch(
+    res <- parse_database_error(msg),
+    error = function(e) {cli::cli_abort(msg, call = call)}
+  )
+
+  cli::cli_abort(
+    c(
+      "i" = "At {.file {res$cnd_context_nanodbc}}, error {res$cnd_context_code} \\
+             from {.field {paste0(res$cnd_context_driver, collapse = '')}}.",
+      set_names(res$cnd_body, nm = c("x", rep("*", length(res$cnd_body) - 1)))
+    ),
+    call = call
+  )
+}
+
+parse_database_error <- function(msg) {
+  # Split nanodbc's context, error code, and message returned from the database
+  cnd_msg <- strsplit(msg, "\n", fixed = TRUE)[[1]]
+
+  # Parse out nanodbc/nanodbc.cpp:<line#> and 5-char error code
+  cnd_context <- strsplit(cnd_msg[1], ": ")[[1]]
+  cnd_context_nanodbc <- cnd_context[1]
+  cnd_context_code <- cnd_context[2]
+
+  # In error returned from the database, find the square-bracketed context
+  # on driver and driver manager
+  cnd_context_driver <- gregexpr("(\\[.*?\\])", cnd_msg[-1])
+  cnd_context_driver <- regmatches(cnd_msg[-1], cnd_context_driver)[[1]]
+
+  # Remove the square-bracketed context from the database error
+  cnd_body <- Reduce(
+    function(p, x) gsub(p, "", x, fixed = TRUE),
+    cnd_context_driver,
+    cnd_msg[-1],
+    right = TRUE
+  )
+
+  return(list(
+    cnd_context_nanodbc = cnd_context_nanodbc,
+    cnd_context_code = cnd_context_code,
+    cnd_context_driver = cnd_context_driver,
+    cnd_body = cnd_body
+  ))
+}
+
 # check helpers for common odbc arguments --------------------------------------
 check_row.names <- function(row.names, call = caller_env()) {
   if (is.null(row.names) || is_scalar_logical(row.names) || is_string(row.names)) {
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -43,6 +43,73 @@ test_that("getSelector", {
   expect_equal(getSelector("mykey", "%", exact = FALSE), " AND mykey LIKE '%'")
 })
 
+test_that("errors are rethrown informatively (#643, #788)", {
+  skip_if_not(has_unixodbc())
+
+  expect_snapshot(error = TRUE, dbConnect(odbc(), dsn = "does_not_exist_db"))
+
+  con <- test_con("SQLITE")
+  expect_snapshot(error = TRUE, dbExecute(con, "SELECT * FROM boopbopbopbeep"))
+})
+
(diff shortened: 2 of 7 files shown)


R/utils.R
```diff
@@ -154,6 +154,53 @@ random_name <- function(prefix = "") {
   paste0(prefix, "odbc_", name)
 }
 
+# error handling ---------------------------------------------------------------
+rethrow_database_error <- function(msg, call = trace_back()$call[[1]]) {
```
simonpcouch, line 158: The current draft of `rethrow_database_error()` tries to keep as much of the plain text of the original error message intact and in order as possible, just adding formatting, so that errors are still internet searchable. 

Another approach would just forefront simplicity/readibility: start with the vendor error message, clear out the "nanodbc/nanodbc<line#>" text (as it's probably only useful to odbc developers), and only include the error code when it's not nanodbc's 00000. A previous draft of this function did that:

```
Error in `dbConnect()`:
! Data source name not found and no default driver specified
ℹ From [unixODBC][Driver Manager].
```

hadley: I wonder if we could split the difference and put the keep the nanodbc line number, but move it to the bottom. i.e. instead of:

```
i At 'nanodbc/nanodbc.cpp:1147', error 00000 from [unixODBC][Driver Manager].
x Data source name not found and no default driver specified
```

Do

```
i Error 00000 from [unixODBC][Driver Manager].
x Data source name not found and no default driver specified
i From 'nanodbc/nanodbc.cpp:1147'
```

And I think it's probably worth being explicit that this is an ODBC error, and it's unusual to start with an info bullet:

```
! ODBC failed with error 00000 from [unixODBC][Driver Manager].
x Data source name not found and no default driver specified
i From 'nanodbc/nanodbc.cpp:1147'
```


Also that's an unusual specification for the `call` argument.

R/utils.R
```diff
@@ -154,6 +154,53 @@ random_name <- function(prefix = "") {
...
+  tryCatch(
+    res <- parse_database_error(msg),
+    error = function(e) {cli::cli_abort(msg, call = call)}
```
hadley, line 161: Either drop the `{}` or add line breaks

R/utils.R
```diff
@@ -154,6 +154,53 @@ random_name <- function(prefix = "") {
...
+    cnd_msg[-1],
+    right = TRUE
+  )
+
+  return(list(
+    cnd_context_nanodbc = cnd_context_nanodbc,
+    cnd_context_code = cnd_context_code,
+    cnd_context_driver = cnd_context_driver,
+    cnd_body = cnd_body
+  ))
```
hadley, line 201:
```suggestion
  list(
    cnd_context_nanodbc = cnd_context_nanodbc,
    cnd_context_code = cnd_context_code,
    cnd_context_driver = cnd_context_driver,
    cnd_body = cnd_body
  )
```

tests/testthat/test-utils.R
```diff
@@ -43,6 +43,73 @@ test_that("getSelector", {
...
+
+test_that("parse_database_error() works with messages from the wild", {
+  expect_snapshot(
+    error = TRUE,
+    rethrow_database_error(
+      "nanodbc/nanodbc.cpp:1135: 00000
+       [unixODBC][Driver Manager]Data source name not found and no default driver specified",
+      call = NULL
+    )
+  )
```
hadley, line 70: I'd pull the message out of the snapshot avoid some of cruft:

```suggestion
  msg <- "nanodbc/nanodbc.cpp:1135: 00000
       [unixODBC][Driver Manager]Data source name not found and no default driver specified"
  expect_snapshot(rethrow_database_error(msg, call = NULL), error = TRUE)
  )
```

hadley (commented):
I like the idea, and I think you've made a great start. The only reason I'm not accepting now is just so we can do a bit more iteration on the error message itself.

R/utils.R
simonpcouch: Totally on board. Just pushed some changes reflecting your last suggestion in [ebaf6dd](https://github.com/r-dbi/odbc/pull/789/commits/ebaf6dd9acbb94b95b37d94384392d023536c475), though am game to remove that last `i From 'nanodbc/nanodbc.cpp:1147'` line entirely if you are.

re: `call`, this is indeed wonky :/ `parent.frame()` (and thus `caller_env()`) don't seem to know how to traverse across the Rcpp code (since `rethrow_database_error()` is called from C++)—setting `call = caller_env(n)` with n of 0/1/2/3 gives no error context. 

There are certainly other ways of going about this (searching for a `call` in the environments listed by `trace_back()` or attempting to plumb a call through nanodbc.cpp)—let me know if you think it's worth spending more time there.

hadley: Errors look good to me, and I think you just need to document why you need the weird `call` spec.

hadley (approved):

simonpcouch:
Confirmed GHA failures on macOS are unrelated in the issue linked above.

PR #796, opened 2024-04-27
Feature/interruptible execution clean2

detule:
Hi All:

Please consider adding this feature - the ability to "Ctrl-C" past a long running query.  For example in SQL Server

```
> dbGetQuery(conn, "WAITFOR DELAY '00:10'", immediate = TRUE);
^C
Caught user interrupt, attempting a clean exit...
Error: nanodbc/nanodbc.cpp:1711: 00000
<SQL> 'WAITFOR DELAY '00:10''
>
```

Some notes - i feel like I am being wordy below but I am sure I still left quite a bit out:

* Design: Created a `run_interruptible` wrapper that takes two arguments: a function to be executed away from the main thread, and a cleanup function to be invoked once the signal is caught on the main thread.  Perhaps this is an overkill but I did this with an eye that maybe some day we may need to be able to interrupt other calls from the `odbc` API, not just `SQLExecute`.  We can then just re-use the wrapper.
* Design: Used `std::async` rather than `std::thread`: I personally am a little biased towards `std::async` but mostly based on semantics.  I know there's quite a bit left up to the implementation but at this point it's been around for long enough that I think there should be little in the way of surprises.  I've witnessed some passionate thread-vs-async discussions before but I think it's mostly a matter of style.  If folks feel strongly one way or another, happy to change it.
* Design: Signal handling ( away from Windows ) - used the `pthread` API to make sure the SIGINT signal is delivered to the main thread.  This is pretty idiosyncratic, but i've seen issues where depending on the implementation, a signal can end up with the wrong thread.
* Opt-out: Struggled with this for a bit.  I am not a fan of `dbConnect` accruing yet more arguments.  At the same time also not in love with introducing some sort of a secret option --- `getOption(.odbc.interruptible.exec)` --- or somesuch.  There is an `attributes` arguments to `dbConnect` but when we introduced it, I think we envisioned `ODBC`/API connection arguments.  So - hated myself a little bit for it - but ended up writing in a new argument to `dbConnect`.
* Testing: Pipelines are (hopefully) passing on both windows and linux and that's a good sign.  Beyond that, on my linux box, I tested DB2, Oracle, Terradata, Databricks and did not see any issues.  Note, the `WAITFOR` style queries are not always interruptible.  For example on `snowflake`, interrupting the equivalent `SELECT system$wait(10)` won't save you any time since the `SQLCancel` call blocks on the main thread after catching the interrupt.  However, testing an actual long running execution I saw better behavior.
  * MySQL: Saw some issues with the MariaDB odbc driver 3.1.15 on Linux.  I went back to installing the driver from their github tree as the behavior is much better with that version ( 3.1.17).  In particular, when using the canned v3.1.15 driver, executing the following will hang execution:
    * `dbSend(Get)Query(conn, NA_character_)` ( part of `DBItest::send(get)_query_non_string` )
    * `dbSendStatement(conn, NA_character_)` ( part of `DBItest::send_statement_non_string` )
    * `dbExecute(conn, NA_character_)` ( part of `DBItest::execute_non_string` )

  Note, all remaining (~3.5K) `DBItest` tests pass with both drivers just fine.  Spent quite a bit of time comparing odbc traces to no avail, and thought it's something **so** idiosyncratic ( and fixed in their newer drivers to boot ) that it should probably not hold this feature back.  Options to ignore were to use newer driver, or to add the tests to the "skip" list for `driver-mysql`.
  
Appreciate any feedback - and thanks for taking a look!

### DIFF
--- a/.github/odbc/install-mariadb-driver.sh
+++ b/.github/odbc/install-mariadb-driver.sh
@@ -1,0 +1,9 @@
+sudo apt-get install -y cmake
+cd /tmp && git clone https://github.com/MariaDB/mariadb-connector-odbc.git connector
+cd connector && git checkout 3.1.17
+mkdir build && cd build
+cmake ../ -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCONC_WITH_UNIT_TESTS=Off -DCMAKE_INSTALL_PREFIX=/usr/local -DWITH_SSL=OPENSSL
+cmake --build . --config RelWithDebInfo
+sudo make install
+sudo ln -s /usr/local/lib/mariadb/libmariadb.so.3 /usr/local/lib/
+sudo ldconfig
--- a/.github/odbc/install-mariadb-driver.sh
+++ b/.github/odbc/install-mariadb-driver.sh
@@ -1,3 +1,5 @@
+# See #796 for some reasoning behind not using the canned mariadb-odbc-driver (v3.1.15), at time of writing
+# and needing to install v3.1.17 from source
 sudo apt-get install -y cmake
 cd /tmp && git clone https://github.com/MariaDB/mariadb-connector-odbc.git connector
 cd connector && git checkout 3.1.17
--- a/R/dbi-driver.R
+++ b/R/dbi-driver.R
@@ -63,8 +63,11 @@ setMethod("show", "OdbcDriver",
 #'   name for the OdbcConnection object returned from [dbConnect()]. However, if
 #'   the driver does not return a valid value, it can be set manually with this
 #'   parameter.
-#' @param attributes An S4 object of connection attributes that are passed
+#' @param attributes A list of connection attributes that are passed
 #'   prior to the connection being established. See \link{ConnectionAttributes}.
+#' @param interruptibleExecution Logical.  If true calls to SQLExecute and
+#'   SQLExecuteDirect can be interrupted when the user sends SIGINT ( ctrl-c ).
+#'   Otherwise they block.  Defaults to TRUE.
 #' @param ... Additional ODBC keywords. These will be joined with the other
 #'   arguments to form the final connection string.
 #'
@@ -167,6 +170,7 @@ setMethod("dbConnect", "OdbcDriver",
       pwd = NULL,
       dbms.name = NULL,
       attributes = NULL,
+      interruptibleExecution = TRUE,
       .connection_string = NULL) {
     check_string(dsn, allow_null = TRUE)
     check_string(timezone, allow_null = TRUE)
@@ -196,6 +200,7 @@ setMethod("dbConnect", "OdbcDriver",
       pwd = pwd,
       dbms.name = dbms.name,
       attributes = attributes,
+      interruptibleExecution = interruptibleExecution,
       .connection_string = .connection_string
     )
 
--- a/R/dbi-driver.R
+++ b/R/dbi-driver.R
@@ -65,9 +65,9 @@ setMethod("show", "OdbcDriver",
 #'   parameter.
 #' @param attributes A list of connection attributes that are passed
 #'   prior to the connection being established. See \link{ConnectionAttributes}.
-#' @param interruptibleExecution Logical.  If true calls to SQLExecute and
+#' @param interruptibleExecution Logical.  If `TRUE` calls to SQLExecute and
 #'   SQLExecuteDirect can be interrupted when the user sends SIGINT ( ctrl-c ).
-#'   Otherwise they block.  Defaults to TRUE.
+#'   Otherwise, they block.  Defaults to `TRUE`.
 #' @param ... Additional ODBC keywords. These will be joined with the other
 #'   arguments to form the final connection string.
 #'
@@ -184,6 +184,7 @@ setMethod("dbConnect", "OdbcDriver",
     check_string(uid, allow_null = TRUE)
     check_string(pwd, allow_null = TRUE)
     check_string(dbms.name, allow_null = TRUE)
+    check_bool(interruptibleExecution)
 
     con <- OdbcConnection(
       dsn = dsn,
--- a/R/dbi-driver.R
+++ b/R/dbi-driver.R
@@ -65,9 +65,11 @@ setMethod("show", "OdbcDriver",
 #'   parameter.
 #' @param attributes A list of connection attributes that are passed
 #'   prior to the connection being established. See \link{ConnectionAttributes}.
-#' @param interruptibleExecution Logical.  If `TRUE` calls to SQLExecute and
-#'   SQLExecuteDirect can be interrupted when the user sends SIGINT ( ctrl-c ).
-#'   Otherwise, they block.  Defaults to `TRUE`.
+#' @param interruptible Logical.  If `TRUE` calls to `SQLExecute` and
+#'   `SQLExecuteDirect` can be interrupted when the user sends SIGINT ( ctrl-c ).
+#'   Otherwise, they block.  Defaults to `TRUE`.  It can be set either by
+#'   manipulating this argument, or setting the global option `odbc.interruptible`
+#'   to `FALSE`.
 #' @param ... Additional ODBC keywords. These will be joined with the other
 #'   arguments to form the final connection string.
 #'
@@ -170,7 +172,7 @@ setMethod("dbConnect", "OdbcDriver",
       pwd = NULL,
       dbms.name = NULL,
       attributes = NULL,
-      interruptibleExecution = TRUE,
+      interruptible = getOption("odbc.interruptible", TRUE),
       .connection_string = NULL) {
     check_string(dsn, allow_null = TRUE)
     check_string(timezone, allow_null = TRUE)
@@ -184,7 +186,7 @@ setMethod("dbConnect", "OdbcDriver",
     check_string(uid, allow_null = TRUE)
     check_string(pwd, allow_null = TRUE)
     check_string(dbms.name, allow_null = TRUE)
-    check_bool(interruptibleExecution)
+    check_bool(interruptible)
 
     con <- OdbcConnection(
       dsn = dsn,
@@ -201,7 +203,7 @@ setMethod("dbConnect", "OdbcDriver",
       pwd = pwd,
       dbms.name = dbms.name,
       attributes = attributes,
-      interruptibleExecution = interruptibleExecution,
+      interruptible = interruptible,
       .connection_string = .connection_string
     )
 
--- a/R/dbi-driver.R
+++ b/R/dbi-driver.R
@@ -67,9 +67,9 @@ setMethod("show", "OdbcDriver",
 #'   prior to the connection being established. See \link{ConnectionAttributes}.
 #' @param interruptible Logical.  If `TRUE` calls to `SQLExecute` and
 #'   `SQLExecuteDirect` can be interrupted when the user sends SIGINT ( ctrl-c ).
-#'   Otherwise, they block.  Defaults to `TRUE`.  It can be set either by
-#'   manipulating this argument, or setting the global option `odbc.interruptible`
-#'   to `FALSE`.
+#'   Otherwise, they block.  Defaults to `TRUE` in interactive sessions, and
+#'   `FALSE` otherwise.  It can be set explicitly either by manipulating this
+#'   argument, or by setting the global option `odbc.interruptible`.
 #' @param ... Additional ODBC keywords. These will be joined with the other
 #'   arguments to form the final connection string.
 #'
@@ -172,7 +172,7 @@ setMethod("dbConnect", "OdbcDriver",
       pwd = NULL,
       dbms.name = NULL,
       attributes = NULL,
-      interruptible = getOption("odbc.interruptible", TRUE),
+      interruptible = getOption("odbc.interruptible", interactive()),
       .connection_string = NULL) {
     check_string(dsn, allow_null = TRUE)
     check_string(timezone, allow_null = TRUE)
--- a/src/odbc_result.cpp
+++ b/src/odbc_result.cpp
@@ -8,7 +8,9 @@
 namespace odbc {
 
 using odbc::utils::run_interruptible;
-using odbc::utils::pretty_print_message;
+using odbc::utils::raise_message;
+using odbc::utils::raise_warning;
+using odbc::utils::raise_error;
 odbc_result::odbc_result(
     std::shared_ptr<odbc_connection> c, std::string sql, bool immediate)
     : c_(c),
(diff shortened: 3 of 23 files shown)


detule:
* The business of the PR is in `feature: interruptible execution`; remaining commits are auxiliary / supporting cast.
* I think `snowflake` tests are failing perhaps because I don't have access to the github credentials / secret/

src/utils.cpp
```diff
@@ -47,4 +50,54 @@ namespace utils {
...
+  {
+    std::exception_ptr eptr;
+#if !defined(_WIN32) && !defined(_WIN64)
+    sigset_t set, old_set;
+    sigemptyset(&set);
+    sigaddset(&set, SIGINT);
+    int rc = pthread_sigmask(SIG_BLOCK, &set, &old_set);
+    if ( rc != 0 )
+    {
+      Rcpp::warning("Unable to properly mask SIGINT from execution thread.");
```
simonpcouch, line 64:
```suggestion
      // Unable to properly mask SIGINT from execution thread.
      Rcpp::warning("<a more user-facing message>");
```

What does this mean for the user? I'm wondering if this would be better situated as a message, error, or just a code comment.

If this remains a warning, could we `pretty_print_message()` it?

src/odbc_result.h
```diff
@@ -24,14 +24,23 @@ class odbc_error : public Rcpp::exception {
...
+      m += "\n<SQL> '" + sql + "'";
+    }
     // #432: [R] expects UTF-8 encoded strings but both nanodbc and sql are
     // encoded in the database encoding, which may differ from UTF-8
     message = Rf_translateChar(
         output_encoder.makeSEXP(m.c_str(), m.c_str() + m.length()));
   }
   const char* what() const NANODBC_NOEXCEPT { return message.c_str(); }
 
+  void pretty_print() const {
```
simonpcouch, line 38: Thinking about the names `pretty_print()` and `pretty_print_message()`... Some bits of context from R and its cli package:

* The side effects that `rethrow_database_error()` and `print_message()` are called for are conditions rather than just printed output. "Raise" or "throw" are often the verbs applied to conditions.
* In R, "messages" are a specific kind of condition (along with warnings and errors), but `pretty_print_message()` can raise either a message or warning.
* The cli functions `cli_inform()`, `cli_warn()`, and `cli_abort()` raise messages, warnings, and errors, respectively. I know this is kind of weird.

I'm wondering if we could refactor to, in Cpp,

* `raise_message()`: calls `cli::cli_inform()` directly via the `Rcpp::Environment::namespace_env("cli"); ...` pattern below
* `raise_warning()`: calls `cli::cli_warn()` directly via the same pattern
* `raise_error()`: calls `rethrow_database_error()` (which in turn calls `cli::cli_abort()`) via the current pattern

...and then writing out the `print_message()` function defined in R.

simonpcouch (commented):
This is so great. Stoked about this one.

Leaving a "Comment" rather than "Approve" review as there are quite a few Cpp changes here that I don't feel quite qualified enough to approve. Mostly feedback on the interface to cli conditions from me.

tests/testthat/_snaps/utils.md
```diff
@@ -37,7 +37,7 @@
       Error in `dbExecute()`:
       ! ODBC failed with error 00000 from [SQLite].
       x no such table: boopbopbopbeep (1)
-      i From 'nanodbc/nanodbc.cpp:1719'.
+      i From 'nanodbc/nanodbc.cpp:1713zz'.
```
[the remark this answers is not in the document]
detule: Hopefully fixed / good catch.

R/dbi-driver.R
```diff
@@ -63,8 +63,11 @@ setMethod("show", "OdbcDriver",
 #'   name for the OdbcConnection object returned from [dbConnect()]. However, if
 #'   the driver does not return a valid value, it can be set manually with this
 #'   parameter.
-#' @param attributes An S4 object of connection attributes that are passed
+#' @param attributes A list of connection attributes that are passed
 #'   prior to the connection being established. See \link{ConnectionAttributes}.
+#' @param interruptibleExecution Logical.  If true calls to SQLExecute and
```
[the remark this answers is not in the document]
detule: Done/thanks

R/dbi-driver.R
```diff
@@ -167,6 +170,7 @@ setMethod("dbConnect", "OdbcDriver",
       pwd = NULL,
       dbms.name = NULL,
       attributes = NULL,
+      interruptibleExecution = TRUE,
       .connection_string = NULL) {
     check_string(dsn, allow_null = TRUE)
```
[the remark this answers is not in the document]
detule: Done/thanks

R/dbi-driver.R
```diff
@@ -63,8 +63,11 @@ setMethod("show", "OdbcDriver",
 #'   name for the OdbcConnection object returned from [dbConnect()]. However, if
 #'   the driver does not return a valid value, it can be set manually with this
 #'   parameter.
-#' @param attributes An S4 object of connection attributes that are passed
+#' @param attributes A list of connection attributes that are passed
 #'   prior to the connection being established. See \link{ConnectionAttributes}.
+#' @param interruptibleExecution Logical.  If true calls to SQLExecute and
+#'   SQLExecuteDirect can be interrupted when the user sends SIGINT ( ctrl-c ).
+#'   Otherwise they block.  Defaults to TRUE.
```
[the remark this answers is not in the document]
detule: Done/thanks

src/odbc_result.cpp
```diff
@@ -43,21 +45,32 @@ std::shared_ptr<nanodbc::statement> odbc_result::statement() const {
...
+          s_->execute());
       num_columns_ = r_->columns();
-    } catch (const nanodbc::database_error& e) {
-      c_->set_current_result(nullptr);
+    }
+  } catch (const nanodbc::database_error& e) {
+    c_->set_current_result(nullptr);
+    if (c_->interruptible_execution_) {
+      // Executing in a thread away from main.  Signal
+      // that we have encountered an error using ane exception.
```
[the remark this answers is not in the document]
detule: done/thanks

src/utils.h
```diff
@@ -44,5 +45,23 @@ namespace utils {
...
+  /// \param exec_fn Function executed on a separate thread.  Exceptions
+  /// are caught and re-thrown on the main thread.
+  /// \param cleanup_fn Function executed on main thread in the event a
+  /// user interrupt is caught.
+  void run_interruptible(const std::function<void()>& exec_fn, const std::function<void()>& cleanup_fn);
+
+  /// \brief Entry point form package::odbc:::print_message
+  ///
+  /// \param messge Message to be shown.
+  /// \param is_warning Should displaying this message also generate an [R] warning.
```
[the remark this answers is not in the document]
detule: done/thanks

src/odbc_result.h
detule: I like it - great suggestion.  Let me know if the changes I made as part of the 'code-review' commit are not in the direction you outlined above.

src/utils.cpp
detule: Done/thanks

src/odbc_result.h
simonpcouch: Not seeing that `code-review` commit—is it possible those changes are still local?

detule: Sorry about that @simonpcouch 
Hopefully managed to push to the right branch this time.

tests/testthat/_snaps/utils.md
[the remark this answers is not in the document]
simonpcouch: Still some [snap failures](https://github.com/r-dbi/odbc/actions/runs/8975098066/job/24648931989?pr=796#step:13:592) with additions to error messages that I'm able to reproduce locally. In the case of that failing test, newly seeing `• <SQL> 'SELECT * FROM boopbopbopbeep'`.

src/odbc_result.h
simonpcouch: Thanks for all of the work here! That refactor feels much better.

.github/odbc/install-mariadb-driver.sh
```diff
@@ -0,0 +1,9 @@
+sudo apt-get install -y cmake
```
hadley, line 1: Can you please add a comment here to remind us of why we're not just using `apt-get`?

R/dbi-driver.R
```diff
@@ -63,8 +63,11 @@ setMethod("show", "OdbcDriver",
 #'   name for the OdbcConnection object returned from [dbConnect()]. However, if
 #'   the driver does not return a valid value, it can be set manually with this
 #'   parameter.
-#' @param attributes An S4 object of connection attributes that are passed
+#' @param attributes A list of connection attributes that are passed
 #'   prior to the connection being established. See \link{ConnectionAttributes}.
+#' @param interruptibleExecution Logical.  If `TRUE` calls to SQLExecute and
```
hadley, line 68: Should we mark this as experimental, and to file an issue on GitHub if it's used? I assume we'll eventually take this away once we're confident that it's ok.

R/dbi-driver.R
```diff
@@ -167,6 +170,7 @@ setMethod("dbConnect", "OdbcDriver",
       pwd = NULL,
       dbms.name = NULL,
       attributes = NULL,
+      interruptibleExecution = TRUE,
```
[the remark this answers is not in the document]
hadley: Yeah, I'd suggest `interruptible`

src/odbc_result.h
```diff
@@ -24,7 +24,10 @@ class odbc_error : public Rcpp::exception {
       const std::string& sql,
       Iconv& output_encoder)
       : Rcpp::exception("", false) {
-    std::string m = std::string(e.what()) + "\n<SQL> '" + sql + "'";
+    std::string m = std::string(e.what());
+    if (sql != "" ) {
```
hadley, line 28:
```suggestion
    if (sql != "") {
```

hadley (approved):
Needs a news bullet too

Looks good from my end and I know users will really be excited about this issue. But I have no ability to review the threaded C++ code, so please wait on @gaborcsardi's review before merging.

gaborcsardi:
Some quick comments wrt the new argument and have a (not so) secret option instead. I think an option or an env var is much friendlier for the users than the new argument. 

- If interruptible execution causes you issues (which it might, especially on older systems that barely support C++11), it is much easier to opt out with a global option than updating your code at multiple places, in multiple projects. You might not even own the code for which you want to opt out.
- An option makes the transition much smoother, e.g. you default first to `FALSE` (if you are conservative), and then in the next release to `TRUE`, and ideally at some point drop the option altogether. You can do all this without changing APIs or code that uses odbc.
- You can document the option, e.g. you should mention it in the NEWS, and mention it in the manual page(s), and/or in the README. I agree that options tend to be hidden, because of our lack of conventions for documenting them, but we can try to do better in this regard. In any case I think the benefits of a global option outweigh its clumsiness.

gaborcsardi:
A quick question about concurrency and resources. If there is an interrupt, then `cleanup_fn()` runs concurrently with the async thread, right? 

And after an interrupt, the main thread invalidates the DB handle that the async thread is using? If that's the case, are the ODBC drivers prepared for this?

simonpcouch:
Not doing this myself in case @detule has local changes, but this branch should pass CI with changes merged from `main`!

detule:
HI @gaborcsardi 

Appreciate you taking a look. 
*  Will make the `interruptible` argument default to a global option. Good call.
* Right - if interrupted `cleanup_fn` would run concurrently with the async/execution thread.  Per the ODBC spec this should be fine, since before invalidating the statement as part of the clean up, we first call `SQLCancel`.  In the async thread that should cause the statement to terminate processing.
  * This mode of operation is covered in the ODBC documentation [here](https://learn.microsoft.com/en-us/sql/odbc/reference/syntax/sqlcancel-function?view=sql-server-ver16#canceling-functions-executing-on-another-thread).
  * In testing locally (albeit Linux only) almost everything I tested was well behaved modulo an older vintage of the mysql driver.
  * Note there is a notion of "async" execution in the [ODBC spec](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/asynchronous-execution-notification-method?view=sql-server-ver16).  I would expect **that** feature to be poorly supported across various back-ends and drivers, but this is not that.

tests/testthat/_snaps/utils.md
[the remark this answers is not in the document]
detule: Dang it - hopefully I managed to resolve all the snap discrepancies this time. :crossed_fingers:

R/dbi-driver.R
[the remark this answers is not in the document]
detule: Done, thanks team.

.github/odbc/install-mariadb-driver.sh
detule: Done: Added a note pointing to this PR.  Expanded the description of this PR with detail on which tests are failing and how with the canned mysql driver.

src/odbc_result.h
detule: Done! Thanks.

detule:
> Not doing this myself in case @detule has local changes, but this PR should pass CI with changes merged from `main`!

Thanks @simonpcouch - rebased on top of `main`.  Hopefully it's all green.

gaborcsardi:
> Per the ODBC spec this should be fine, since before invalidating the statement as part of the clean up, we first call SQLCancel

I don't see where that happens, probably in `close()`?

In any case, this looks pretty good then. The only concern I have is that you need to install dev MariaDB to make ODBC run reliably. I would think that many ODBC users are stuck on older Linux systems, so we could potentially break their projects. 

So maybe you could make this feature opt-in for now? And make it the default in the next release?

detule:
> > Per the ODBC spec this should be fine, since before invalidating the statement as part of the clean up, we first call SQLCancel
> 
> I don't see where that happens, probably in `close()`?

Yes, both `set_current_result`, and `close` in the cleanup function will cancel the current result, as needed.

> 
> In any case, this looks pretty good then. The only concern I have is that you need to install the dev MariaDB drivers to make ODBC run reliably. I would think that many ODBC users are stuck on older Linux systems, so we could potentially break their projects.
> 
> So maybe you could make this feature opt-in for now? And make it the default in the next release?

Thanks @gaborcsardi .  Appreciate you taking a look.  On `mariadb` - do you still feel the same way knowing that the only test that the driver seems to be misfiring on is the relatively contrived `dbGetQuery(conn, NA_character_)` ? All other tests, more than three thousand of them, seem to be passing on both Linux and Windows ( admittedly MySQL, not MariaDB ) without issues.  At any rate, I don't have a problem making it opt-in and happy to do it, just a bit concerned that it won't get enough mileage.

gaborcsardi:
> At any rate, I don't have a problem making it opt-in and happy to do it, just a bit concerned that it won't get enough mileage.

You can make a much more informed decision than me. In general I tend to be conservative in these kind of issues, especially with packages that may be used on older systems, like odbc.

You can also make the default depend on whether the session is interactive or not. I would think that interruption is much less important in non-interactive sessions, where an interruption just kills the process, because there are already ways to kill the process. So you could turn it on in interactive sessions only.

r2evans:
(I cannot express sufficiently how much I've wanted this and never knew how to ask for it.)

PR #800, opened 2024-05-07
configure snowflake error logs

simonpcouch:
Closes #798.

This PR will allow for retrieving more informative error messages from Snowflake's ODBC driver in our CI. CI will probably still fail with the first commit in this PR, just with a new message.

PR #802, opened 2024-05-14
update mysql driver

simonpcouch:
~~Cherry-picks [0a255a1](https://github.com/r-dbi/odbc/pull/796/commits/0a255a1bfc87b0e25641085218c9554ae5f71fb6) from #796 so that we can get GHA passing again.~~ Now that #800 is also in, can merge remaining PRs upstream after this is merged and expect GHA to pass.

EDIT: The failures we're seeing on `DBs (Windows) / MySQL` in the last couple weeks are actually unrelated to the changes in the linked PR, which are related to the Linux MySQL installation. mysql.odbc was updated to v8.4.0 a couple weeks ago and that was actually the cause of the issue.

### DIFF
--- a/.github/odbc/install-mariadb-driver.sh
+++ b/.github/odbc/install-mariadb-driver.sh
@@ -1,0 +1,9 @@
+sudo apt-get install -y cmake
+cd /tmp && git clone https://github.com/MariaDB/mariadb-connector-odbc.git connector
+cd connector && git checkout 3.1.17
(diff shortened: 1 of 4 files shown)


simonpcouch:
Ah, I had missed that `DBs (Windows)` had also started failing in later commits to #796. Will look into this further.

hadley (approved):

Issue #808, opened 2024-05-23
CRAN WARNINGs on M1 Macs re: `-Wdeprecated-declarations`


simonpcouch:
CRAN's M1 Mac checks [surface a number of new warnings](https://www.stats.ox.ac.uk/pub/bdr/M1mac/odbc.out) related to deprecated SQL/ODBC deprecations in nanodbc.

<details>
  <summary>WARNING plain text</summary>
  
```
* checking whether package ‘odbc’ can be installed ... [21s/22s] WARNING
Found the following significant warnings:
  connection.cpp:299:3: warning: 'SQLSetConnectAttr' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:370:26: warning: 'SQLGetDiagRec' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:388:26: warning: 'SQLGetDiagRec' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:734:21: warning: 'SQLFreeHandle' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:746:21: warning: 'SQLAllocHandle' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:753:13: warning: 'SQLSetEnvAttr' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:778:25: warning: 'SQLAllocHandle' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:944:25: warning: 'SQLSetConnectAttr' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:1055:26: warning: 'SQLConnect' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
  nanodbc/nanodbc.cpp:1127:26: warning: 'SQLDriverConnect' is deprecated: first deprecated in macOS 10.8 [-Wdeprecated-declarations]
[...]

Issue #807, opened 2024-05-23
CRAN NOTE on installed package size on macOS


simonpcouch:
CRAN's macOS check systems surface the following NOTE on the release version of the package:

```
checking whether package ‘odbc’ can be installed ... [29s/32s] OK
See the [install log](https://www.r-project.org/nosvn/R.check/r-release-macos-arm64/odbc-00install.html) for details.
used C++ compiler: ‘Apple clang version 14.0.0 (clang-1400.0.29.202)’
used SDK: ‘MacOSX11.3.sdk’
checking installed package size ... NOTE
  installed size is 10.8Mb
  sub-directories of 1Mb or more:
    libs  10.0Mb
```

The relevant checks are [here](https://cran.r-project.org/web/checks/check_results_odbc.html). This is a narrow margin over 10Mb--if it's a tricky one we can address in CRAN comments.

Related to #806.

simonpcouch:
<details>
  <summary>Here are those linked logs.</summary>
```
* installing *source* package ‘odbc’ ...
** package ‘odbc’ successfully unpacked and MD5 sums checked
** using staged installation
Found odbc_config cflags and libs!
PKG_CFLAGS=-DHAVE_UNISTD_H -DHAVE_PWD_H -DHAVE_SYS_TYPES_H -DHAVE_LONG_LONG -DSIZEOF_LONG_INT=8 -I/opt/R/arm64/include 
PKG_LIBS=-L/opt/R/arm64/lib -lodbc
** libs
using C++ compiler: ‘Apple clang version 14.0.0 (clang-1400.0.29.202)’
using SDK: ‘MacOSX11.3.sdk’
(cd cctz && \
make libcctz.a CC="clang -arch arm64" CXX="clang++ -arch arm64 -std=gnu++17" CXXFLAGS="-falign-functions=64 -Wall -g -O2 " AR="ar" ARFLAGS=rcs CXXPICFLAGS="-fPIC")
clang++ -arch arm64 -std=gnu++17 -falign-functions=64 -Wall -g -O2  -Wall -Iinclude -std=c++11 -pthread  -O -fPIC -MMD  -c -o time_zone_format.o src/time_zone_format.cc
clang++ -arch arm64 -std=gnu++17 -falign-functions=64 -Wall -g -O2  -Wall -Iinclude -std=c++11 -pthread  -O -fPIC -MMD  -c -o time_zone_if.o src/time_zone_if.cc
[...]

PR #809, opened 2024-05-23
ignore GCC `Wdeprecated-declarations` warnings

### LINKED ISSUE #806: new WARNING on CRAN `r-devel-linux-x86_64-fedora-clang`
Before we send out a new release of the package, we need to address the [new CRAN check warning](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/odbc-00check.html) with R-devel on Fedora Linux:

```
checking whether package ‘odbc’ can be installed ... [171s/79s] WARNING
Found the following significant warnings:
  /usr/local/clang18/bin/../include/c++/v1/locale:3257:1: warning: 'wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>' is deprecated [-Wdeprecated-declarations]
See the [install log](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/odbc-00install.html) for details.
used C++ compiler: ‘clang version 18.1.5’
```

This is:

```
using R Under development (unstable) (2024-05-10 r86529)
using platform: x86_64-pc-linux-gnu
R was compiled by
    clang version 18.1.5
    flang-new version 18.1.5
running under: Fedora Linux 36 (Workstation Edition)
```

None of CRANs other check systems surface this warning.

simonpcouch:
Closes #806 and #808, I think. Expands the existing `# pragma GCC diagnostic ignored "-Wdeprecated-declarations"` to much more cpp code: nearly all of `nanodbc.cpp` and a new section in `connection.cpp`.

Very much open to the idea that this may not be what we want to do long-term, though I believe will do the trick to make a release possible soon (before June 4th, so that new Snowflake wrapper can head out). Will defer to others on whether we ought to file a separate issue and get to work on addressing the use of this deprecated functionality after the upcoming release.

### DIFF
--- a/src/connection.cpp
+++ b/src/connection.cpp
@@ -304,12 +304,17 @@ Rcpp::IntegerVector transactionLevels() {
   return out;
 }
 
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+
 // [[Rcpp::export]]
 void set_transaction_isolation(connection_ptr const& p, size_t level) {
   auto c = (*p)->connection();
   SQLSetConnectAttr(
       c->native_dbc_handle(), SQL_ATTR_TXN_ISOLATION, (SQLPOINTER)level, 0);
 }
+
+# pragma GCC diagnostic pop
 
 // [[Rcpp::export]]
 Rcpp::IntegerVector bigint_mappings() {
--- a/src/nanodbc/nanodbc.cpp
+++ b/src/nanodbc/nanodbc.cpp
@@ -268,6 +268,8 @@ inline std::size_t strarrlen(T (&a)[N])
     return i;
 }
 
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+
 inline void convert(const wide_string_type& in, std::string& out)
 {
 #ifdef NANODBC_USE_BOOST_CONVERT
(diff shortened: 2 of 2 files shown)


hadley:
I don't think these are particularly important concerns, so I think it's fine to ignore for CRAN's sake.

detule:
Hey @simonpcouch 

On #808 I think the warnings are an indication of something perhaps a bit more nefarious - in particular that on s-u's build server the `iODBC` headers ( that carry the deprecation warnings ) are clobbering over the `unixODBC` headers.  I purused this in a [number](https://github.com/R-macos/recipes/issues/41) of [issues](https://github.com/R-macos/recipes/pull/46) with [Simon](https://github.com/R-macos/recipes/issues/48) when I last saw it happen, and he addressed it.   It looks like it has snuck in again.

simonpcouch:
Following discussion in #806, #807, and #808, our approach here will now be:

807 - Since the MacOS builds include all debug symbols and CRAN policy doesn't permit building without them, there's not much we can do here—just note in CRAN comments.
808 - Revisit discussions with Simon re: iODBC headers if this is an issue on submission—note in CRAN comments.

That leaves 806, which is on an incoming check systems build and thus needs to be addressed. I've attempted doing so by reverting the previous deprecations changes and instead applying the existing deprecations on `NANODBC_CODECVT_TYPE` to the remaining three instances.

Issue #814, opened 2024-06-06
MSSQL Server dbWriteTable fails without batch_row = nrow(value)


meztez:
MSSQL Server dbWriteTable fails without batch_row = nrow(value)

``` r
# Something to create a connexion
con <- ...
tb <- mtcars[rep(seq_len(nrow(mtcars)), 200),]
odbc::dbWriteTable(con, name = "#TMP_CARS", value = mtcars, overwrite = TRUE, immediate = TRUE)
odbc::dbWriteTable(con, name = "#TMP_XCARS", value = tb, overwrite = TRUE, immediate = TRUE)
#> Error in eval(expr, envir, enclos): nanodbc/nanodbc.cpp:1783: 00000
#> The incoming tabular data stream (TDS) remote procedure call (RPC) protocol stream is incorrect. Parameter 2 (""): The supplied length is not valid for data type nvarchar(max). Check the source data for invalid lengths. An example of an invalid length is data of nchar type with an odd length in bytes.,
```

``` r
odbc::dbWriteTable(con, name = "#TMP_XCARS", value = tb, overwrite = TRUE, immediate = TRUE, batch_rows = nrow(tb))
```

<sup>Created on 2024-06-06 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

<details>
<summary>Database</summary>

```
> DBI::dbGetInfo(con)

$dbms.name
[1] "Microsoft SQL Server"

$db.version
[1] "14.00.3465"

$host
[1] ""

$port
[1] ""

$sourcename
[1] ""

$drivername
[1] "RStudio SQL Server ODBC Driver"

$odbc.version
[1] "03.52"

$driver.version
[1] "1.5.23.1024"

$odbcdriver.version
[1] "03.80"

$supports.transactions
[1] TRUE

$supports.catalogs
[1] TRUE

$supports.schema
[1] TRUE

$getdata.extensions.any_column
[1] TRUE

$getdata.extensions.any_order
[1] TRUE
```
</details>

<details>
<summary>Session Info</summary>

```
> devtools::session_info() ─ Session info ─ Session info ─────────────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.4.0 (2024-04-24)
 os       Ubuntu 22.04.4 LTS
 system   x86_64, linux-gnu
 ui       RStudio
 language (EN)
 collate  C.UTF-8
 ctype    C.UTF-8
 tz       America/Montreal
 date     2024-06-06
[...]

PR #816, opened 2024-06-11
nanodbc: more frugal with calls to SQLDescribeParam

detule:
Fixes https://github.com/r-dbi/odbc/issues/814

It seems the SIMBA/Professional MSSQL driver does not faithfully return correct data when invoking `SQLDescribeParam` multiple times for the same index.

That aside, the commit below should be considered a performance improvement when writing large data sets and with a `batch_size` a fraction of the total number of rows as the API call is called only once for each column ( previously once per-column-per-batch ).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # odbc (development version)
 
+* SQL Server: Fix issue related to writing when using SIMBA drivers (#816).
+
 # odbc 1.5.0
 
 ## Major changes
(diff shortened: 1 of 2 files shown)


hadley (approved):
Can we capture this problem/fix with a test?

detule:
> Can we capture this problem/fix with a test?

We can write an easy test but it will only be effective once we have a pipeline for the PRO drivers.  I'll make sure to do that once we have that in place.

PR #818, opened 2024-06-13
transition to JWT for snowflake authentication

simonpcouch:
Folks working on the Snowflake integration at Posit recently transitioned to a different auth system for our dev instance. This PR sets up auth via JWT using a repository secret `SNOWFLAKE_PRIVATE_KEY`.

This should be enough to once again get CI passing. :)

PR #820, opened 2024-06-14
update GHA workflows

simonpcouch:
Just the output of `usethis::use_tidy_github_actions()` to resolve the R 3.6 installation failure on CI. Snowflake test will continue to fail until it has #818. Figured this can go in first and then I'll merge the other two PRs from yesterday upstream.


hadley (approved):
Can you please bump the required R version too?

PR #825, opened 2024-06-27
check for unixODBC in `dbConnect(odbc(), ...)`

simonpcouch:
This PR adds a call to `locate_install_unixodbc()` to the main connection path, now that we've "piloted" it in the macOS x Databricks connection path for a release without issue. Closes #782.

PR #826, opened 2024-07-02
update MySQL driver name

simonpcouch:
Seeing failures for **DBs (Windows) / MySQL** on #825 that I believe are unrelated to that PR—seems related to a version change in the driver.


hadley (approved):
I wonder if there's some way to make this more robust.

PR #828, opened 2024-07-09
update docs on dev setup for Oracle

### LINKED ISSUE #734: update Oracle setup docs
Some notes-to-self after an afternoon trying to make this happen... the following from Jim remains true:

https://github.com/r-dbi/odbc/blob/e8b68fa5de44bf3d3cd7cfedb204bf8f7c260f16/vignettes/develop.Rmd#L203

😝

Oracle Cloud Free as a quick setup to test against is a no-go: requires a credit card and, even with that, had some opaque account setup issues that I wasn't able to resolve.

Tried with colima and then reverted to the docker host, but think I'll need to transition back to trying colima again.

simonpcouch:
😵‍💫

Closes #734.

### DIFF
--- a/vignettes/develop.Rmd
+++ b/vignettes/develop.Rmd
@@ -222,10 +222,6 @@ The above example assumes the configured SQLite DSN is called `"SQLite"`.
 
 A huge pain.
 
-## Oracle
-
-A huge pain.
-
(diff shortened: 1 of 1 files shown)


hadley (approved):
🤯

PR #830, opened 2024-07-17
run DBI tests for Oracle

### LINKED ISSUE #743: Run DBI tests for Oracle

simonpcouch:
Closes #743. Started with no skips and iteratively re-ran and added until all of the failures I saw locally were skipped. Anticipating that there may be some OS-specific or driver-specific failures here, so may need to iterate.

Quite a few.😬


hadley (approved):
Obviously we'll need to eventually make more of these pass, but this is a great start 😄

Issue #653, opened 2023-12-07
Support for automatic credential refresh with supported drivers


atheriel:
### Issue Description and Expected Result

Right now if you're using an OAuth2 token instead of a username/password with certain drivers and the token expires, queries start returning errors.

Microsoft has a section on this [in the ODBC 4.0 spec](https://github.com/microsoft/ODBC-Specification/blob/master/ODBC%204.0.md#324-token-expiration), and there's more precise detail in the documentation for the Databricks driver itself:

![manual1](https://github.com/r-dbi/odbc/assets/1448326/60244916-3a38-4353-8fe6-bd9a5a28c327)
![manual2](https://github.com/r-dbi/odbc/assets/1448326/c5688e09-0b87-4e1f-82a9-be72b9a6ebd5)

I'm wondering if it's possible to catch these errors, refresh the tokens (if possible), and retry the query automatically.

### Database

Databricks, but potentially others.

### Reproducible Example

I'm not sure I can provide a reproducible example for this, it's transient and requires all of the OAuth2 stuff to be set up.

<details>
<summary>Session Info</summary>

```r
devtools::session_info()
#> output
```
</details>

fh-mthomson:
We're currently leveraging https://github.com/r-lib/httr2/pull/349 to handle the OAuth token refresh for connecting to Snowflake from Workbench (since, as you noted, external browser auth doesn't necessarily work).

Currently, I'm working through leveraging the same pattern in Connect (but requires setting access / refresh tokens as env vars), but that's proving to be a bit more clunky when needing to update env vars on the fly. 

I'd be happy to discuss further, if helpful!

PR #840, opened 2024-08-27
add context to "Data source name not found..." error

simonpcouch:
Rather than sending folks out to the "Data source name not found" google search results, we could refer to the revamped `odbcListDataSources` docs as a more beginner-accessible starting place to start debugging.

This is stepping a bit further into the territory of "patching" uninformative error messages from the driver manager and driver. Understood if we feel this is outside of the scope of the package.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -214,7 +214,7 @@ contextualize_database_error <- function(cnd_body) {
     cnd_body <-
       c(
         cnd_body,
-        "See {.help [?odbcListDataSources](odbc::odbcListDataSources)} to learn more."
+        "See {.help odbc::odbcListDataSources} to learn more."
       )
   }
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -199,12 +199,26 @@ parse_database_error <- function(msg) {
     right = TRUE
   )
 
+  cnd_body <- contextualize_database_error(cnd_body)
+
   list(
     cnd_context_nanodbc = cnd_context_nanodbc,
     cnd_context_code = cnd_context_code,
     cnd_context_driver = cnd_context_driver,
     cnd_body = cnd_body
   )
+}
+
+contextualize_database_error <- function(cnd_body) {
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -207,6 +209,18 @@ parse_database_error <- function(msg) {
   )
 }
 
+contextualize_database_error <- function(cnd_body) {
```
simonpcouch, line 212: Extracted this out as a function in case we're interested in applying this to other common messages that we can't resolves ourselves behind-the-scenes.

R/utils.R
```diff
@@ -207,6 +209,18 @@ parse_database_error <- function(msg) {
...
+  if (any(grepl("Data source name not found", cnd_body))) {
```
simonpcouch, line 213: Generally wary of `grep()`ping on messages, but from my understanding the unixODBC driver manager doesn't provide translations of its error messages to other languages.

R/utils.R
simonpcouch: https://github.com/r-dbi/odbc/issues/653 is another situation where I imagine us doing this. With an expired token, the databricks message essentially reads like a 403, but all the user actually has to do is refresh their token.

R/utils.R
hadley: Yeah, and the failure mode isn't too bad here. Might be worth noting/checking that `cnd_body` is always length 1 and not missing.

R/utils.R
```diff
@@ -207,6 +209,18 @@ parse_database_error <- function(msg) {
...
+    cnd_body <-
+      c(
+        cnd_body,
+        "See {.help odbc::odbcListDataSources} to learn more."
```
hadley, line 217: Should this be an `i` bullet?

hadley (approved):

R/utils.R
simonpcouch: Good point! Required a bit of refactoring that feels overkill for now but will make extensions straightforward.

Issue #834, opened 2024-08-23
Encoding/Collation problems with 1.5.0 SQL server


soetang:
We have been expriencing encoding problems with odbc 1.5.0 on SQL server. This works fine on odbc 1.4.2

All our databases unfortunately uses the varchar collation "Danish_Norwegian_CI_AS" - with odbc 1.4.2 however we were able to create a connection so that we could correctly read and write to the database. With odbc 1.5.0 the column names are no longer formatted correctly.

Is this a bug or can we change the connection settings so that it works correctly?

``` r
library('magrittr')
conn <- DBI::dbConnect(
  odbc::odbc()
  , Driver = "ODBC Driver 17 for SQL Server"
  , Server = "server"
  , Database = 'database'
  , Trusted_Connection = "Yes"
  , encoding = 'latin1'
  , AutoTranslate = 'no'
)

test_table <- DBI::Id(schema = 'test', table = 'test_table')
df <- tibble::tibble(
  var_char_col = c('kanin', 'ræven', 'ålens', 'ørred'),
  bøvs = 1
)


DBI::dbWriteTable(
  conn
  , name = test_table
  , df
  , field.types = c('var_char_col' = 'varchar(5)')
  , overwrite = TRUE
)

db_data <- DBI::dbReadTable(conn, test_table) %>%
  tibble::as_tibble()

db_data
#> # A tibble: 4 × 2
#>   var_char_col bÃ.vs
#>   <chr>        <dbl>
#> 1 kanin            1
#> 2 ræven            1
#> 3 ålens            1
#> 4 ørred            1
```

Expected value:
``` r
db_data
#> # A tibble: 4 × 2
#>   var_char_col  bøvs
#>   <chr>        <dbl>
#> 1 kanin            1
#> 2 ræven            1
#> 3 ålens            1
#> 4 ørred            1
```

If we remove the encoding parameter - we get correct colum names, however the character vector no longer fit within the varchar(5) datatype even though we only have 5 characters:

``` r
library('magrittr')
conn <- DBI::dbConnect(
  odbc::odbc()
  , Driver = "ODBC Driver 17 for SQL Server"
  , Server = "server"
  , Database = 'database'
  , Trusted_Connection = "Yes"
  # , encoding = 'latin1'
  , AutoTranslate = 'no'
)

test_table <- DBI::Id(schema = 'test', table = 'test_table')
df <- tibble::tibble(
  var_char_col = c('kanin', 'ræven', 'ålens', 'ørred'),
[...]

PR #845, opened 2024-09-06
Feature/independent column name encoding

detule:
This closes https://github.com/r-dbi/odbc/issues/834.

Some history:
* In https://github.com/r-dbi/odbc/pull/531 we made it so that we use the same "output encoder" to convert both the column contents as well as the column names.  This encoder is determined by the `encoding` argument to `dbConnect`.
* In https://github.com/r-dbi/odbc/issues/834 we saw some evidence that there are instances when the table contents may be encoded with `latin1`/8-bit ascii, for example, but the DB driver returns the column names using `UTF-8`.

To handle this, in this PR:
* I separated the contents and field name encoders into separate arguments to `dbConnect`.  New argument is `columnNameEncoding`.
* As a small optimization, I made the encoder resources part of the connection.  Previously they were allocated for each result.

To do before merging:

- [x] Add news item.
- [x] Deal with testthat/windows/utf-8.


tests/testthat/test-driver-sql-server.R
```diff
@@ -328,3 +328,26 @@ test_that("can create / write to temp table", {
...
+  on.exit({
+    dbExecute(conn, "DROP TABLE deleteme_Danish_Norwegian_CI_AS.testschema.deleteme")
+    dbExecute(conn, "DROP SCHEMA IF EXISTS testschema")
+    dbExecute(conn, "USE tempdb")
+    dbExecute(conn, "DROP DATABASE IF EXISTS deleteme_Danish_Norwegian_CI_AS")
+  })
+  DBI::dbExecute(conn, "CREATE DATABASE deleteme_Danish_Norwegian_CI_AS COLLATE Danish_Norwegian_CI_AS")
+  DBI::dbExecute(conn, "USE deleteme_Danish_Norwegian_CI_AS")
+  DBI::dbExecute(conn, "CREATE SCHEMA testschema")
+  tbl_id <- DBI::Id(catalog="deleteme_Danish_Norwegian_CI_AS", schema="testschema", name = "deleteme")
```
simonpcouch, line 349:
```suggestion
  tbl_id <- DBI::Id(catalog = "deleteme_Danish_Norwegian_CI_AS", schema = "testschema", name = "deleteme")
```

tests/testthat/test-driver-sql-server.R
```diff
@@ -328,3 +328,26 @@ test_that("can create / write to temp table", {
   expect_snapshot_warning(sqlCreateTable(con, notTempTblName, values, temporary = TRUE))
   expect_no_warning(sqlCreateTable(con, notTempTblName, values, temporary = FALSE))
 })
+
+# See https://github.com/r-dbi/odbc/issues/834
+test_that("encoding test", {
```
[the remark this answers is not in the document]
detule: Done / thanks

tests/testthat/test-driver-sql-server.R
detule: Done / thanks

src/odbc_connection.h
```diff
@@ -20,12 +22,13 @@ class odbc_connection {
...
   odbc_connection(
-      std::string connection_string,
-      std::string timezone = "UTC",
-      std::string timezone_out = "UTC",
-      std::string encoding = "",
-      bigint_map_t bigint_mapping = i64_to_integer64,
-      long timeout = 0,
+      std::string const& connection_string,
+      std::string const& timezone = "UTC",
+      std::string const& timezone_out = "UTC",
```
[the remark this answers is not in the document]
detule: Yes - was already touching the declaration to add an argument and thought it is a good opportunity to fix that as well.  My OCD got the better of me.

tests/testthat/test-driver-sql-server.R
```diff
@@ -328,3 +328,26 @@ test_that("can create / write to temp table", {
...
+    dbExecute(conn, "USE tempdb")
+    dbExecute(conn, "DROP DATABASE IF EXISTS deleteme_Danish_Norwegian_CI_AS")
+  })
+  DBI::dbExecute(conn, "CREATE DATABASE deleteme_Danish_Norwegian_CI_AS COLLATE Danish_Norwegian_CI_AS")
+  DBI::dbExecute(conn, "USE deleteme_Danish_Norwegian_CI_AS")
+  DBI::dbExecute(conn, "CREATE SCHEMA testschema")
+  tbl_id <- DBI::Id(catalog="deleteme_Danish_Norwegian_CI_AS", schema="testschema", name = "deleteme")
+  DBI::dbWriteTable(conn, name = tbl_id, df, field.types = c('var_char_col' = 'varchar(5)'), overwrite = TRUE)
+  res <- DBI::dbReadTable(conn, tbl_id)
+  expect_identical(df, res);
```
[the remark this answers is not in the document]
detule: Done / thanks

R/dbi-driver.R
```diff
@@ -48,6 +48,10 @@ setMethod("show", "OdbcDriver",
 #'   not using UTF-8 you will need to set the encoding to get accurate
 #'   re-encoding. See [iconvlist()] for a complete list of available encodings
 #'   on your system. Note strings are always returned `UTF-8` encoded.
+#' @param columnNameEncoding The text encoding for column names used on the
```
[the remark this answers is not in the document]
detule: * Happy to rename / i'll wait and see if @hadley also has a preference.
* On where to slot the new argument: I, at times, struggle with making changes to an existing API that might break existing code.  However I also don't feel strongly about it and can go either way.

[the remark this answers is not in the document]
hadley: Since it's already after `...` there's no breakage because all args after `...` can only be specified by full name. So I'd put it after `encoding`.

R/dbi-driver.R
```diff
@@ -48,6 +48,10 @@ setMethod("show", "OdbcDriver",
...
+#'   Database.  May be different than the `encoding` argument.  Defaults to
+#'   empty string which is equivalent to returning the column names without
```
hadley, line 53: I'm slightly surprised that this doesn't default to `encoding`. Are the column names and values encoded differently that often?

R/dbi-driver.R
```diff
@@ -174,11 +178,13 @@ setMethod("dbConnect", "OdbcDriver",
       dbms.name = NULL,
       attributes = NULL,
       interruptible = getOption("odbc.interruptible", interactive()),
+      columnNameEncoding = "",
       .connection_string = NULL) {
     check_string(dsn, allow_null = TRUE)
     check_string(timezone, allow_null = TRUE)
     check_string(timezone_out, allow_null = TRUE)
     check_string(encoding, allow_null = TRUE)
+    check_string(columnNameEncoding, allow_null = TRUE)
```
hadley, line 187: What does `NULL` mean here?

hadley (approved):

R/dbi-driver.R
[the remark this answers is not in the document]
detule: Good points all - done.

R/dbi-driver.R
detule: Hi Hadley:

The issue that precipitated this PR is an instance where the content is encoded `latin1` but the names are `UTF-8`.  Prior to 1.5.0 the names were never de/coded regardless how how the `encoding` argument was specified, so in a sense we are just reverting to prior behavior, while retaining the flexibility that @shrektan added in 1.5.0.

R/dbi-driver.R
detule: Good question - i think it doesn't make sense, and changed it.
In fact, I think allowing `NULL`s for the remaining arguments that are passed to `odbc_connect` ( timezone, timeout, encoding, etc ) also doesn't make sense, as `Rcpp` won't allow that.  I'll open a chore to update those checks.

detule: Cleaned up NULL handling of various arguments elsewhere.

hadley:
I think this is good to merge?

simonpcouch:
Thumbs up from me!

Issue #841, opened 2024-08-28
dbDataType chooses `VARCHAR(255)` even when `nchar` > 255


metanoid:
For a connection of type `NetezzaSQL`, the `dbDataType` function returns `"VARCHAR(255)"` for seemingly all character vectors, regardless of the number of characters of the longest string in that vector

```r
library(DBI)
con <- dbConnect(odbc::odbc(), dsn = "Netezza")
test <- paste0(rep("a", 1000), collapse = "")
inferred_type <- dbDataType(con, test)
```
This returns: `"VARCHAR(255)"` even though there are clearly more than 255 characters required

detule:
Hi @metanoid:

I don't have a `Netezza` instance around so shooting from the hip a bit.  Do you have the ability to test a development branch?  If so, can you see if [this version](https://github.com/detule/odbc/tree/testing/netezzaDataType) performs better with your mini example?

Thanks

metanoid:
Hi,

Thanks for trying!  I think something is missing from the commit.  See output message:

``` r
devtools::load_all(".")

#>ℹ Loading odbc
#>Error in add_classes_to_exports(ns = nsenv, package = package, exports = exports,  : 
#>  in package ‘odbc’ classes NetezzaSQL were specified for export but not defined
#>In addition: Warning message:
#>package ‘testthat’ was built under R version 4.3.3 
```

detule:
Sorry about that @metanoid.
Should be better now, hopefully.  Let me know what you find.

Thanks again for testing.

metanoid:
It works!
```r
library(DBI)
con <- dbConnect(odbc::odbc(), dsn = "Netezza")
test <- paste0(rep("a", 1000), collapse = "")
inferred_type <- dbDataType(con, test)


#"varchar(1000)"
```

I'd be interested to hear why this works?  

Thanks!

PR #847, opened 2024-09-10
Netezza / custom odbcDataType

detule:
Hello:

This closes https://github.com/r-dbi/odbc/issues/841

* At some point I will investigate if there are any back ends that don't support the `VARCHAR(n)` syntax.  At that point we can maybe make this part of the generic method.
* At any rate, have seen a few `Netezza` issues come up; declaring a new class seems appropriate.

To do before merging:

- [x] Add NEWS item.


simonpcouch (approved):
Don't have a Netezza instance to test against either, but I'm convinced! Agreed that a new class is in scope. Looks clean!

PR #853, opened 2024-10-09
Add support for viewer-based Databricks & Snowflake credentials

atheriel:
This commit expands the `databricks()` and `snowflake()` helpers to support the viewer-based OAuth credentials [recently introduced in Posit Connect][0].

It is designed to support writing Shiny apps that take the following form:

```r
library(shiny)

ui <- fluidPage(textOutput("user"))

server <- function(input, output, session) {
  conn <- reactive({
    pool::dbPool(odbc::snowflake(), session = session)
  })

  output$user <- renderText({
    res <- DBI::dbGetQuery(conn(), "SELECT CURRENT_USER()")
    res[[1]]
  })
}

shinyApp(ui = ui, server = server)
```

Checks for viewer-based credentials are designed to fall back gracefully to existing authentication methods. This is intended to allow users to -- for example -- develop and test a Shiny app that uses Databricks or Snowflake credentials in desktop RStudio or Posit Workbench and deploy it with no code changes to Connect.

In addition, making the caller pass a `session` argument (rather than just detecting, for example, if we're in a Shiny server context) is very intentional: it makes them express that they *want* viewer-based credentials, as opposed to using shared credentials for all viewers. It seems likely to reduce the number of cases where publisher credentials are used unexpectedly.

Internally we implement the raw HTTP calls to make the token exchange with Connect to avoid taking a dependency on the `connectapi` package.

Unfortunately, there are no new tests for this change; it's pretty hard to emulate what Connect is doing here without extensive mocking.

[0]: https://docs.posit.co/connect/user/oauth-integrations/

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,6 +31,7 @@ Imports:
 Suggests:
     covr,
     DBItest,
+    httr2,
     knitr,
     magrittr,
     rmarkdown,
--- a/R/driver-databricks.R
+++ b/R/driver-databricks.R
@@ -31,6 +31,8 @@ NULL
 #'   default name.
 #' @param uid,pwd Manually specify a username and password for authentication.
 #'   Specifying these options will disable automated credential discovery.
+#' @param session A Shiny session object, when using viewer-based credentials on
+#'   Posit Connect.
 #' @param ... Further arguments passed on to [`dbConnect()`].
 #'
 #' @returns An `OdbcConnection` object with an active connection to a Databricks
@@ -42,6 +44,16 @@ NULL
 #'   odbc::databricks(),
 #'   httpPath = "sql/protocolv1/o/4425955464597947/1026-023828-vn51jugj"
 #' )
+#'
+#' # Use credentials from the viewer (when possible) in a Shiny app
+#' # deployed to Posit Connect.
+#' server <- function(input, output, session) {
+#'   conn <- DBI::dbConnect(
+#'     odbc::databricks(),
+#'     httpPath = "sql/protocolv1/o/4425955464597947/1026-023828-vn51jugj",
+#'     session = session
+#'   )
+#' }
 #' }
 #' @export
 databricks <- function() {
@@ -64,6 +76,7 @@ setMethod("dbConnect", "DatabricksOdbcDriver",
            HTTPPath,
            uid = NULL,
            pwd = NULL,
+           session = NULL,
            ...) {
     call <- caller_env()
     # For backward compatibility with RStudio connection string
@@ -74,6 +87,7 @@ setMethod("dbConnect", "DatabricksOdbcDriver",
     check_string(driver, allow_null = TRUE, call = call)
     check_string(uid, allow_null = TRUE, call = call)
     check_string(pwd, allow_null = TRUE, call = call)
+    check_shiny_session(session, allow_null = TRUE, call = call)
 
     args <- databricks_args(
       httpPath = if (missing(httpPath)) HTTPPath else httpPath,
@@ -82,6 +96,7 @@ setMethod("dbConnect", "DatabricksOdbcDriver",
       driver = driver,
       uid = uid,
       pwd = pwd,
+      session = session,
       ...
     )
     inject(dbConnect(odbc(), !!!args))
@@ -94,6 +109,7 @@ databricks_args <- function(httpPath,
                             driver = NULL,
                             uid = NULL,
                             pwd = NULL,
+                            session = NULL,
                             ...) {
   host <- databricks_host(workspace)
 
@@ -104,18 +120,21 @@ databricks_args <- function(httpPath,
     useNativeQuery = useNativeQuery
   )
 
-  auth <- databricks_auth_args(host, uid = uid, pwd = pwd)
+  auth <- databricks_auth_args(host, uid = uid, pwd = pwd, session = session)
   all <- utils::modifyList(c(args, auth), list(...))
 
   arg_names <- tolower(names(all))
   if (!"authmech" %in% arg_names && !all(c("uid", "pwd") %in% arg_names)) {
-    abort(
-      c(
-        "x" = "Failed to detect ambient Databricks credentials.",
-        "i" = "Supply `uid` and `pwd` to authenticate manually."
-      ),
-      call = quote(DBI::dbConnect())
+    msg <- c(
+      "Failed to detect ambient Databricks credentials.",
+      "i" = "Supply {.arg uid} and {.arg pwd} to authenticate manually."
     )
+    if (running_on_connect()) {
+      msg <- c(
+        msg, "i" = "Or pass {.arg session} for viewer-based credentials."
+      )
+    }
+    cli::cli_abort(msg, call = quote(DBI::dbConnect()))
   }
 
   all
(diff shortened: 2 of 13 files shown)


DESCRIPTION
```diff
@@ -24,6 +24,7 @@ Imports:
     cli,
     DBI (>= 1.0.0),
     hms,
+    httr2,
```
hadley, line 27: httr2 is a fairly meaty dependency. Would it be better to put it in suggests?

R/driver-databricks.R
```diff
@@ -42,6 +43,15 @@ NULL
 #'   odbc::databricks(),
 #'   httpPath = "sql/protocolv1/o/4425955464597947/1026-023828-vn51jugj"
 #' )
+#'
+#' # Use credentials from the viewer (when possible) in a Shiny app:
```
hadley, line 47: when deployed on connect?

R/driver-databricks.R
```diff
@@ -202,7 +215,19 @@ databricks_user_agent <- function() {
   user_agent
 }
 
-databricks_auth_args <- function(host, uid = NULL, pwd = NULL) {
+databricks_auth_args <- function(host, uid = NULL, pwd = NULL, session = NULL) {
+  # If a session is supplied, any viewer-based auth takes precedence.
+  if (!is.null(session)) {
+    access_token <- connect_viewer_token(session, paste0("https://", host))
```
hadley, line 221: Should we emit something here? Seems like it would be useful to know when you've opted-in and it's not working. (And it wouldn't be too annoying to see that message when running locally.)

R/utils.R
```diff
@@ -454,3 +454,50 @@ replace_or_append <- function(lines, pattern, replacement) {
...
   lines
 }
+
+# Request an OAuth access token for the given resource from Posit Connect. The
+# OAuth token will belong to the user owning the given Shiny session.
+connect_viewer_token <- function(session, resource) {
+  # Ensure we're running on Connect.
+  server_url <- Sys.getenv("CONNECT_SERVER")
+  api_key <- Sys.getenv("CONNECT_API_KEY")
+  if (nchar(server_url) == 0 || nchar(api_key) == 0) {
```
hadley, line 464: Or maybe there's some debug mode you can activate that additionally prints a message for each of these early returns?

R/utils.R
```diff
@@ -454,3 +454,50 @@ replace_or_append <- function(lines, pattern, replacement) {
...
+    return(NULL)
+  }
+
+  # See: https://docs.posit.co/connect/api/#post-/v1/oauth/integrations/credentials
+  req <- httr2::request(server_url)
+  req <- httr2::req_url_path_append(
+    req, "__api__/v1/oauth/integrations/credentials"
+  )
+  req <- httr2::req_headers(req,
+    Authorization = paste("Key", api_key), .redact = "Authorization"
```
hadley, line 486: This is a custom auth protocol, right? Not a standard. (i.e. it's fine that httr2 doesn't have an existing helper for this)

R/utils.R
```diff
@@ -454,3 +454,50 @@ replace_or_append <- function(lines, pattern, replacement) {
...
+  )
+  req <- httr2::req_body_form(
+    req,
+    grant_type = "urn:ietf:params:oauth:grant-type:token-exchange",
+    subject_token_type = "urn:posit:connect:user-session-token",
+    subject_token = token,
+    resource = resource
+  )
+
+  # TODO: Do we need more precise error handling?
```
hadley, line 496: If this request fails for some reason do you want to error or fall back to the next auth method?

hadley (commented):
Is there some place that this will get an integration test?

R/driver-databricks.R
atheriel: Modified this so you see the following when running locally:
```
! Ignoring `sesssion` parameter.
ℹ Viewer-based credentials are only available when running on Connect.
```

R/utils.R
atheriel: I think that if you get this far we should probably error, because it means that something is misconfigured.

R/utils.R
atheriel: Modified this a bit so that there is only one early return, and it prints an informative message.

R/utils.R
hadley: With `Key` and not `Bearer` as a prefix?

atheriel:
FWIW I ran the general user-facing API design by @jcheng5 in some depth yesterday and he was on board.

hadley:
LGTM. Feel free to squash-merge when you're done.

tnederlof:
I got a chance to test this @atheriel, and it worked as expected, except I was unable to get the expected message on Workbench with Databricks (with `session = session` defined).

This is such a nice improvement. I will outline my experience below:

## Snowflake

With the following code in my app:
```
dbPool(
  odbc::snowflake(),
  warehouse = "DEFAULT_WH",
  database = "<DB_NAME_HERE>",
  schema = "PUBLIC",
  account = "<ACCOUNT_HERE>"
)
```

Running on Workbench works with no extra messages and the app deploys successfully to Connect but the data pull fails with this in the logs:
```
Error in DBI::dbConnect() : 
  ✖ Failed to detect ambient Snowflake credentials.
ℹ Supply `uid` and `pwd` to authenticate manually.
```

Changing the code to:
```
dbPool(
  odbc::snowflake(),
  warehouse = "DEFAULT_WH",
  database = "<DB_NAME_HERE>",
  schema = "PUBLIC",
  account = "<ACCOUNT_HERE>",
  session = session
)
```

Runs on Workbench successfully with the expected message:
```
! Ignoring `sesssion` parameter.
ℹ Viewer-based credentials are only available when running on Connect.
```

It publishes successfully to Connect, and with the OAuth integration assigned and the user logged in, the data displays as expected.

## Databricks

With the following code in my app:
```
dbPool(
  odbc::databricks(),
  httpPath = Sys.getenv("DATABRICKS_PATH")
)
```

Running on Workbench works with no extra messages and the app deploys successfully to Connect but the data pull fails with this in the logs:
```
Error in DBI::dbConnect() : 
  ✖ Failed to detect ambient Databricks credentials.
ℹ Supply `uid` and `pwd` to authenticate manually.
```

Changing the code to:
```
dbPool(
  odbc::databricks(),
  session = session
)
```

Runs on Workbench successfully but I never received the message I expected to get saying its ignoring the `session` paramter.

It publishes successfully to Connect, and with the OAuth integration assigned and the user logged in, the data displays as expected.

hadley:
Can we detect if we're running connect and/or workbench and add an extra hint to the error?

```
Error in DBI::dbConnect() : 
✖ Failed to detect ambient Snowflake credentials.
ℹ Supply `uid` and `pwd` to authenticate manually.
ℹ Or pass `session = session` for viewer-based credentials
```

tnederlof:
I like adding that to make users aware especially if they accidentally omit session since they won't be used to passing it before.

I have seen the Connect team use `if (Sys.getenv("RSTUDIO_PRODUCT") == "CONNECT") {}` to test if code is running on Connect, that might be better than the [current method of checking](https://github.com/r-dbi/odbc/blob/1b084f25f2e6475a1ee4b3f9caf0ae9c4fc583a2/R/utils.R#L486) for the presence of `CONNECT_SERVER` or `CONNECT_API_KEY` because while those are automatically set when running on Connect, they are commonly set in sessions on desktop and Workbench when using [connectapi](https://github.com/rstudio/connectapi) or when making direct API calls.

atheriel:
Pushed a bunch of unit tests to cover the new errors and info conditions. We now also hint about viewer-based credentials on Connect, and are more careful about what "running on Connect" means:

```
! Failed to detect ambient Snowflake credentials.
i Supply `uid` and `pwd` to authenticate manually.
i Or pass `session` for viewer-based credentials. <-- shown only on Connect
```

With Trevor's testing I now consider this safe enough to merge.

PR #855, opened 2024-10-09
Add support for default application value within `snowflake()`

blairj09:
This aligns the behavior of `snowflake()` with `databricks()` so that in each case the user identifier is set to `r-odbc/{version}` by default.

### DIFF
--- a/R/driver-snowflake.R
+++ b/R/driver-snowflake.R
@@ -193,11 +193,11 @@ snowflake_args <- function(account = Sys.getenv("SNOWFLAKE_ACCOUNT"),
 
   # Set application value and respect the Snowflake Partner environment variable, if present.
   if (is.null(all$application)) {
-    all$application <- ifelse(
-      nchar(Sys.getenv("SF_PARTNER")) != 0,
-      all$application <- Sys.getenv("SF_PARTNER"),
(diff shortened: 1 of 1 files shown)


R/driver-snowflake.R
```diff
@@ -191,9 +191,13 @@ snowflake_args <- function(account = Sys.getenv("SNOWFLAKE_ACCOUNT"),
...
-  # Respect the Snowflake Partner environment variable, if present.
-  if (is.null(all$application) && nchar(Sys.getenv("SF_PARTNER")) != 0) {
-    all$application <- Sys.getenv("SF_PARTNER")
+  # Set application value and respect the Snowflake Partner environment variable, if present.
+  if (is.null(all$application)) {
+    all$application <- ifelse(
+      nchar(Sys.getenv("SF_PARTNER")) != 0,
+      all$application <- Sys.getenv("SF_PARTNER"),
+      paste0("r-odbc/", packageVersion("odbc"))
+    )
```
hadley, line 200: Can you please make this a regular if statement? https://style.tidyverse.org/syntax.html#if-statements

blairj09: Done 😁

hadley (approved):

PR #857, opened 2024-10-28
snowflake: runtime driver config checks

detule:
Hi @simonpcouch:

Following up on a conversation we had a few weeks ago where I noticed that the simba `snowflake` drivers on macOS suffer from similar issues as the ones you ran into with `databricks`.  In particular, and this is one that bit me as well, the simba configuration file (.ini) needs to specify `DriverManagerEncoding=UTF-16`.  The OEM SNOWFLAKE driver, for example, looks to be configured for `iODBC`, and has a different value that doesn't work for us.

In this PR I expanded some of the checks you had written for `databricks` to  `snowflake` as well, with the following changes:
* Rather than modifying the config by default, I made the `snowflake` code path throw a warning.  Left the `databricks` default behavior unchanged.
* Added an `option` to disable to config check altogether, for both.
* Moved the config check ( whether warn or modify ) further down the call stack so as to be able to read the decided driver.  This to avoid a situation where we may be warning about, or modifying one driver, but the user has specified a different one.

The last bullet in particular, I think is why the diff ended up being (much) larger than I originally would have anticipated.

Rather than duplicating methods, tried to make some of the methods "simba" generic ( tried to keep those in `utils.R` ), while others that were specific one of the two back ends I moved to `driver-*.R`. We are close to perhaps needing to make some of those specific methods S3 or S4, but at this point IMO, that seems like a bit of an overkill.

Cheers.

TODO:
- [x] Add news item


hadley:
@atheriel can you take a look at this too please?

R/utils.R
```diff
@@ -288,28 +288,47 @@ check_attributes <- function(attributes, call = caller_env()) {
...
+    if ( action == "modify" ){
+      fun <- cli::abort
+    }
+    func(
       c(
-        "Unable to locate the needed spark ODBC driver.",
-        i = "Please install the needed driver from https://www.databricks.com/spark/odbc-drivers-download."
+        "Unable to locate the config for the ODBC driver.",
+        i = paste("Please make sure you've installed the appropriate driver",
+          "for your platform and back end."),
```
simonpcouch, line 321:
```suggestion
        i = "Please make sure you've installed the appropriate driver
             for your platform and back end."),
```

simonpcouch: GitHub doesn't seem to properly format these changes, but cli allows you to add newlines in error messages with no further formatting, i.e. no need to `paste()`. This can also apply to the next two lines, but I think these three `i` elements could likely be collapsed into a single:

```
i = "Please install the needed driver from {substitute_in_the_appropriate_link_based_on_dbms}"
```

R/utils.R
```diff
@@ -369,51 +388,48 @@ error_install_unixodbc <- function(call) {
...
+  simba_lines_new <- res$new_lines
+  res <- replace_or_append(
+    lines = simba_lines_new,
+    key_pattern = "^DriverManagerEncoding=",
+    accepted_value = "UTF-16|utf-16",
     replacement = "DriverManagerEncoding=UTF-16"
   )
+  if (action != "modify" && res$modified) {
+    cli::cli_warn(
+      paste0("Detected potentially unsafe driver settings. ",
```
simonpcouch, line 418: Same comment re: no need to `paste0()` messages.

R/utils.R
```diff
@@ -288,28 +288,47 @@ check_attributes <- function(attributes, call = caller_env()) {
...
     error_install_unixodbc(call)
   }
 
-  spark_config <- locate_config_spark()
-  if (length(spark_config) == 0) {
-    abort(
+  simba_config <- locate_config_callback()
+  if (length(simba_config) == 0) {
+    func <- cli::warn
+    if ( action == "modify" ){
```
simonpcouch, line 314:
```suggestion
    if (action == "modify"){
```

[Tidy style](https://style.tidyverse.org/syntax.html)--these changes apply to a few places in this PR, incl for before and after curly braces as well.

R/utils.R
```diff
@@ -369,51 +388,48 @@ error_install_unixodbc <- function(call) {
...
+  res <- replace_or_append(
+    lines = simba_lines_new,
+    key_pattern = "^DriverManagerEncoding=",
+    accepted_value = "UTF-16|utf-16",
     replacement = "DriverManagerEncoding=UTF-16"
   )
+  if (action != "modify" && res$modified) {
+    cli::cli_warn(
+      paste0("Detected potentially unsafe driver settings. ",
+       "Please consider revising the `DriverManagerEncoding` setting in ",
```
simonpcouch, line 419: Could we explicitly recommend a driver setting here if we do warn? As a user, this wouldn't feel particularly actionable to me. What was the problematic setting and in what file was it set?

R/utils.R
```diff
@@ -288,28 +288,47 @@ check_attributes <- function(attributes, call = caller_env()) {
...
 # apple + spark drive config (#651) --------------------------------------------
-configure_spark <- function(call = caller_env()) {
+# Method will attempt to:
+# 1. Locate the simba driver config using the user supplied callback.  Callback
+# method defaults to locate_config_spark.
+# 2. Inspect the config for some settings that can impact how our package
+# performs.
+# 3. If action == "modify" then we attempt to modify the config in-situ.
+# 4. Otherwise we throw a warning asking the user to revise.
+configure_simba <- function(locate_config_callback = (function() character()),
```
simonpcouch, line 298:
```suggestion
configure_simba <- function(locate_config_callback = character(),
```

Poking at this for a bit, it's unclear to me why this is a function rather than a character string? In this function, we just call the inputted function which is inlined as returning a character everywhere it's currently used, and then use that character result as usual. Maybe this function just takes the config file as a string and we pass e.g. `snowflake_simba_config(args$driver)` rather than `function() snowflake_simba_config(args$driver)`?

R/driver-databricks.R
```diff
@@ -33,7 +37,7 @@ NULL
 #'   Specifying these options will disable automated credential discovery.
 #' @param ... Further arguments passed on to [`dbConnect()`].
 #'
-#' @returns An `OdbcConnection` object with an active connection to a Databricks
+#' @return An `OdbcConnection` object with an active connection to a Databricks
```
simonpcouch, line 40:
```suggestion
#' @returns An `OdbcConnection` object with an active connection to a Databricks
```

`@return` is superseded in favor of `@returns`.

simonpcouch (commented):
A first pass through! I'm really happy we're aiming to resolve this issue for Snowflake users, too.

I said this in PM as well but wanted to reiterate once more here in case other folks support (feel free to ignore and I won't bug you any longer😜): I think the same issue should lead to the same outcome across drivers, and I'd argue that we ought to just edit the file as we do with Databricks. Especially given that this PR introduces `odbc.no_config_override` (on board!), I'm fine with the more invasive approach given that it allows for this connection to "just work" for users. No longer needing to manage the different behavior for difference DBMS', I think, would greatly simplify this PR as well.

tests/testthat/test-utils.R
```diff
@@ -248,9 +271,15 @@ test_that("configure_unixodbc_spark() writes reasonable entries", {
     con = spark_config_path
   )
 
-  configure_unixodbc_spark(
+  expect_snapshot(configure_unixodbc_simba(
+    unixodbc_install = unixodbc_install_path,
+    simba_config = spark_config_path,
+    action = "warn"
+  ))
+  configure_unixodbc_simba(
```
atheriel, line 279: Might be missing an `expect_snapshot()` here?

R/utils.R
atheriel, line 419: Nit: you can use cli inline formatting here (and elsewhere) instead of manually-including backticks:
```suggestion
       "Please consider revising the {.arg DriverManagerEncoding} setting in ",
```

atheriel (commented):
I think efforts to make drivers "just work" are worthwhile for sure, so in that sense I'm supportive. However, it's a little unclear to me from the code whether this only applies to macOS users or if it's designed to help Linux users, too.

If the former, I think the code should be more explicit about that. If the latter, I think the Linux paths are missing from `spark_simba_config()` and `snowflake_simba_config()`.

detule:
> A first pass through! I'm really happy we're aiming to resolve this issue for Snowflake users, too.
> 
> I said this in PM as well but wanted to reiterate once more here in case other folks support (feel free to ignore and I won't bug you any longer😜): I think the same issue should lead to the same outcome across drivers, and I'd argue that we ought to just edit the file as we do with Databricks. Especially given that this PR introduces `odbc.no_config_override` (on board!), I'm fine with the more invasive approach given that it allows for this connection to "just work" for users. No longer needing to manage the different behavior for difference DBMS', I think, would greatly simplify this PR as well.

Thanks Simon - appreciate you taking a look.  Apologies for misinterpreting your earlier note on whether to modify in place or just warn. I changed `snowflake` so that the behavior is the same as `databricks` ( modify ).

I did leave the `warn` code in the method, in case we want to pursue that option in the future.  However, happy to remove if you have concerns.

R/utils.R
detule: Done / thanks.

R/utils.R
detule: Done / thanks!

R/utils.R
detule: Thanks! Still looking for a good tidy style vim plugin that works in my existing configuraion.  Done.

R/utils.R
detule: Hey Simon:

Wrote it as a callback because in may cases ( not macos, or  `no_config_override` is set ) it may not get executed at all.  Can move those checks outside of the method, but it felt like they belong within ( and they would need to be duplicated for each call ).

Let me know if you think it's worth trying to chase this down further / refactor more.

R/driver-databricks.R
detule: Sweet / reverted.  Thanks

tests/testthat/test-utils.R
detule: Hey @atheriel - thanks for taking a look!
I am not sure I follow - wasn't able to identify the spot where I can throw a snapshot comparison.

R/utils.R
detule: Done! Thanks.

tests/testthat/test-utils.R
atheriel: I might have misinterpreted, sorry.

detule:
> I think efforts to make drivers "just work" are worthwhile for sure, so in that sense I'm supportive. However, it's a little unclear to me from the code whether this only applies to macOS users or if it's designed to help Linux users, too.
> 
> If the former, I think the code should be more explicit about that. If the latter, I think the Linux paths are missing from `spark_simba_config()` and `snowflake_simba_config()`.

Hey @atheriel - appreciate the feedback.

It's the former - at least for now.  Most of these issues are coming from the fact that vendors seem to be distributing **MacOS drivers** that, out-of-the-box, are configured to work when paired with iODBC, rather than unixODBC.  Our package, can't switch between `unixODBC` and `iODBC` at run-time; rather, that's determined at build-time ( and is configured for `unixODBC`).   I haven't seen these issues with the OEM drivers for Linux.

At some point, wouldn't mind doing some investigative work to determine what it would take to be able to adjust our posture relative to the driver managers more dynamically.

- The `configure_simba` method opens with

```
if (!is_macos()) {
  return(invisible())
}
```
- I added comments prior to the invocation in `driver-snowflake|databricks` to remind the reader of the same.

Hopefully that makes it more explicit.

R/utils.R
simonpcouch: Gotcha! Since those callbacks take some time (and could even raise some unanticipated error), you opt to wait to evaluate the code until you're sure you need its result.

We're actually lucky here in that [R will evaluate function arguments lazily](https://adv-r.hadley.nz/functions.html#lazy-evaluation). So, if you pass a function call as an argument to `f()`, do some checks, and then reference the result of the function call after the checks, the function call will never actually evaluate if the check returns beforehand. e.g.:

``` r
fn_a <- function() {
  cli::cli_abort("Stoppp!")
}

fn_b <- function(x) {
  if (TRUE) {
    cli::cli_abort("{.fun fn_a} hasn't evaluated yet!")
  }
  
  fn_a
}

fn_b(x = fn_a())
#> Error in `fn_b()`:
#> ! `fn_a()` hasn't evaluated yet!
```

<sup>Created on 2024-11-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

...the implication here being that we can pass `snowflake_simba_config(args$driver)` directly as the function argument and it won't actually run until those Not macOS or No `no_config_override` checks have evaluated.

simonpcouch (commented):
Looking good! Leaving review as Comment one more time just to get eyes on the callback revision before sending this one in. Thanks for making this happen!

detule:
> Looking good! Leaving review as Comment one more time just to get eyes on the callback revision before sending this one in. Thanks for making this happen!

Thanks Simon.  Good suggestions all of 'em; tried to address all, hopefully I didn't miss anything.

Thanks again for the feedback.

R/utils.R
```diff
@@ -369,51 +385,51 @@ error_install_unixodbc <- function(call) {
...
+  res <- replace_or_append(
+    lines = simba_lines_new,
+    key_pattern = "^DriverManagerEncoding=",
+    accepted_value = "UTF-16|utf-16",
     replacement = "DriverManagerEncoding=UTF-16"
   )
+  if (action != "modify" && res$modified) {
+     warnings <- c(warnings, c("*" = "Please consider revising the
+       {.arg DriverManagerEncoding} field in {simba_config} and setting its
+       value to 'UTF-16'"))
```
simonpcouch, line 415:
```suggestion
     warnings <- c(warnings, c("*" = "Please consider revising the
       {.arg DriverManagerEncoding} field in {.file {simba_config}} and setting its
       value to {.val UTF-16}."))
```

R/utils.R
```diff
@@ -369,51 +385,51 @@ error_install_unixodbc <- function(call) {
...
     replacement = paste0("ODBCInstLib=", unixodbc_install)
   )
-
-  spark_lines_new <- replace_or_append(
-    lines = spark_lines_new,
-    pattern = "^DriverManagerEncoding=",
+  warnings <- character()
+  if (action != "modify" && res$modified) {
+     warnings <- c(warnings, c("*" = "Please consider revising the {.arg ODBCInstLib}
+       field in {.file {simba_config}} and setting its value to {unixodbc_install}"))
```
simonpcouch, line 403:
```suggestion
       field in {.file {simba_config}} and setting its value to {.val {unixodbc_install}}"))
```

R/utils.R
simonpcouch: Just aligning the markup in these two messages. These changes will result in snap changes as well!

R/driver-databricks.R
```diff
@@ -17,6 +17,10 @@ NULL
 #' Workbench or the Databricks CLI on desktop.
 #' All of these credentials are detected automatically if present using
 #' [standard environment variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
+#' In addition, on macOS platforms, the `dbConnect` method will check
```
simonpcouch, line 20:
```suggestion
#' In addition, on macOS platforms, the `dbConnect()` method will check
```

simonpcouch (approved):
Looking good. Thank you, @detule!

PR #860, opened 2024-10-30
ci: keep up with mysql driver updates

detule:
We should probably think about how we can avoid needing to do this in the future.

Couple of options:
* We can ask `choco` to install a fixed version, I believe; or
* Extract the driver version programmatically and inject it into the connection string.

First one is easy; the second would allow us to move forward with the driver versions without upkeep.


detule:
Ended up going with the latter.

simonpcouch (approved):
Perfect. Thanks for making this happen.

Issue #518, opened 2022-11-04
Consider adding unicode support to SQL server


erikvona:
Currently, the implementation for SQL server does not allow for unicode support when writing tables, quoting literals, or using parameters.

This leads to tables being created without unicode support or special unicode characters, and queries being created with literals without special unicode characters, leading to a comparison with a non-ascii unicode character returning the wrong results.

There have been some issues, like https://github.com/r-dbi/DBI/issues/215, where it's suggested to encode strings as bytes, but that results in fragile code that relies on specific SQL server settings and locales, and only solves specific issues like writing tables. I'd like to write code that "just works" if at all possible.

``` r
library(DBI)
#> Warning: package 'DBI' was built under R version 4.1.3
library(odbc)
conn <- dbConnect(odbc(), driver = "ODBC Driver 17 for SQL Server", server = "localhost", database = "example", trusted_connection = "yes", encoding = "utf-8")
unicode_string <- "\u2915"
odbc::dbQuoteString(conn, unicode_string) # Expected: <SQL> N'⤕'
#> <SQL> '<U+2915>'
odbc::odbcDataType(conn, unicode_string) # Expected: nvarchar(255)
#> [1] "varchar(255)"
dbGetQuery(conn, "SELECT ?", params = list(unicode_string)) # Expected: 1  '⤕'
#>    
#> 1 ?

# Note that the connection does support it and properly returns the expected character
dbGetQuery(conn, "SELECT N'\u2915'")
#>           
#> 1 <U+2915>
```

<sup>Created on 2022-11-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value
#>  version  R version 4.1.2 (2021-11-01)
#>  os       Windows 10 x64 (build 19044)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  Dutch_Netherlands.1252
#>  ctype    Dutch_Netherlands.1252
#>  tz       Europe/Berlin
#>  date     2022-11-04
#>  pandoc   2.19.2 @ H:/Programs/RStudio/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> - Packages -------------------------------------------------------------------
#>  package     * version date (UTC) lib source
[...]

PR #863, opened 2024-11-30
OdbcResult: Ability to explicitely describe parameters

detule:
Hi @simonpcouch @hadley 

This is a draft/RFC of a feature that would allow (power) users to explicitly describe parameter attributes when submitting parametrized queries.

The details are in https://github.com/r-dbi/odbc/issues/518 and this PR would address the issue there.  In summary:
* We do a very good job of inferring the parameter metadata, but at the end of the day it's only an informed guess.
* In some cases the user may wish to override those inferred values.  Linked issue above is one of those.
* Offer them the ability to do that.  This type of feature is not uncommon in other ODBC programming interfaces.

Some notes about the PR:
* Added an ODBC_TYPE helper method that would help with the user not having to guess what integer corresponds to SQL_WVARCHAR, for example.
* The example is in ?OdbcResult.
* I have a few items below as TODOs, including offering the usage of this parameter more broadly.

TODO:
- [ ] Rename ODBC_TYPE to SQL_TYPE
- [ ] Permeate parameter_description to `dbSendQuery`, `dbGetQuery`
- [ ] Add unit tests


R/dbi-result.R
```diff
@@ -126,17 +126,54 @@ setMethod("dbGetRowsAffected", "OdbcResult",
...
+#' # column with SQL server
+#' DBI::dbRemoveTable(conn, "#tmp")
+#' dbExecute(conn, "CREATE TABLE #tmp (col1 VARCHAR(50) COLLATE Latin1_General_100_CI_AI_SC_UTF8);")
+#' res <- dbSendQuery(conn, "INSERT INTO #tmp SELECT ? COLLATE Latin1_General_100_CI_AI_SC_UTF8")
+#' description <- data.frame(param_index = 1, data_type = ODBC_TYPE("WVARCHAR"),
+#'   column_size = 5, decimal_digits = NA_integer_)
+#' dbBind(res, params = list("\u2915"), param_description = description)
+#' dbClearResult(res)
+#' DBI::dbReadTable(conn, "#tmp")
+#' }
```
simonpcouch, line 153:
```suggestion
#' @examplesIf FALSE
#' # Writing UNICODE into a VARCHAR column with SQL server
#' dbRemoveTable(conn, "#tmp")
#' 
#' dbExecute(
#'   conn, 
#'   "CREATE TABLE #tmp (col1 VARCHAR(50) COLLATE Latin1_General_100_CI_AI_SC_UTF8);"
#' )
#' 
#' res <- dbSendQuery(
#'   conn, 
#'   "INSERT INTO #tmp SELECT ? COLLATE Latin1_General_100_CI_AI_SC_UTF8"
#' )
#' 
#' description <- data.frame(
#'   param_index = 1, 
#'   data_type = ODBC_TYPE("WVARCHAR"),
#'   column_size = 5, 
#'   decimal_digits = NA_integer_
#' )
#' 
#' dbBind(res, params = list("\u2915"), param_description = description)
#' dbClearResult(res)
#' 
#' dbReadTable(conn, "#tmp")
```

Some style edits I'd make:

* `examples` with `\dontrun{}` -> `examplesIf FALSE` so that users don't see `## Not run` tags and CRAN definitely won't try to run the examples
* I _believe_ the `DBI::` namespacing here may not be needed?
* Line-breaking more liberally via [codegrip](https://github.com/lionel-/codegrip)

R/dbi-result.R
```diff
@@ -126,17 +126,54 @@ setMethod("dbGetRowsAffected", "OdbcResult",
...
 )
 
+#' @param param_description A data.frame with per-parameter attribute
+#' overrides.  Argument is optional; if used it must have columns:
+#' * param_index Index of parameter in query ( beginning with 1 ).
+#' * data_type Integer corresponding to the parameter SQL Data Type.
+#'   See \code{\link{ODBC_TYPE}}.
+#' * column_size Size of parameter.
+#' * decimal_digits Either precision or the scale of the parameter
+#'   depending on type.
```
simonpcouch, line 136:
```suggestion
#' @param param_description A data frame containing per-parameter attribute overrides.
#'   Optional argument that, if provided, must contain the following columns:
#'   \describe{
#'     \item{param_index}{Index of parameter in query (beginning with 1).}
#'     \item{data_type}{Integer corresponding to the parameter SQL Data Type.
#'       See [ODBC_TYPE()].
#'     \item{column_size}{Size of parameter.}
#'     \item{decimal_digits}{Either precision or the scale of the parameter
#'       depending on type.}
#'   }
```

* Transitions to `\describe` for data frame columns
* Links to `ODBC_TYPE()` with roxygen shorthand
* Refers to "data frame" rather than `data.frame` since tibbles and other data.frame subclasses are fine

R/dbi-result.R
```diff
@@ -126,17 +126,54 @@ setMethod("dbGetRowsAffected", "OdbcResult",
...
+#' DBI::dbReadTable(conn, "#tmp")
+#' }
 #' @rdname OdbcResult
 #' @inheritParams DBI::dbBind
 #' @inheritParams DBI-tables
 #' @export
 setMethod("dbBind", "OdbcResult",
-  function(res, params, ..., batch_rows = getOption("odbc.batch_rows", NA)) {
+  function(res, params, ...,
+           param_description = data.frame(),
```
simonpcouch, line 160:
```suggestion
           param_description = NULL,
```

`NULL` indicates a bit more clearly to me that "this is the default that won't actually be used if you don't supply anything"

R/dbi-result.R
```diff
@@ -126,17 +126,54 @@ setMethod("dbGetRowsAffected", "OdbcResult",
...
+           batch_rows = getOption("odbc.batch_rows", NA)) {
     params <- as.list(params)
     if (length(params) == 0) {
       return(invisible(res))
     }
 
+    if (nrow(param_description)) {
```
simonpcouch, line 167:
```suggestion
    if (!is.null(param_description)) {
```

Aligning with above.

R/dbi-result.R
```diff
@@ -126,17 +126,54 @@ setMethod("dbGetRowsAffected", "OdbcResult",
...
+      if (!all(c("param_index", "data_type", "column_size", "decimal_digits")
+        %in% colnames(param_description))) {
+        cli::cli_abort(
+          "param_description data.frame does not have necessary columns."
+        )
+      }
```
simonpcouch, line 173:
```suggestion
      check_data_frame(param_description)
      needed_columns <- c("param_index", "data_type", "column_size", "decimal_digits")
      if (!all(needed_columns %in% colnames(param_description))) {
        cli::cli_abort(
          "{.arg param_description} must have columns {.field {needed_columns}}, but 
           doesn't have column{?s} 
           {.field {.or {needed_columns[needed_columns %in% colnames(param_description)]}}}."
        )
      }
```

`check_data_frame()` will first ensure that the thing is a data frame and provide an informative message if not before we do the finer-grain check for needed columns.

R/RcppExports.R
```diff
@@ -77,6 +77,24 @@ bigint_mappings <- function() {
...
+#'
+#' @seealso <https://github.com/microsoft/ODBC-Specification/blob/master/Windows/inc/sql.h>
+#' @rdname odbcDataType.
+#' @examples
+#' \dontrun{
+#' library(odbc)
+#' ODBC_TYPE("LONGVARCHAR")
+#' }
+#' @export
+ODBC_TYPE <- function(type) {
```
simonpcouch, line 94: I _think_, just to situate more comfily in this package's namespace, this may be better formatted as:

```suggestion
odbcType <- function(type) {
```

simonpcouch (commented):
Seems very much useful for power users! I don't have much to offer in terms of critique on the `src/` side, but left some notes on the R interface.

PR #870, opened 2024-12-09
prevent application of inline markup within rethrown messages

### LINKED ISSUE #859: A failing query using `DBI::dbGetQuery()` yields a useless error message
When a query fails in `DBI::dbGetQuery()` a useless error message is shown instead of quoting the error message that is issued by the remote data base:

``` R
Error in `fun(..., .envir = .envir)`:
! Could not parse cli `{}` expression: `100.96.8.10:55372 -…`.
Caused by error in `parse(text = code, keep.source = FALSE)`:
! <text>:1:7: unerwartete numerische Konstante
1: 100.96.8
          ^
Type .Last.error to see the more details.
```

Since the error depends on the remote system (HANA DB) and a failing query the issue is hard to reproduce, so I cannot share a useful reprex.

It looks like some problems when parsing the remote error message (see https://rlang.r-lib.org/reference/topic-condition-formatting.html#transitioning-from-abort-to-cli-abort-) in this code piece:
https://github.com/r-dbi/odbc/blob/bd6686766a503dc69040d19e99c03a9ad3a73967/R/utils.R#L162-L178

### Stack trace
``` R
11. DBI::dbGetQuery(remote_con(x), get_sql(x, cte, header, hints))
12. DBI::dbGetQuery(remote_con(x), get_sql(x, cte, header, hints))
13. local .local(conn, statement, ...)
14. DBI::dbSendQuery(conn, statement, params = params, immediate = immediate, …
15. odbc::dbSendQuery(conn, statement, params = params, immediate = immediate, …
16. local .local(conn, statement, ...)
17. odbc:::OdbcResult(connection = conn, statement = statement, params = params, …
18. odbc:::new_result(p = connection@ptr, sql = statement, immediate = immediate)
19. (function (msg, call = trace_back()$call[[1]]) …
20. cli::cli_abort(c(`!` = "ODBC failed with error {res$cnd_context_code} from \\\n   …
21. cli:::vcapply(message, format_inline, .envir = .envir)
22. base::vapply(X, FUN, FUN.VALUE = character(1), ..., USE.NAMES = USE.NAMES)
23. local FUN(X[[i]], ...)
24. cli::cli_fmt(fun(..., .envir = .envir), collapse = collapse, strip_newline = TRUE)
25. cli:::cli__rec(expr)
26. local fun(..., .envir = .envir)
27. cli:::cli__message("inline_text", list(text = glue_cmd(..., .envir = .envir, …
28. "id" %in% names(args)
29. cli:::glue_cmd(..., .envir = .envir, .call = sys.call(), .trim = FALSE)
30. cli:::glue(str, .envir = .envir, .transformer = transformer, .cli = TRUE, …
31. (function (expr) …
32. .transformer(expr, .envi…
33. local .transformer(expr, .envi…
34. parse(text = code, keep.…
35. cli:::chain_error(expr, err, s…
36. | base::withCallingHandlers({ …
37. base::parse(text = code, keep.…
38. base::.handleSimpleError(funct…
39. | local h(simpleError(msg, call))
40. | cli:::throw_error(err, parent …
```

simonpcouch:
Closes #859. 

Safeguards against mistakenly applying inline markup to driver messages when rethrowing based on the [cli documentation](https://cli.r-lib.org/reference/inline-markup.html?q=escape#escaping-and-).


R/utils.R
```diff
@@ -162,7 +162,7 @@ random_name <- function(prefix = "") {
 rethrow_database_error <- function(msg, call = trace_back()$call[[1]]) {
   tryCatch(
     res <- parse_database_error(msg),
-    error = function(e) cli::cli_abort(msg, call = call)
+    error = function(e) cli::cli_abort("{msg}", call = call)
```
simonpcouch, line 165: My initial guess re: the OP was that this had been the source of the failure (i.e. `parse_database_error()` failed and then the error handling mishandled the message). Turns out it wasn't, but I brought this [in line with cli's documentation](https://cli.r-lib.org/reference/inline-markup.html?q={msg}#escaping-and-) + added a test in the process that felt worth keeping around.

simonpcouch (commented):
This is ready for review, @hadley!

hadley (approved):

Issue #869, opened 2024-12-09
failures re: independent column name encodings


simonpcouch:
In [recent runs on DBs + Posit Pro Drivers (Linux) / database](https://github.com/r-dbi/odbc/actions/runs/12239347859/job/34139569935?pr=868), seeing:

```
── Failed tests ────────────────────────────────────────────────────────────────
Failure ('test-driver-sql-server.R:355:3'): independent encoding of column entries and names (#834)
`df` (`actual`) not identical to `res` (`expected`).

`names(actual)`:   "var_char_col" "bøvs"
`names(expected)`: "var_char_col" "b.vs"

`actual$var_char_col`:   "kanin" "ræven" "ålens" "ørred"
`expected$var_char_col`: "kanin" "r?ven" "?lens" "?rred"

`actual$bøvs` is a double vector (1, 1, 1, 1)
`expected$bøvs` is absent

`actual$b.vs` is absent
`expected$b.vs` is a double vector (1, 1, 1, 1)

[ FAIL 1 | WARN 0 | SKIP 408 | PASS 10246 ]
```

Related to #834—seems to be with Pro Drivers only?

Note that actual/expected are flipped in this test.

PR #871, opened 2024-12-15
ci: pro: test update

detule:
Skip SQL server encoding test since OEM driver does not provide for a way to disable "AutoTranslate" feature.

Fixes https://github.com/r-dbi/odbc/issues/869


detule:
Hey Simon:

This should hopefully fix the issue error with PRO CI workflow.  I don't think I have the ability to run that pipeline - if you get a chance to run it manually, we can verify if the test is skipped appropriately.

detule:
Looks like either I broke something unintentionally, or there is an issue with the runner.  I'll look into it and ping back when ready.

detule:
Just adding some running commentary: I think our runner started using 24.04 ( was using Ubuntu 22.04 ) - i suspect this is the culprit of the new issue.

simonpcouch (approved):
Just locally checked the Pro driver pipeline and saw no issues. Thank you!

CI failures addressed in #876.

Apologies for the delay—spent a good bit of time away for the holidays.🎄

PR #872, opened 2024-12-16
ci: peg ubuntu image

detule:
This is likely temporary, but would allow business as usual while we resolve some of the issues with the new runner image.

* First note the announcement [here](https://github.com/actions/runner-images/issues/10636).
* Some issues I see in the "DB" pipeline with the new image - will tackle subsequently:

1. MySQL driver installation fails.  This could be due to a newer `cmake` or the newer driver manager.

```
 CMake Error at CMakeLists.txt:263 (MESSAGE):
  Driver Manager was not found
```
2. SQL Server tests fail with
```
ODBC failed with error 00000 from [Microsoft][ODBC Driver 17 for SQL Server][OpenSSL library could not be loaded, make sure OpenSSL 1.0 or 1.1 is installed]
```
3. Oracle tests fail with
```
[unixODBC][Driver Manager]Can't open lib '/opt/oracle/instantclient_21_12/libsqora.so.21.1' : file not found ")
```
( though as best as I can tell, the driver is actually installed ).

For reference, something similar happened during the [last "latest" migration](https://github.com/microsoft/linux-package-repositories/issues/39).  Unclear if we are running into the same issue again.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -30,12 +30,12 @@ jobs:
           # use 4.1 to check with rtools40's older compiler
           - {os: windows-latest, r: '4.1'}
 
-          - {os: ubuntu-latest,  r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-latest,  r: 'release'}
-          - {os: ubuntu-latest,  r: 'oldrel-1'}
-          - {os: ubuntu-latest,  r: 'oldrel-2'}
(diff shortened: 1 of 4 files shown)


detule:
I just checked, this is still needed ( not fixed in ubuntu 24 image ).   I suggest we go with this in the short term.  In the meantime I'll start researching possible solutions that will take us back to being able to use `:latest`.

detule:
Closing in favor of https://github.com/r-dbi/odbc/pull/876

PR #875, opened 2024-12-29
nanodbc: ensure_pdata: Fix memory leak on error

detule:
Upstream: https://github.com/nanodbc/nanodbc/pull/414

Hi all: this back-ports a fix for a possible memory leak that was addressed upstream.


simonpcouch (approved):
Stellar, thanks!

hadley (approved):

PR #876, opened 2024-12-29
CI: Update workflows for ubuntu 24.04 ( now latest )

detule:
See: https://github.com/r-dbi/odbc/pull/872

Fixed:

- [x] MySQL driver installation failure
- [x] Oracle driver installation failure
- [x] SQL Server driver installation faliure

Outstanding / TODO:

- [x] SQL Server connection failure with new driver ( 18 )
- [x] ODBCSYSINI test failure.

Some notable changes in the Linux/SQL Server testing workflow: We are now using SQL Server OEM Driver 18 ( were using 17 ), and are now testing against Microsoft SQL Server 19 ( were testing against SQL Server 17 ).  These updates were necessary since Ubuntu 24.04 carries the openssl 3.0 libs, which seem to be incompatible with the v17 driver.  The update to the test server/database ( 17 -> 19 ) was necessary to accommodate the SSL lib updates as well.


simonpcouch (approved):
Oof, this seems like it would have been a painful one to sort through but I'm very grateful you did. Changes look good to me—many thanks.

hadley (approved):

PR #881, opened 2025-01-10
Add `/usr/lib/x86_64-linux-gnu` to the list of common `libodbcinst` dirs

### LINKED ISSUE #880: `locate_install_unixodbc()` does not work on a fresh Ubuntu 22.04 install
On a fresh Ubuntu 22.04 box @tnederlof and I see the following error when trying to use `dbConnect()`:

```
Unable to locate the unixODBC driver manager
```

On the CRAN release we see no such error, and the driver manager most certainly *is* installed.

We tracked this down to `locate_install_unixodbc()` returning no results, and as of #825 we verify that in `dbConnect()`. I believe this is because it assumes `pkg-config` is available, which it is not on a fresh Ubuntu machine: https://github.com/r-dbi/odbc/blob/4ecd5099276e90e2840ff0100bdbed9edcf6f4e2/R/utils.R#L332-L335

There is also a list of "common" paths: https://github.com/r-dbi/odbc/blob/4ecd5099276e90e2840ff0100bdbed9edcf6f4e2/R/utils.R#L341-L348

But this does not include the path `libodbcinst.so` is actually installed to, which is under `/usr/lib/x86_64-linux-gnu`.

I believe this can be fixed by adding that path to the common list; it should fix issues on many Debian-derived distros. We could also advise that folks install `pkg-confg`, but that seems wrong to me.

atheriel:
This fixes `locate_install_unixodbc()` on fresh Ubuntu 22.04 installs, which don't otherwise include `pkg-confg` or use the existing common paths.

Fixes #880.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -343,6 +343,7 @@ locate_install_unixodbc <- function() {
     "/usr/local",
     "/lib",
     "/usr/local/lib",
+    "/usr/lib/x86_64-linux-gnu",
     "/opt/homebrew/lib",
     "/opt/homebrew/opt/unixodbc/lib"
   )


simonpcouch (approved):
Stellar, thank you🏄

Your call on whether to add a NEWS entry for this one: unlikely to affect existing users but certainly an improvement for newcomers.

atheriel:
My inclination is to say "no `NEWS` item", because I think this only affects people using the unreleased version.

Issue #793, opened 2024-04-23
Determine DATETIME2 precision based on parameter description


detule:
Related to: https://github.com/r-dbi/odbc/issues/790 https://github.com/r-dbi/odbc/issues/208 https://github.com/r-dbi/odbc/issues/130

When writing to a datetime/datetime2 target, we use a fixed precision of three decimals to avoid needing to know the user-configurable precision of the datetime2 field.

However, when writing, we attempt to discover the target table column descriptions and pass these down to `nanodbc` via `statement::describe_parameters`.   We do this to avoid (nanodbc) needing to call `SQLDescribeParams` for each of the individual columns/parameters when binding the data buffers.

If we already have the column descriptions, we could have `nanodbc::statement` expose these - see for example `nanodbc::statement::parameter_size(int)` that already does something similar.  Then we can use this API endpoint in `odbc_result` to maybe deduce the parameter precision.

PR #885, opened 2025-01-25
Fixup/sqlserver datetime2

detule:
Closes https://github.com/r-dbi/odbc/issues/793

This is a long standing issue where we were forcing a precision of 3 for all DATETIME types despite some ( for example DATETIME2 ) having potentially greater precision.

Requires a backport from nanodbc.

### DIFF
--- a/src/nanodbc/nanodbc.cpp
+++ b/src/nanodbc/nanodbc.cpp
@@ -1907,7 +1907,7 @@ public:
         describe_parameters(param_index);
         const SQLSMALLINT& param_type = param_descr_data_.at(param_index).type_;
         NANODBC_ASSERT(param_type < static_cast<SQLULEN>(std::numeric_limits<short>::max()));
-        return static_cast<unsigned long>(param_type);
+        return static_cast<short>(param_type);
     }
 
     static SQLSMALLINT param_type_from_direction(param_direction direction)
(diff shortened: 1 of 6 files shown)


src/nanodbc/nanodbc.cpp
```diff
@@ -1870,35 +1870,44 @@ class statement::statement_impl
...
+    {
+        if (param_descr_data_.count(param_index))
+        {
+            return static_cast<short>(param_descr_data_.at(param_index).type_);
+        }
+
+        describe_parameters(param_index);
+        const SQLSMALLINT& param_type = param_descr_data_.at(param_index).type_;
+        NANODBC_ASSERT(param_type < static_cast<SQLULEN>(std::numeric_limits<short>::max()));
+        return static_cast<unsigned long>(param_type);
```
simonpcouch, line 1910: This may just be my C++ naïveté—apologies if so. From what I can tell, the function declares it returns a short but can actually return an unsigned long. Is this ever an issue?

simonpcouch (approved):
Don't have the skillset to review `src/` changes confidently but see no reason not to approve from my perspective. Thanks!

src/nanodbc/nanodbc.cpp
detule: Good call and thank you for taking a look!

The compiler will clean that up and cast back to short, but its poor form on my part.  Fixed!

PR #887, opened 2025-01-28
sql server: Roundtrip of hms data

detule:
Last failing column from the `test_roundtrip` suite for SQL Server.


simonpcouch (approved):
😎😎😎

PR #894, opened 2025-02-10
Overhaul viewer-based credentials for Databricks & Snowflake

atheriel:
This commit takes a second stab at the viewer-based credential support added in #853.

It contains two major changes:

1. Users no longer need to pass a `session` argument. We figure it out internally instead. This allows users to deploy apps to Connect with no code changes.

2. The underlying code to get these credentials is now hosted in the `connectcreds` package, which also provides a number of nice mocking features we can use in `odbc`'s unit tests. The `connectcreds` implementation is based on the code that this commit removes, though it has a number of improvements as well, so this should be a net win, too.

Unit tests and documentation is included.


R/driver-databricks.R
```diff
@@ -51,11 +51,11 @@ NULL
 #'
```
simonpcouch, line 51: Can't suggest on the line directly, but since connectcreds is only in Suggests, let's transition `@examples` to `@examplesIf rlang::is_installed("connectcreds")`.

R/driver-snowflake.R
```diff
@@ -146,8 +147,9 @@ setMethod("odbcDataType", "Snowflake",
 #'
 #' # Use credentials from the viewer (when possible) in a Shiny app
 #' # deployed to Posit Connect.
+#' library(connectcreds)
```
simonpcouch, line 150: Same conditional example run comment

simonpcouch (approved):
Stellar. Thanks for making this happen!

For the GitHub archeologists: that `session` argument never made it to CRAN so the removal of that argument only affects users of the dev version of the package.

R/driver-databricks.R
simonpcouch: Ah, you're right—my suggested change would be redundant with `\dontrun{}`. Feel free to ignore.

PR #899, opened 2025-03-01
utils: locate_unixodbc: account for statically built unixodbc (CRAN)

detule:
Attempt to fix errors reported on CRAN ( where the build machine is using a statically buit `unixodbc` ).
Used a simple "stop" rather than something with better formatting since that error does not propagate to the user.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -361,7 +361,9 @@ locate_install_unixodbc <- function() {
     "/usr/lib/x86_64-linux-gnu",
     "/opt/homebrew/lib",
     "/opt/homebrew/etc",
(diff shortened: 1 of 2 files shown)


simonpcouch (approved):
Very grateful you looked into this CRAN issue.

`devtools::check_mac_release()` [fails with CRAN odbc](https://mac.r-project.org/macbuilder/results/1741012003-36ea9a6fe3000f1b/) and [passes with this PR](https://mac.r-project.org/macbuilder/results/1741011916-7ffcb12a8227c085/). Huzzah!

PR #903, opened 2025-03-11
v1.6.1 release candidate

### LINKED ISSUE #902: Release odbc 1.6.1
Prepare for release:

* [x] `git pull`
* [x] Check [current CRAN check results](https://cran.rstudio.org/web/checks/check_results_odbc.html)
* [x] [Polish NEWS](https://style.tidyverse.org/news.html#news-release)
* [x] `urlchecker::url_check()`
* [x] `devtools::build_readme()`
* [x] `devtools::check(remote = TRUE, manual = TRUE)`
* [x] `devtools::check_win_devel()`
* [x] `revdepcheck::cloud_check()`
* [x] Update `cran-comments.md`
* [x] `git push`

Submit to CRAN:

* [x] `usethis::use_version('patch')`
* [x] `devtools::submit_cran()`
* [x] Approve email

Wait for CRAN...

* [ ] Accepted :tada:
* [ ] `usethis::use_github_release()`
* [ ] `usethis::use_dev_version(push = TRUE)`

simonpcouch:
The CRAN check issue was on macos systems—that's [now resolved](https://mac.r-project.org/macbuilder/results/1741708254-a7d5d8f39dfe4db1/).

Closes #902.

EDIT: FYI that I'll be OOO the 12th-14th. The submission deadline is Friday the 14th.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: odbc
 Title: Connect to ODBC Compatible Databases (using the DBI Interface)
-Version: 1.6.0.9000
+Version: 1.6.1
 Authors@R: c(
     person("Jim", "Hester", role = "aut"),
     person("Hadley", "Wickham", , "hadley@posit.co", role = c("aut", "cre")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # odbc (development version)
 
+* odbc will now automatically find statically built installations of
+  unixodbc (#899).
+
 # odbc 1.6.0
 
 ## New features
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# odbc (development version)
+# odbc 1.6.1
 
 * odbc will now automatically find statically built installations of
   unixodbc (#899).
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -1,6 +1,6 @@
 ## R CMD check results
 
-Recent MacOS builds trigger the NOTE "installed size is 10.8Mb." MacOS builds include all debug symbols and CRAN policy doesn't permit building without them.
+Recent MacOS builds trigger the NOTE "installed size is 11.0Mb." MacOS builds include all debug symbols and CRAN policy doesn't permit building without them.
 
 ## revdepcheck results
 
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -1,3 +1,5 @@
+Trying again - I think we should now be passing on M1mac. Sorry for the multiple attempts here but something is configured in an unexpected way on that system and we're having a hard time replicating locally.
+
 ## R CMD check results
 
 Recent MacOS builds trigger the NOTE "installed size is 11.0Mb." MacOS builds include all debug symbols and CRAN policy doesn't permit building without them.
@@ -8,4 +10,3 @@ We checked 35 reverse dependencies, comparing R CMD check results across CRAN an
 
  * We saw 0 new problems
  * We failed to check 0 packages
-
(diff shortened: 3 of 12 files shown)


simonpcouch:
Pro Driver failures existed with the odbc submission currently on CRAN without issue, so ignoring.

@hadley, #902 is up-to-date with the current status of the release process; should be ready to send out!

simonpcouch:
The failure we see is:

```
  ── Failure ('test-utils.R:180:3'): locate_install_unixodbc() returns reasonable values ──
  file.exists(res[1]) is not TRUE
  
  `actual`:   FALSE
  `expected`: TRUE 
```

#899 did not resolve it, apparently (though note that it _did_ resolve the same failure on mac-builder).

`locate_install_unixodbc()` either returns:

1) `c(system_safely("odbc_config --lib-prefix"), system_safely("pkg-config --variable=libdir odbc"))` _if_ that vector is length > 0, or
2)  Results from `list.files(..., full.names = TRUE)`

Call the results from either of the above `res`. In the failing test, `file.exists(res[1])` is `FALSE`. Notably, it's not a subset error, so `res` is at least length 1. This makes me believe that `locate_install_unixodbc()` likely doesn't return from 2) on CRAN, as I can't think of an explanation for `file.exists(list.files(..., full.names = TRUE)[1])` being `FALSE` other than some sort of tempfile local to `list.files()` or some path normalization wonk. So I'm inclined to believe that `locate_install_unixodbc()` returns from 1), and CRAN has set odbc_config cflags in such a way that they point to files that don't exist. Opting to just skip the test on CRAN.

simonpcouch:
Was about to replicate the `error_install_unixodbc()` errors on an M1 Mac via removing the homebrew unixodbc install and its reverse dependencies. There are now quite a few new test skips here.

detule:
> Was able to replicate the `error_install_unixodbc()` errors on an M1 Mac via removing the homebrew unixodbc install and its reverse dependencies. There are now quite a few new test skips here.

Wait, we think there is no `unixodbc` installed at all on the CRAN machine, despite it being part of the recipes here: https://github.com/R-macos/recipes ?

simonpcouch:
My hypothesis from https://github.com/r-dbi/odbc/pull/903#issuecomment-2730124049 is that `c(system_safely("odbc_config --lib-prefix"), system_safely("pkg-config --variable=libdir odbc"))` is length > 0, though the files that one of `odbc_config` or `pkg-config` is pointing us to doesn't actually exist, making for a unixodbc installation that's not actually usable. Very much open to the idea that I'm wrong or that we ought to instead troubleshoot the recipe!

detule:
Thanks Simon!

Ugh, i really wish there was a simpler mechanism to test releases against their build environment.

Apologies for asking, I should know, but how are you getting feedback that the tests error out in their build environment?
Provided there is a way to throw tar balls at it, and iterate on this, I can take a look this weekend.  I am curious why/where those missing files are.

But generally don't think this should hold the release / think it's OK to skip these tests on CRAN.

simonpcouch:
Very much agreed.

> Apologies for asking, I should know, but how are you getting feedback that the tests error out in their build environment?

Nope, not a silly question at all! The mechanism here is that odbc is currently being flagged for manual review, and a CRAN team member emailed Hadley back with pasted-in testthat output from an M1 system ("M1mac", the one [linked here](https://cran.r-project.org/web/checks/check_results_odbc.html)) in response to our submission a couple days ago. This seems to be a different system than [Mac-builder](https://mac.r-project.org/macbuilder/submit.html)—from what I understand, Mac-builder is consistent with the recipes you've linked. I'm unfortunately not aware of a way to throw tarballs at the thing without manual intervention from CRAN folks.

hadley:
I've submitted it again 🤞

Issue #901, opened 2025-03-05
Calling `DBI::dbConnect(odbc::databricks())` a second time fails with `Error in driver_config$path ` on MacOS


benesom:
Establishing a odbc connection works as expect for
```
args <- list(
  odbc::databricks(),
  auth_accesstoken = token,
  workspace = workspace,
  authMech = 11,
  auth_flow = 0,
  HTTPPath = httppath
)
conn <- do.call(DBI::dbConnect, args)
```
```
> conn
<OdbcConnection> Spark SQL
  Database: foundata_investigations_bronze_prd
  Spark SQL Version: 3.5.0
```
However, trying to "re-connect" to the cluster fails
```
> do.call(DBI::dbConnect, args)
Error in driver_config$path : $ operator is invalid for atomic vectors
```

The traceback points at `odbc:::configure_simba`, which was introduced in odbc v.1.6.0
```
5: configure_simba(spark_simba_config(args$driver), action = "modify")
4: .local(drv, ...)
3: (new("standardGeneric", .Data = function (drv, ...) 
   {
       ans <- standardGeneric("dbConnect")
       .valueClassTest(ans, "DBIConnection", "dbConnect")
   }, generic = structure("dbConnect", package = "DBI"), package = "DBI", 
       group = list(), valueClass = "DBIConnection", signature = "drv", 
       default = NULL, skeleton = (function (drv, ...) 
       stop(gettextf("invalid call in method dispatch to '%s' (no default method)", 
           "dbConnect"), domain = NA))(drv, ...)))(new("DatabricksOdbcDriver", ), 
       auth_accesstoken = "xxxx", 
       workspace = "xxxx", 
       authMech = 11, auth_flow = 0, HTTPPath = "xxx")
2: (new("standardGeneric", .Data = function (drv, ...) 
   {
       ans <- standardGeneric("dbConnect")
       .valueClassTest(ans, "DBIConnection", "dbConnect")
   }, generic = structure("dbConnect", package = "DBI"), package = "DBI", 
       group = list(), valueClass = "DBIConnection", signature = "drv", 
       default = NULL, skeleton = (function (drv, ...) 
       stop(gettextf("invalid call in method dispatch to '%s' (no default method)", 
           "dbConnect"), domain = NA))(drv, ...)))(new("DatabricksOdbcDriver", ), 
       auth_accesstoken = "xxx", 
       workspace = "xxxx", 
       authMech = 11, auth_flow = 0, HTTPPath = "xxx")
[...]

PR #911, opened 2025-04-05
databricks/snowflake: fixup *simba_config so that return value is sam…

detule:
I believe this will close / address https://github.com/r-dbi/odbc/issues/901


simonpcouch (approved):
Ah, yup—looking good!

Feels worth a NEWS entry.

PR #915, opened 2025-04-27
unit test fixup

detule:
Update unit tests to keep pace with #911 

Apologies I didn't catch this as part of the original PR.


simonpcouch (approved):
Great, thank you!

PR #918, opened 2025-05-10
Add support for SQL Server DATETIMEOFFSET data type

detule:
Hi:

This PR closes a longstanding TODO to better support timezones for SQL Server users.

Currently, we allow the user to specify a *fixed* timezone using the `connection.timezone` parameter.  However, some back ends ( most notably, `MSSQL`) support much more granular time+zone offset expressions, for example:

```
SELECT CURRENT_TIMESTAMP AT TIME ZONE 'Central European Standard Time'
```

which is of a data type that is not currently handled by `nanodbc` and `package:odbc`.  In particular, data would currently come back to the [R] user as a string `2025-05-10 20:07:55.410 +02:00`.  Users would then need to parse the string, manually extract the time zone, and convert at some overhead.

This PR adds a more native integration for this data type.  For example:
```
  > dbGetQuery(conn, "SELECT CAST(tz AS VARCHAR) tz_char, tz FROM (SELECT CURRENT_TIMESTAMP AT TIME ZONE 'Central European Standard Time' AS tz) TBL;")
                           tz_char                  tz
  1 2025-05-10 20:11:28.640 +02:00 2025-05-10 22:11:28
```

PR is broken up in three commits:
1. Changes to `nanodbc`.  A [FR](https://github.com/nanodbc/nanodbc/issues/379) already exists there.  I'll attempt to upstream this there.
2. Additions (no changes) to `cctz`.  These updates marry nicely against existing upstream features.  Where I made departures, I marked using `package::odbc` comments.
3. Updates to `package:odbc` business logic.

TODO:

- [x] Add news item

ping @r2evans

### DIFF
--- a/tests/testthat/test-driver-sql-server.R
+++ b/tests/testthat/test-driver-sql-server.R
@@ -385,7 +385,7 @@ test_that("DATETIMEOFFSET", {
     field.types = list("tz_char" = "VARCHAR(50)", "tz" = "DATETIMEOFFSET"), overwrite = TRUE)
   res <- DBI::dbReadTable(con, tbl)
   expect_s3_class(res[[2]], "POSIXct")
-  expect_equal(as.double(res[[2]][1] - as.POSIXct(res[[1]][1]), units="hours"), 2);
+  expect_equal(as.double(res[[2]][1] - as.POSIXct(res[[1]][1]), units = "hours"), 2)
 })
 
 test_that("package:odbc roundtrip test", {
--- a/tests/testthat/test-driver-sql-server.R
+++ b/tests/testthat/test-driver-sql-server.R
@@ -384,7 +384,7 @@ test_that("DATETIMEOFFSET", {
   tbl <- local_table(con, "test_datetimeoffset", df,
     field.types = list("tz_char" = "VARCHAR(50)", "tz" = "DATETIMEOFFSET"), overwrite = TRUE)
   res <- DBI::dbReadTable(con, tbl)
-  expect_true("POSIXct" %in% class(res[[2]]))
+  expect_s3_class(res[[2]], "POSIXct")
   expect_equal(as.double(res[[2]][1] - as.POSIXct(res[[1]][1]), units="hours"), 2);
 })
 
--- a/tests/testthat/test-driver-sql-server.R
+++ b/tests/testthat/test-driver-sql-server.R
@@ -376,6 +376,18 @@ test_that("DATETIME2 precision (#790)", {
   expect_equal(as.POSIXlt(df[[2]])$sec, as.POSIXlt(res[[2]])$sec, tolerance = 1E-7)
 })
 
+test_that("DATETIMEOFFSET", {
+  con <- test_con("SQLSERVER")
+
+  df <- data.frame(tz_char = rep("2025-05-10 19:35:03.123 +02:00", 3), tz = rep("2025-05-10 19:35:03.123 +02:00", 3))
+
+  tbl <- local_table(con, "test_datetimeoffset", df,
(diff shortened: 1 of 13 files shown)


r2evans:
Thank you @detule! This is indeed a long-running FR ...

Just to make sure ... you aren't requiring users to use the `'Central European Standard Time'` TZ format, is that right? While I guess that some users would really prefer to know the TZ it is stored with, I don't care, I just need it to be correct ... I often change the TZs locally anyway in an `O(1)` attribute change.

> Users would then need to parse the string, manually extract the time zone, and convert at some overhead.

"Some overhead" is funny (for me): I regularly pull 5+M rows with 2-4 `DATETIMEOFFSET` fields, which adds up to a very palpable experience. It'll be very nice when we don't have to do this (and R's global string hash will get to relax some, too).

detule:
> Just to make sure ... you aren't requiring users to use the `'Central European Standard Time'` TZ format, is that right?

Right; that was just a way to add a non trivial offset.

Sweet - excellent that this is still relevant; we'll find a way to fold it in the next few weeks I suspect.  Thanks again for the patience.

r2evans:
> > Just to make sure ... you aren't requiring users to use the `'Central European Standard Time'` TZ format, is that right?
> 
> Right; that was just a way to add a non trivial offset.
> 
> Sweet - excellent that this is still relevant; we'll find a way to fold it in the next few weeks I suspect. Thanks again for the patience.

I work in all three OSes (windows, macos, and linux) and since I haven't found a reliable conversion from _that_ "crap" format to more normal "country/region" formats (for access no matter the caller OS), I have a table that maps all that I could find and unambiguously assign ... and the rest I assign as close as i can get. :-(

Thanks1

tests/testthat/test-driver-sql-server.R
```diff
@@ -376,6 +376,18 @@ test_that("DATETIME2 precision (#790)", {
...
+test_that("DATETIMEOFFSET", {
+  con <- test_con("SQLSERVER")
+
+  df <- data.frame(tz_char = rep("2025-05-10 19:35:03.123 +02:00", 3), tz = rep("2025-05-10 19:35:03.123 +02:00", 3))
+
+  tbl <- local_table(con, "test_datetimeoffset", df,
+    field.types = list("tz_char" = "VARCHAR(50)", "tz" = "DATETIMEOFFSET"), overwrite = TRUE)
+  res <- DBI::dbReadTable(con, tbl)
+  expect_true("POSIXct" %in% class(res[[2]]))
+  expect_equal(as.double(res[[2]][1] - as.POSIXct(res[[1]][1]), units="hours"), 2);
```
simonpcouch, line 388:
```suggestion
  expect_equal(as.double(res[[2]][1] - as.POSIXct(res[[1]][1]), units = "hours"), 2)
```

tests/testthat/test-driver-sql-server.R
```diff
@@ -376,6 +376,18 @@ test_that("DATETIME2 precision (#790)", {
...
 
+test_that("DATETIMEOFFSET", {
+  con <- test_con("SQLSERVER")
+
+  df <- data.frame(tz_char = rep("2025-05-10 19:35:03.123 +02:00", 3), tz = rep("2025-05-10 19:35:03.123 +02:00", 3))
+
+  tbl <- local_table(con, "test_datetimeoffset", df,
+    field.types = list("tz_char" = "VARCHAR(50)", "tz" = "DATETIMEOFFSET"), overwrite = TRUE)
+  res <- DBI::dbReadTable(con, tbl)
+  expect_true("POSIXct" %in% class(res[[2]]))
```
simonpcouch, line 387:
```suggestion
  expect_s3_class(res[[2]], "POSIXct")
```

simonpcouch (approved):
Thanks for the thorough description! This seems reasonable to me. Let's add a NEWS entry before merging!

PR #898, opened 2025-03-01
nanodbc: backport TVP patches

detule:
Hi:

This catches up our vendored nanodbc lib to upstream in terms of TVP functionality.  Not part of this PR:

* Upstream TVP functionality that [is in review](https://github.com/nanodbc/nanodbc/pull/426).
* Updates to 'odbc' API to take advantage of this feature. 

On the last two I will prepare a new PR; the diff on this one alone ended up being large enough that I thought breaking it out might make sense in terms of being able to better attribute any changes in existing behavior ( hopefully none! ).

Compare upstream: https://github.com/nanodbc/nanodbc/pull/259

PR #928, opened 2025-07-09
Ability to bind data.frames to table valued parameters

detule:
Hi:

There are two commits in this PR.

* One is a backport of some [upstream changes](https://github.com/nanodbc/nanodbc/pull/426) in `nanodbc` that complete the TVP API.
* Another is adding [R] bindings to make it possible to bind `data.frame`s to TVPs.

It is relatively common for SQL Server developers to create sprocs that take a [table valued parameters](https://learn.microsoft.com/en-us/sql/relational-databases/tables/use-table-valued-parameters-database-engine?view=sql-server-ver17) (TVP) as an argument, as traditionally this has been one of the more efficient ways to organize the arguments without needing to use intermediate storage.
Until now, we didn't have a way to execute these sprocs from [R]; this PR makes it possible to do so, by adding the ability to bind `data.frame`s to TVP.

There is example usage in the unit tests.  I was thinking about pulling those out as stand alone examples, but was at a bit of a loss of where to put them.  Perhaps in the documentation of `SQLServer-class`? Not sure, let me know if you have suggestions.

I wonder if long term we could benefit from some back-end specific vignettes.

This PR is related to #898 (though, that one stands on its own).

### DIFF
--- a/R/dbi-result.R
+++ b/R/dbi-result.R
@@ -139,6 +139,24 @@ setMethod("dbBind", "OdbcResult",
       return(invisible(res))
     }
 
+    paramDfs <- sapply(params, is.data.frame)
+    if (any(paramDfs)) {
+      # When binding table-valued-parameters verify that:
+      # * The `batch_rows` parameter is either unset or equal to one.
+      # * All non-tvp parameters are of length 1.
+      if (!is.na(batch_rows) && batch_rows > 1) {
+        cli::cli_warn("When passing data.frame(s) as query parameters,",
+                      "the `batch_rows` parameter must be set to 1")
+      }
+      batch_rows <- 1
+      paramLengths <- sapply(params[!paramDfs], length)
+      if (any(paramLengths > 1))
+      {
+        cli::cli_abort("When mixing data.frame(s) with other parameter types,",
+                       "all non-df parameters must be of length one")
+      }
+    }
+
     if (is.na(batch_rows)) {
       batch_rows <- length(params[[1]])
     }
--- a/R/dbi-result.R
+++ b/R/dbi-result.R
@@ -155,9 +155,7 @@ setMethod("dbBind", "OdbcResult",
         cli::cli_abort("When mixing data.frame(s) with other parameter types,
                        all non-df parameters must be of length one")
       }
-    }
-
-    if (is.na(batch_rows)) {
+    } else if (is.na(batch_rows)) {
       batch_rows <- length(params[[1]])
     }
 
--- a/R/dbi-result.R
+++ b/R/dbi-result.R
@@ -145,8 +145,8 @@ setMethod("dbBind", "OdbcResult",
       # * The `batch_rows` parameter is either unset or equal to one.
       # * All non-tvp parameters are of length 1.
       if (!is.na(batch_rows) && batch_rows > 1) {
-        cli::cli_warn("When passing data.frame(s) as query parameters,
-                      the `batch_rows` parameter must be set to 1")
+        cli::cli_warn(c("Since some parameters are data frames, {.arg batch_rows} will be overwritten to `1`.", 
+                       "Set {.code batch_rows = 1} to quiet this warning."))
       }
       batch_rows <- 1
       paramLengths <- sapply(params[!paramDfs], length)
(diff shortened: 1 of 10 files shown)


R/dbi-result.R
```diff
@@ -139,6 +139,24 @@ setMethod("dbBind", "OdbcResult",
       return(invisible(res))
     }
 
+    paramDfs <- sapply(params, is.data.frame)
+    if (any(paramDfs)) {
+      # When binding table-valued-parameters verify that:
+      # * The `batch_rows` parameter is either unset or equal to one.
+      # * All non-tvp parameters are of length 1.
+      if (!is.na(batch_rows) && batch_rows > 1) {
+        cli::cli_warn("When passing data.frame(s) as query parameters,",
```
simonpcouch, line 148: Same goes here.

detule:
Thanks @simonpcouch : 

* Fixed `cli` usage (thanks)
* Added more notes in PR
* Updated NEWS
* Tested one more time.

See note in PR about "unsure where to put example".

R/dbi-result.R
```diff
@@ -139,6 +139,24 @@ setMethod("dbBind", "OdbcResult",
...
+      if (any(paramLengths > 1))
+      {
+        cli::cli_abort("When mixing data.frame(s) with other parameter types,
+                       all non-df parameters must be of length one")
+      }
+    }
+
     if (is.na(batch_rows)) {
       batch_rows <- length(params[[1]])
     }
```
simonpcouch, line 162: I believe if the first parameter is a data frame here, `batch_rows` will be overwritten to the number of columns of the data frame rather than 1?

tests/testthat/test-driver-sql-server.R
```diff
@@ -423,3 +423,45 @@ test_that("Mixed success multiple result-sets (#924)", {
...
+                    ORDER BY col0;
+                    RETURN 0;
+                  END")
+  on.exit({
+    dbExecute(con, "DROP PROCEDURE tvp_test")
+    dbExecute(con, "DROP PROCEDURE tvp_test2")
+    dbExecute(con, "DROP TYPE tvp_param")
+  })
+
+  df.param <- data.frame(int = 1:10, bigint=1:10, vrchr = rownames(mtcars)[1:10],
```
simonpcouch, line 454:
```suggestion
  df.param <- data.frame(int = 1:10, bigint = 1:10, vrchr = rownames(mtcars)[1:10],
```

R/dbi-result.R
```diff
@@ -139,6 +139,24 @@ setMethod("dbBind", "OdbcResult",
...
 
+    paramDfs <- sapply(params, is.data.frame)
+    if (any(paramDfs)) {
+      # When binding table-valued-parameters verify that:
+      # * The `batch_rows` parameter is either unset or equal to one.
+      # * All non-tvp parameters are of length 1.
+      if (!is.na(batch_rows) && batch_rows > 1) {
+        cli::cli_warn("When passing data.frame(s) as query parameters,
+                      the `batch_rows` parameter must be set to 1")
```
simonpcouch, line 149:
```suggestion
        cli::cli_warn(c("Since some parameters are data frames, {.arg batch_rows} will be overwritten to `1`.", 
                       "Set {.code batch_rows = 1} to quiet this warning."))
```

Since the code will still run despite the warning, clarify the action taken internally and how to prevent it

simonpcouch (approved):
This is great!

Is "table-valued parameter" common parlance in the odbc/database space? The reason I ask is that "table" is a base class in R that is different from a data frame:

``` r
v <- c(2, 2, 3)

table_v <- table(v)
table_v
#> v
#> 2 3 
#> 2 1
class(table_v)
#> [1] "table"
is.atomic(table_v)
#> [1] TRUE

data_frame_v <- data.frame(v)
data_frame_v
#>   v
#> 1 2
#> 2 2
#> 3 3
class(data_frame_v)
#> [1] "data.frame"
is.atomic(data_frame_v)
#> [1] FALSE
```

<sup>Created on 2025-08-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

When I read "table-valued parameter," my mind goes to objects of class `table` rather than data frames.

For now, this distinction has only had to be made in user-facing docs in `NEWS.md` (and I like the way you've done it there), but worth repeating in docs you may write re: the below.


> I wonder if long term we could benefit from some back-end specific vignettes.

I like this idea.

In the meantime, I like your suggestion that the docs on this go in [the SQLServer page](https://odbc.r-dbi.org/reference/SQLServer.html). May be worth calling out there that `batch_rows = 1` always, there.

detule:
Thanks Simon:

Unfortunately "table-valued-parameters" is the name these go by, but I agree with you "table" is very much overloaded.  I tried to strike some sort of a middle ground and refer to them as `TVP`s in the docs I just wrote.

Issue #930, opened 2025-07-10
Add support for Posit Connect service account flow with Databricks using `connectcreds`


tnederlof:
There is already a helper function that detects when "Viewer" type credentials exist here: https://github.com/r-dbi/odbc/blob/258b74a71f765f905403701ee181d904550a9844/R/driver-databricks.R#L219

It would be helpful for those using "Service Account" style credentials to be able to use the same helper function. It looks like `connectcreds` already has the functionality needed:
https://github.com/posit-dev/connectcreds/blob/53a4db92eb7b5fc81ed49118f662d88e7ff4886a/R/service-accounts.R#L76
https://github.com/posit-dev/connectcreds/blob/53a4db92eb7b5fc81ed49118f662d88e7ff4886a/R/service-accounts.R#L21

detule:
Closing this - as I think the `databricks` wrappers will query the `connectcreds` SPN api.  Happy to re-open if something is missing.

PR #931, opened 2025-07-11
Add Databricks support for service principals when running in Posit Connect with connectcreds

tnederlof:
~~**CANNOT MERGE until `connectcreds` release**~~

This commit adds support for service principal credentials managed by
Posit Connect alongside our existing support of viewer-based
credentials. The meat of this is handled by the `connectcreds` package.

Documentation and unit test updates are included.

Addresses: https://github.com/r-dbi/odbc/issues/930

### DIFF
--- a/R/driver-databricks.R
+++ b/R/driver-databricks.R
@@ -15,9 +15,9 @@ NULL
 #' model, with support for personal access tokens, OAuth machine-to-machine
 #' credentials, and OAuth user-to-machine credentials supplied via Posit
 #' Workbench or the Databricks CLI on desktop. It can also detect viewer-based
-#' credentials on Posit Connect if the \pkg{connectcreds} package is
-#' installed. All of these credentials are detected automatically if present
-#' using [standard environment variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
+#' and service principal credentials on Posit Connect if the \pkg{connectcreds}
+#' package is installed. All of these credentials are detected automatically if
+#' present using [standard environment variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
 #'
 #' In addition, on macOS platforms, the `dbConnect()` method will check
 #' for irregularities with how the driver is configured,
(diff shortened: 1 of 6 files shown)


R/driver-databricks.R
```diff
@@ -227,6 +231,15 @@ databricks_auth_args <- function(host, uid = NULL, pwd = NULL) {
     ))
   }
 
+  if (is_installed("connectcreds") && connectcreds::has_service_account_token(workspace)) {
```
simonpcouch, line 234: We could also just add a conditional on the connectcreds package version here, I think? And then merge without a Remotes on the dev version.

atheriel:
@simonpcouch We haven't been able to get it to work yet, so I'd say no, don't hold up the release.

R/driver-databricks.R
```diff
@@ -136,9 +137,12 @@ databricks_args <- function(httpPath,
     if (running_on_connect()) {
       msg <- c(
         msg,
-        "i" = "Or consider enabling Posit Connect's Databricks integration \
-              for viewer-based credentials. See {.url \
-              https://docs.posit.co/connect/user/oauth-integrations/#adding-oauth-integrations-to-deployed-content}
+        "i" = "Or consider enabling Posit Connect's Databricks integration. \
+              For viewer-based credentials. See {.url \
```
simonpcouch:
```suggestion
              For viewer-based credentials, see {.url \
```

Will also need a snap update, I think

R/driver-databricks.R
```diff
@@ -51,12 +51,13 @@ NULL
...
 #'
-#' # Use credentials from the viewer (when possible) in a Shiny app
-#' # deployed to Posit Connect.
+#' # Use credentials from the viewer or a service principal (when possible) in
+#' # a Shiny app deployed to Posit Connect.
 #' library(connectcreds)
 #' server <- function(input, output, session) {
 #'   conn <- DBI::dbConnect(
 #'     odbc::databricks(),
+#'     workspace = "https://example.cloud.databricks.com",
```
simonpcouch, line 60: Would it make sense to add this same example parameter to the first `dbConnect` example here? Or it uniquely applicable inside the Shiny server?

simonpcouch (approved):
Thank you!

R/driver-databricks.R
atheriel: We detect the workspace parameter automatically in most cases, so this is really just a way to show how you'd override it if necessary.

PR #933, opened 2025-08-09
snowflake: add private key connection attributes

### LINKED ISSUE #923: Snowflake key-pair authentication and key file
I'm using the Snowflake ODBC Driver for Linux with the `SNOWFLAKE_JWT` authenticator.

This authenticator has a parameter called [PRIV_KEY_FILE](https://docs.snowflake.com/en/developer-guide/odbc/odbc-parameters#using-key-pair-authentication), which I use to point to the file containing the private key, inside my `dbConnect()` calls.

From what I understand its not possible to pass key pair itself directly to `dbConnect()` , just the file name containing the key.

Since I'm pulling the key pair from a vault and prefer not to write the key to a file, I tried using the attribute called [SQL_SF_CONN_ATTR_PRIV_KEY ](https://docs.snowflake.com/en/developer-guide/odbc/odbc-api#label-odbc-api-sqlsetconnectattr-specific-behavior).

I see it's not currently supported after checking the attributes listed [here](https://odbc.r-dbi.org/reference/ConnectionAttributes.html).

detule:
Hi:

Closes #923 

Following up on a user-reported use case - one where using the [suggested](https://posit.co/blog/running-tidymodel-prediction-workflows-inside-databases/) workflow for authenticating with RSA credentials, raises a security concern when a key is written to disk.

With these changes users can communicate the key to the driver as a location in memory, without needing to store it to disk.

Had a chance to test these changes locally with both `OEM` and `pro` drivers.  Not sure how to organize a unit test for them.

### DIFF
--- a/R/odbc-connection.R
+++ b/R/odbc-connection.R
@@ -199,9 +199,8 @@ needs_quoting <- function(x) {
 #'   with limited support away from the OEM Microsoft driver.
 #' * `sf_private_key`: This parameter is specific to establishing a connection
 #'   to `snowflake` and is understood by both OEM, as well as `Posit`
-#'   pro drivers.  Argument hould be a string (scalar); in particular the
-#'   contents of your BASE64 encoded private key (file).  String should
-#'   include newline characters as well as key header and footer.  Note,
+#'   pro drivers.  Argument should be a string (scalar); in particular a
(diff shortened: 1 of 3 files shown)


simonpcouch (approved):
This seems reasonable to me!

PR #951, opened 2025-10-05
mssql: DATETIMEOFFSET: When transforming to UTC, use ISO 8601 convention

### LINKED ISSUE #946: DATETIMEOFFSET converted with wrong sign on offset (times shifted opposite direction)
When reading a SQL Server DATETIMEOFFSET column, the package applies the time zone offset in the wrong direction, causing values to be shifted incorrectly (e.g. 2025-05-10 19:00:00 -04:00 becomes 2025-05-10 15:00:00 UTC instead of 23:00 UTC).

detule:
Closes #946 

In the previous release we introduced a SQL SERVER improvement where `DATETIMEOFFSET` data is automatically transformed into UTC and returned as `POSIXct` ( previously returned as strings containing the offset ).  As part of that transformation we use `cctz` to transform the specified offset into a timezone and in turn into `UTC` time.

As it turns out there are competing (and opposing) conventions for the direction of the offset:

* POSIX: positive offset denotes time zone west of Greenwich.
* ISO8601: positive offset denotes time zone east of Greenwich.

We coded the feature to keep with the `POSIX` convention.  However, I think it probably makes sense to convert to using ISO8601 since it is referenced in the[ Microsoft docs,](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql?view=sql-server-ver17#supported-string-literal-formats-for-datetimeoffset) although the docs don't specifically mention the direction of the offset.

The way coded in this patch, we make a clean break and note it in the release notes.  Given that this is a breaking change, we could introduce a package option (`odbc.mssql.tz_direction`) or somesuch.  I am slightly biased  towards going with what's here since we've only had this code in for a few weeks/months, however can be swayed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,12 @@
 # odbc (development version)
 
+* SQL Server:
+     - Fix time zone interpretation in `DATETIMEOFFSET` data; now follows
+       ISO 8061 convention.  (#946)
+
 # odbc 1.6.3
 
-Addressed a compiler warning on `r-devel-linux-x86_64-fedora-clang` (#941).
+* Addressed a compiler warning on `r-devel-linux-x86_64-fedora-clang` (#941).
 
 # odbc 1.6.2
 
--- a/src/odbc_result.cpp
+++ b/src/odbc_result.cpp
@@ -540,7 +540,15 @@ double odbc_result::as_double(nanodbc::timestampoffset const& tso) {
   // (failed) time-zone.
   std::stringstream stream;
   std::streambuf* oldClogStreamBuf = std::clog.rdbuf(stream.rdbuf());
-  cctz::time_zone tz = fixed_time_zone(offset);
+  // We are looking up fixed offsets
+  // using "Etc/GMT" as the prefix.  The convention
+  // for those zones/offsets is opposite directionally
+  // from the ISO 8061 database convention.  See:
+  // https://github.com/eggert/tz/blob/main/etcetera#L37
+  // https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC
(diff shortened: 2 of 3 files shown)


r2evans:
I was not aware of the posix-v-iso8601 issue, that is ...... frustrating, it's bad enough that some vendors (like microsoft) implement standards in a more creative and selective way. I agree with your plan to stick with whichever method Microsoft implements. However, instead of or in addition to `options(odbc.mssql.tz_direction=..)` thought, is there anything in the ODBC standard where it might be a per-connection thing? I guess I'm just nervous that Microsoft's implementation may be sensitive to server-version, so it should be per-connection instead of on a per-process basis.

NEWS.md
```diff
@@ -1,8 +1,12 @@
 # odbc (development version)
 
+* SQL Server:
+     - Fix time zone interpretation in `DATETIMEOFFSET` data; now follows
+       ISO 8061 convention.  (#946)
```
simonpcouch, line 5:
```suggestion
       ISO 8061 convention where positive offset denotes time zone east of Greenwich. (#946)
```

simonpcouch (approved):
I agree that we're okay to not introduce an option here. This seems reasonable to me!

Will defer to discussion between you and @r2evans on whether this is the best approach, but thumbs up from me if you end up feeling like this is indeed the way.

detule:
> Are you familiar with [nanodbc/nanodbc#432](https://github.com/nanodbc/nanodbc/pull/432)? If they provide native support for `DATETIMEOFFSET` then do we need to parse/interpret the offset here?

Right - that's just code from our patch that we upstream-ed to `nanodbc`; i think there is no duplicated effort.

PR #954, opened 2025-10-18
date/time: handle r types with integer storage

### LINKED ISSUE #952: Dates with integer representation cannot be copied to SQL Server
In R 4.5, `seq.Date()` was changed to no longer coerce the internal representation to floating point and instead keeps the dates as integer representation (https://cran.r-project.org/bin/windows/base/old/4.5.0/NEWS.R-4.5.0.html)

Since I use `seq.Date()` in my own work over at [ssi-dk/diseasystore](https://github.com/ssi-dk/diseasystore), I started seeing failures in our test workflow that uses SQL Server with errors like:

```
Error in `db_copy_to(dest$con, name, df, overwrite = overwrite, types = types, 
    temporary = temporary, unique_indexes = unique_indexes, indexes = indexes, 
    analyze = analyze, in_transaction = in_transaction, ...)`: Can't copy data to table "#ds_all_dates_tkhMQuqMpA".
Caused by error in `dplyr::db_write_table()`:
! Can't write table table "#ds_all_dates_tkhMQuqMpA".
Caused by error:
! nanodbc/nanodbc.cpp:2077: 00000
[Microsoft][ODBC Driver 18 for SQL Server]Restricted data type attribute violation
```

After some digging, I tracked the issue back to the internal representation of dates since R 4.5 and I could fix the issue in my own code base my casting to numeric
`as.Date(as.numeric(..))`

Since I assume I am not the only person that will encounter this issue, I have been tinkering with a fix here to help others.

I have a proposed solution in PR #953 that converts Integer Dates to numeric Dates in `sqlData()`

Let me know if you think this is the right place to tackle this problem.

detule:
Closes #952
Ability to handle [R] `Date` and `POSIXct` data that may have integer storage.


simonpcouch (approved):
🏄

PR #967, opened 2025-11-23
man: figure/width css fixup

### LINKED ISSUE #965: CRAN check NOTE needing resubmission by 12/15
> Specifically, please see the ‘height/width attributes should be in pixels’ NOTEs in the Rd files check.
>
> These are from using the \figure{FILENAME}{options: STRING} expert form in Rd files with options specifying height or width attributes not in pixels, typically in percentages.  The HTML5 standard mandates that these are in pixels: to use percentages or other units, please use CSS style attributes instead.  E.g., change
>
>  \figure{FILENAME}{options: width=“25\%“}
>
> to
>
>  \figure{FILENAME}{options: style=“width: 25\%;“}
>
> which should take only very little time.
>
> Please correct before 2025-12-15 to safely retain your package on CRAN.

The `musl` check failure on the [package check page](https://cran.r-project.org/web/checks/check_results_odbc.html) is addressed already.

detule:
Closes #965

### DIFF
--- a/R/dbi-driver.R
+++ b/R/dbi-driver.R
@@ -120,7 +120,7 @@ setMethod("show", "OdbcDriver",
 #' ![A diagram containing four boxes with arrows linking each pointing left to
 #' right. The boxes read, in order, R interface, driver manager, ODBC driver,
 #' and database. The left-most box, R interface, contains three smaller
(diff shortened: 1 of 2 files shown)


simonpcouch (approved):
Sweet!!!

PR #971, opened 2025-11-29
Fixup/paws.common dependency

detule:
This to help the windows/4.1 pipeline pass.

Our `driver-redshift` methods use `paws.common::locate_credentials` which is only exported in `paws.common >= 0.7.1`.  Our windows/4.1 pipeline is failing b/c CRAN mirrors carry a 4.1+windows/binary release 0.5.x.

Ideally, we would circumvent the binary releases by specifying the version dependency in `DESCRIPTION.Suggests`, however looks like `paws.common` doesn't (easily) build on Windows+4.1.  At first glance it looks like the package has an unstated requirement of `c++14` but Windows/4.1 builds with `c++11` by default.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,7 +36,7 @@ Suggests:
     httr2,
     knitr,
     magrittr,
-    paws.common (>= 0.7.1),
+    paws.common,
     rmarkdown,
     RSQLite,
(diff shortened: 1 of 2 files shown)


detule:
Closing in favor of https://github.com/r-dbi/odbc/pull/1020

PR #975, opened 2025-12-16
Use Snowflake config.toml/connection.toml files

jcheng5:
~**Vibe coded; just for discussion purposes for now**~ Extremely carefully reviewed by @jcheng5, except unit tests (life is too short)

---

Add support for loading connection parameters from TOML config files (~/.snowflake/connections.toml and config.toml) following the Python Snowflake connector conventions.

New features:
- `connection_name` parameter to load a named connection profile
- `connections_file_path` parameter to specify custom config location
- Support for SNOWFLAKE_CONNECTIONS env var (full TOML override)
- Support for SNOWFLAKE_DEFAULT_CONNECTION_NAME env var
- File permission security checks (error on writable, warn on readable)
- Graceful fallback when toml package not installed

Connection resolution:
- Case 1: connection_name provided → load and merge with kwargs
- Case 2: No args → load default connection from config
- Case 3: Programmatic params → use params only, skip config loading

SNOWFLAKE_ACCOUNT env var is only used in Case 3 (programmatic mode), not when loading from config files.

Requires the toml package (Suggests) for config file parsing.

### DIFF
--- a/R/driver-snowflake.R
+++ b/R/driver-snowflake.R
@@ -134,13 +134,6 @@ setMethod("odbcDataType", "Snowflake",
 #' @param drv an object that inherits from [DBI::DBIDriver-class],
 #' or an existing [DBI::DBIConnection-class]
 #' object (in order to clone an existing connection).
-#' @param connection_name The name of a connection profile to load from
-#'   configuration files. If provided, connection parameters are loaded from
-#'   `~/.snowflake/connections.toml` (or the location specified by
-#'   `SNOWFLAKE_HOME`) and merged with any explicitly provided arguments.
-#'   Requires the \pkg{toml} package.
-#' @param connections_file_path Optional custom path to a `connections.toml`
-#'   file. If `NULL`, uses the default location within `SNOWFLAKE_HOME`.
 #' @param account A Snowflake [account
 #'   identifier](https://docs.snowflake.com/en/user-guide/admin-account-identifier),
 #'   e.g. `"testorg-test_account"`. If not provided and no `connection_name`
@@ -157,6 +150,13 @@ setMethod("odbcDataType", "Snowflake",
 #'   Specifying these options will disable ambient credential discovery and
 #'   config file loading (unless `connection_name` is also specified).
 #' @param ... Further arguments passed on to [`dbConnect()`].
+#' @param connection_name The name of a connection profile to load from
+#'   configuration files. If provided, connection parameters are loaded from
+#'   `~/.snowflake/connections.toml` (or the location specified by
+#'   `SNOWFLAKE_HOME`) and merged with any explicitly provided arguments.
+#'   Requires the \pkg{toml} package.
+#' @param connections_file_path Optional custom path to a `connections.toml`
+#'   file. If `NULL`, uses the default location within `SNOWFLAKE_HOME`.
 #'
 #' @returns An `OdbcConnection` object with an active connection to a Snowflake
 #'   account.
@@ -213,8 +213,6 @@ setClass("SnowflakeOdbcDriver", contains = "OdbcDriver")
 setMethod(
   "dbConnect", "SnowflakeOdbcDriver",
   function(drv,
-           connection_name = NULL,
-           connections_file_path = NULL,
            account = NULL,
            driver = NULL,
            warehouse = NULL,
@@ -222,7 +220,9 @@ setMethod(
            schema = NULL,
            uid = NULL,
            pwd = NULL,
-           ...) {
+           ...,
+           connection_name = NULL,
+           connections_file_path = NULL) {
     call <- caller_env()
     check_string(connection_name, allow_null = TRUE, call = call)
     check_string(connections_file_path, allow_null = TRUE, call = call)
--- a/R/driver-snowflake.R
+++ b/R/driver-snowflake.R
@@ -270,19 +270,6 @@ setMethod(
 # Authentication is delegated to snowflake_auth_args(), which handles various
 # auth methods including uid/pwd, externalbrowser, SNOWFLAKE_JWT, OAuth tokens
 # from Posit Connect, and Workbench-managed credentials.
-#
-# @param connection_name Name of connection to load from connections.toml
-# @param connections_file_path Optional path to connections.toml file
-# @param account Snowflake account identifier
-# @param driver ODBC driver name or path
-# @param warehouse Snowflake warehouse name
-# @param database Snowflake database name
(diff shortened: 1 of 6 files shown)


R/driver-snowflake.R
```diff
@@ -202,18 +251,101 @@ setMethod(
...
+  if (!is.null(account)) programmatic_params$account <- account
+  if (!is.null(warehouse)) programmatic_params$warehouse <- warehouse
+  if (!is.null(database)) programmatic_params$database <- database
+  if (!is.null(schema)) programmatic_params$schema <- schema
+  if (!is.null(uid)) programmatic_params$uid <- uid
+  if (!is.null(pwd)) programmatic_params$pwd <- pwd
+
+  # Add any additional params from ...
+  dots <- list(...)
+  programmatic_params <- c(programmatic_params, dots)
```
jcheng5: Is it guaranteed that the keys in `programmatic_params` and `dots` are disjoint?

R/driver-snowflake.R
```diff
@@ -392,3 +524,358 @@ workbench_snowflake_token <- function(account, sf_home) {
...
+  } else {
+    # Linux: respect XDG_CONFIG_HOME or default to ~/.config/snowflake/
+    xdg_config <- Sys.getenv("XDG_CONFIG_HOME", "")
+    if (nchar(xdg_config) > 0) {
+      return(file.path(xdg_config, "snowflake"))
+    } else {
+      return(path.expand("~/.config/snowflake"))
+    }
+  }
+}
```
jcheng5: Nailed it AFAICT, I'm impressed.

R/driver-snowflake.R
```diff
@@ -392,3 +524,358 @@ workbench_snowflake_token <- function(account, sf_home) {
...
+  env_connections <- Sys.getenv("SNOWFLAKE_CONNECTIONS", "")
+  if (nchar(env_connections) > 0) {
+    result$connections <- parse_toml_string(env_connections)
+    # Also check for default connection name override
+    env_default <- Sys.getenv("SNOWFLAKE_DEFAULT_CONNECTION_NAME", "")
+    if (nchar(env_default) > 0) {
+      result$default_connection_name <- env_default
+    }
+    return(result)
+  }
```
jcheng5: If `$SNOWFLAKE_CONNECTIONS`, then config.toml is no longer read--is that the same as Python? Like, could you use `$SNOWFLAKE_CONNECTIONS` to define the connections but config.toml to declare which one should be the default? (It does sound far-fetched and I think either answer would be defensible if the Python reference implementation didn't exist... and arguably still defensible regardless of what Python does.)

jcheng5: Actually this function needs to be completely rewritten--as it stands, it's too hard to reason about edge cases.

jcheng5: Rewritten to use promises (the other kind of promises), way way simpler to reason about now.

R/driver-snowflake.R
jcheng5: WOW, I asked Claude and it said it's not guaranteed, because R has case sensitive parameters but ODBC is typically case insensitive! Impressive!

jcheng5: (That being said, I don't care about that case)

R/driver-snowflake.R
```diff
@@ -202,18 +251,133 @@ setMethod(
...
+                           schema = NULL,
+                           uid = NULL,
+                           pwd = NULL,
                            ...) {
+  # Check for Workbench mode early - skip config loading if SNOWFLAKE_HOME
+  # contains "posit-workbench" (Workbench handles credentials differently)
+  sf_home <- Sys.getenv("SNOWFLAKE_HOME")
+  is_workbench_mode <- grepl("posit-workbench", sf_home, fixed = TRUE)
+
+  # 1. Collect programmatic parameters (only non-NULL values)
```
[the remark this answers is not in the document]
simonpcouch: In this case, "programmatic" means "arguments passed explicitly by the user rather than implicitly via config files"?

R/driver-snowflake.R
```diff
@@ -202,18 +251,133 @@ setMethod(
...
+# @param account Snowflake account identifier
+# @param driver ODBC driver name or path
+# @param warehouse Snowflake warehouse name
+# @param database Snowflake database name
+# @param schema Snowflake schema name
+# @param uid User ID for authentication
+# @param pwd Password for authentication
+# @param ... Additional connection parameters
+# @return Named list of connection arguments suitable for dbConnect(odbc(), ...)
+# @noRd
```
simonpcouch: These are `#` rather than `#'` so no need for `@noRd`, and the arguments are already documented in the user-facing function.

R/driver-snowflake.R
```diff
@@ -392,3 +556,365 @@ workbench_snowflake_token <- function(account, sf_home) {
...
+#'
+#' Resolves the configuration directory with the following precedence:
+#' 1. SNOWFLAKE_HOME environment variable (with ~ expansion)
+#' 2. ~/.snowflake/ if it exists
+#' 3. Platform-specific defaults
+#'
+#' @return Character string path to config directory
+#' @noRd
+snowflake_config_dir <- function() {
+  # 1. Check SNOWFLAKE_HOME env var with ~ expansion
```
simonpcouch: Same comment re `1. `, `2. `, `3. `. 

The helpers from here on feel tightly-scoped enough that I think we can portions of the roxygen2-style comments that document the "what" rather than the "why".

tests/testthat/test-driver-snowflake.R
```diff
@@ -189,3 +194,689 @@ test_that("Workbench-managed credentials are ignored for other accounts", {
...
+
+  expect_equal(config$default_connection_name, "staging")
+})
+
+test_that("Case 1: Named connection loads and merges with kwargs", {
+  skip_if_not_installed("toml")
+
+  config_dir <- tempfile()
+  dir.create(config_dir)
+  withr::defer(unlink(config_dir, recursive = TRUE))
```
simonpcouch:
```suggestion
  config_dir <- withr::local_tempdir()
```

And elsewhere in this test file

simonpcouch (commented):
I don't have strong reflexes here on the big idea, but trust that this is a reasonable approach! Reviewing for R code style etc. 

Could you add a NEWS entry?

simonpcouch (approved):
Looking good🗃️

simonpcouch:
Closing in favor of #1000!

PR #976, opened 2025-12-17
configure: CXX11 and CXX1X no longer supported

detule:
Follow up on BDR email.

I think this is straightforward - at this point, in all supported versions of [R] the default compiler is C++11 capable.  Therefore it is OK to use the "CXX" variable.

### DIFF
--- a/configure
+++ b/configure
@@ -53,11 +53,7 @@ else
 fi
 
 # Find compiler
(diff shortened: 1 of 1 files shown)


simonpcouch (approved):
🫠 Thanks for making this happen!

PR #984, opened 2026-01-16
Fix error and/or r-crash when lengths(params) are different

r2evans:


R/dbi-result.R
```diff
@@ -163,8 +163,18 @@ setMethod("dbBind", "OdbcResult",
         cli::cli_abort("When mixing data.frame(s) with other parameter types,
                        all non-df parameters must be of length one")
       }
-    } else if (is.na(batch_rows)) {
-      batch_rows <- length(params[[1]])
+    } else {
+      lens <- lengths(params[!paramDfs])
+      maxlen <- max(lens)
+      if (!all(lens %in% c(maxlen, 1L))) {
+        cli::cli_abort("When sending multiple parameters, all must be the same length or length 1")
```
hadley: odbc already implicitly depends on vctrs via blob and hms, so I think we could make an explicit dependency, and then you could do:

```R
params[!paramDfs] <- vctrs::vec_common_recycle(!!!params[!paramDfs], .arg = "params")
```

And that will handle both the error message and the recycling.

tests/testthat/test-driver-sql-server.R
```diff
@@ -495,3 +495,30 @@ test_that("Variable date type storage", {
...
+
+test_that("Recycling in dbBind works (#491)", {
+  dat_in <- data.frame(id = 1000:1004, timestamp = sprintf("2022-04-01 12:00:%02d -04:00", 1:5))
+  dat_expect <- subset(dat_in, id %in% 1003:1004) |>
+    transform(timestamp = as.POSIXct(timestamp, tz = "America/New_York"))
+  rownames(dat_expect) <- NULL
+
+  con <- test_con("SQLSERVER")
+  tbl_name <- "recyclingtemptable"
+  dbWriteTable(con, tbl_name, dat_in, field.types = c(id = "int", timestamp = "DATETIMEOFFSET"))
```
detule: Consider using `local_table` here, (or properly namespacing `withr::defer`)

PR #997, opened 2026-04-07
snowflake: sf_private_key: fixup buffer allocation

### LINKED ISSUE #982: Crash while using Snowflake sf_private_key key
Hello!

A few months back I created #923.

Back then it worked flawlessly but when I tried again today I stumbled across a session crash. I'm using the same snippet as back then:

```
con <- dbConnect(
  odbc::odbc(),
  driver = "SnowflakeDSIIDriver",
  server = "XXX",
  uid = "YYY",
  authenticator = "snowflake_jwt",
  attributes = list("sf_private_key" = pk),
)
```
**The R session crashes with a coredump: `corrupted size vs. prev_size`.**

**I found out that if I remove the last `\n` of the private key it works normally. I'm just using `sub("\n$", "", pk)`.**

The difference is exactly 1 character:

```
nchar(pk_raw)
[1] 1704
nchar(pk_fixed)
[1] 1703

pk_raw: ends with -----END PRIVATE KEY-----\n ←
pk_fixed: ends with -----END PRIVATE KEY----- ←
```

However, from what I know the last `\n` in an openssh private key file is perfectly normal. I checked straigth from keys created using `ssh-keygen`:

```
0000640   D       O   P   E   N   S   S   H       P   R   I   V   A   T
0000660   E       K   E   Y   -   -   -   -   -  \n
0000673
```

And the openssh library also reports the key to be vaild before removing the last `\n`.

```
library(openssl)
Linking to: OpenSSL 3.0.13 30 Jan 2024
tryCatch({
+   read_key(pk)
+   message("Key format is valid!")
+ }, error = function(e) {
+   message("Key format invalid: ", e$message)
+ })
Key format is valid!
```

detule:
Attempt to resolve #982

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -60,7 +60,7 @@ namespace utils {
       {
         std::shared_ptr<std::string> priv_key =
           std::make_shared<std::string>(Rcpp::as<std::string>(r_attributes["sf_private_key"]));
-        std::shared_ptr< void > buffer(malloc(priv_key->size()), std::free);
+        std::shared_ptr< void > buffer(malloc(priv_key->size() + 1), std::free);
         // Copy null terminator as well
         std::memcpy(buffer.get(), priv_key->c_str(), priv_key->size() + 1);
         attributes.push_back(nanodbc::connection::attribute(
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -72,7 +72,7 @@ namespace utils {
       {
         std::shared_ptr<std::string> key_pass =
           std::make_shared<std::string>(Rcpp::as<std::string>(r_attributes["sf_private_key_password"]));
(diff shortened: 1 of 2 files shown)


detule:
I had a chance to test this locally and was able to replicate the user report and test the fix (works)!

src/utils.cpp
simonpcouch:
```suggestion
        std::shared_ptr< void > buffer(malloc(key_pass->size() + 1), std::free);
```

Not touched by this PR, but feels worth nothing—does this line have the same bug?

(For whatever reason, the suggestion UI looks botched to me locally, but I've just added the same ` + 1` here.

simonpcouch (approved):
This is great--looks good to me.

One comment, though I trust your judgment if it's not relevant and can be ignored.

src/utils.cpp
detule: Good call!

PR #1000, opened 2026-04-23
Use the 'snowflakeauth' package for odbc::snowflake()

atheriel:
The `snowflakeauth` package targets feature parity with the `connections.toml` and `config.toml` files used in existing Snowflake packages and SDKs. This commit refactors our partial support for these files to use this package, which has a number of benefits:

* Support for non-OAuth authenticators (including "externalbrowser").
* Support for the `host` field.
* Support for named connections via a new `connection_name` argument.
* Generally improved compatibility with the Snowflake CLI and Python ecosystem.

Unit tests are included.

Note that this commit has very similar functionality to #975, but it makes use of the `snowflakeauth` package instead, part of an effort on my part to improve the consistency of our handling of Snowflake auth across the R ecosystem.

### DIFF
--- a/R/driver-snowflake.R
+++ b/R/driver-snowflake.R
@@ -99,10 +99,11 @@ setMethod("odbcDataType", "Snowflake",
 #' driver](https://docs.snowflake.com/en/developer-guide/odbc/odbc).
 #'
 #' In particular, the custom `dbConnect()` method for the Snowflake ODBC driver
-#' detects ambient OAuth credentials on platforms like Snowpark Container
-#' Services or Posit Workbench. It can also detect viewer-based
-#' credentials on Posit Connect if the \pkg{connectcreds} package is
-#' installed.
+#' supports the `connections.toml` and `config.toml` files used by the Snowflake
+#' Connector for Python and the Snowflake CLI via the \pkg{snowflakeauth}
+#' package. It also detects ambient OAuth credentials on platforms like Snowpark
+#' Container Services or Posit Workbench. Finally, it can detect viewer-based
+#' credentials on Posit Connect if the \pkg{connectcreds} package is installed.
 #'
 #' In addition, on macOS platforms, the `dbConnect` method will check and warn
 #' if it detects irregularities with how the driver is configured, unless the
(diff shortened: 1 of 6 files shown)


atheriel:
I've manually validated that this works in a few environments with Snowflake credentials available.

R/driver-snowflake.R
```diff
@@ -205,15 +214,25 @@ setMethod(
...
-    server = snowflake_server(account),
+    server = snowflake_server(account, host = host),
     # Connections to Snowflake are always over HTTPS.
     port = 443
   )
-  auth <- snowflake_auth_args(account, ...)
-  all <- utils::modifyList(c(args, auth), list(...))
+  dots <- list(...)
+  dots$connection_name <- NULL
+  dots <- Filter(Negate(is.null), dots)
```
simonpcouch: We have a function:

```
compact <- function(x) x[!vapply(x, is.null, logical(1))]
```

that should do the trick when used in this PR.

R/driver-snowflake.R
```diff
@@ -347,48 +370,83 @@ snowflake_auth_args <- function(account,
...
+    "private_key_file_pwd", "token", "token_file_path", "password", "host"
+  )
+  dots <- Filter(Negate(is.null), dots)
+  sf_args <- c(sf_args, dots[intersect(names(dots), sf_params)])
+  if (file.exists("/snowflake/session/token") &&
+      is.null(sf_args$token_file_path) &&
+      is.null(sf_args$token)) {
+    sf_args$token_file_path <- "/snowflake/session/token"
+  }
+  conn <- inject(snowflakeauth::snowflake_connection(!!!sf_args))
```
simonpcouch, line 389: Could we add a comment here that we're purposefully not passing `account` so that we can compare the supplied one and the one that snowflakeauth resolves? Was initially non-obvious to me that we'd omit `account` here.

simonpcouch (approved):
I'm for it!

Confirming that I'm fine to close the linked PR from Joe? Is there any functionality there that we'd still be interested in incorporating?

R/driver-snowflake.R
atheriel: Fixed!

R/driver-snowflake.R
atheriel: Yep, fixed!

atheriel:
I can confirm we should close Joe's PR, I think I already mentioned to him that I wanted to go this route instead.

PR #994, opened 2026-03-10
databricks: handle new OEM driver

detule:
Closes #993

PR #1001, opened 2026-04-30
fix typo in `configure_simba()`

simonpcouch:
Noticed this while reviewing #994 but was unrelated. Elsewhere in the function, we refer to `func` rather than `fun`.


detule (approved):
Good eye!

PR #1002, opened 2026-04-30
Add support for Databricks workload identity federation

atheriel:
Databricks recently added support for [workload identity federation][0], and Posit Connect now supports it as well, but only for Python content up until now. This commit wires up workload identity inside `odbc::databricks()`.

Unit tests are included, as are updates to the documentation.

[0]: https://docs.databricks.com/aws/en/dev-tools/auth/oauth-federation-exchange

### DIFF
--- a/R/driver-databricks.R
+++ b/R/driver-databricks.R
@@ -11,13 +11,20 @@ NULL
 #' driver](https://www.databricks.com/spark/odbc-drivers-download).
 #'
 #' In particular, the custom `dbConnect()` method for the Databricks ODBC driver
-#' implements a subset of the [Databricks client unified authentication](https://docs.databricks.com/en/dev-tools/auth.html#databricks-client-unified-authentication)
-#' model, with support for personal access tokens, OAuth machine-to-machine
-#' credentials, and OAuth user-to-machine credentials supplied via Posit
-#' Workbench or the Databricks CLI on desktop. It can also detect viewer-based
-#' and service principal credentials on Posit Connect if the \pkg{connectcreds}
-#' package is installed. All of these credentials are detected automatically if
-#' present using [standard environment variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
+#' implements a variant of Databricks's [unified authentication](https://docs.databricks.com/aws/en/dev-tools/auth/unified-auth)
+#' model when no `uid` or `pwd` is supplied, checking for ambient credentials in
+#' the following order:
+#'
+#' * Viewer-based or service principal credentials supplied by Posit Connect
+#'   (requires the \pkg{connectcreds} package).
+#' * Personal access tokens.
+#' * Workload identity federation.
+#' * OAuth machine-to-machine credentials.
+#' * OAuth user-to-machine credentials suplied via Posit Workbench or the
+#'   Databricks CLI on desktop.
+#'
+#' This aims to provide broad compatibility between \pkg{odbc} and the standard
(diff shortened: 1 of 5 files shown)


R/driver-databricks.R
```diff
@@ -13,11 +13,12 @@ NULL
...
-#' Workbench or the Databricks CLI on desktop. It can also detect viewer-based
-#' and service principal credentials on Posit Connect if the \pkg{connectcreds}
-#' package is installed. All of these credentials are detected automatically if
-#' present using [standard environment variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
+#' credentials, workload identity federation, and OAuth user-to-machine
+#' credentials supplied via Posit Workbench or the Databricks CLI on desktop. It
+#' can also detect viewer-based and service principal credentials on Posit
+#' Connect if the \pkg{connectcreds} package is installed. All of these
+#' credentials are detected automatically if present using [standard environment
+#' variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
```
simonpcouch:
```suggestion
#' variables](https://docs.databricks.com/en/dev-tools/auth.html#environment-variables-and-fields-for-client-unified-authentication).
#'
#' When multiple credentials are available, `databricks()` uses the first
#' match in the following order:
#'
#' 1. Viewer-based or service principal credentials on Posit Connect
#' 2. Manual `uid`/`pwd`
#' 3. Workload identity federation (`DATABRICKS_AUTH_TYPE` set to
#'    `"env-oidc"` or `"file-oidc"`)
#' 4. Personal access token (`DATABRICKS_TOKEN`)
#' 5. OAuth machine-to-machine (`DATABRICKS_CLIENT_ID` +
#'    `DATABRICKS_CLIENT_SECRET`)
#' 6. Posit Workbench-managed credentials
#' 7. Databricks CLI
```

In trying to wrap my head around the logic/precedence, I figured it's probably worth documenting that for users too.

tests/testthat/test-driver-databricks.R
```diff
@@ -176,3 +177,210 @@ test_that("tokens can be requested from a Connect server", {
...
+  )
+  expect_snapshot(databricks_auth_args("host"), error = TRUE)
+})
+
+test_that("workload identity env-oidc errors when JWT env var is empty", {
+  withr::local_envvar(
+    DATABRICKS_TOKEN = "",
+    DATABRICKS_CLIENT_ID = "sp",
+    DATABRICKS_AUTH_TYPE = "env-oidc",
+    DATABRICKS_OIDC_TOKEN = ""
```
simonpcouch:
```suggestion
    DATABRICKS_OIDC_TOKEN = "",
    DATABRICKS_OIDC_TOKEN_ENV = ""
```

Should we be unsetting this variable (/other variables) in the tests that rely on it not being around? I don't know which ones actually rely on which, but it seems like depending on the user's local setup, we could easily stumble into a case that's earlier in precedence. I think, in this case, it would read from `DATABRICKS_OIDC_TOKEN_ENV` and potentially succeed if it were set.

simonpcouch (approved):
Nice!

tests/testthat/test-driver-databricks.R
atheriel: Yep, fixed.

R/driver-databricks.R
atheriel: I didn't take this exact wording but I rewrote the docs here to be much more clear about what is happening and in what order.

PR #1004, opened 2026-05-05
odbc 1.7.0

simonpcouch:
Closes #1003.

> ℹ Check <simon.couch@posit.co> for the results in 15-30 mins (~09:55 AM).

PR #1005, opened 2026-05-05
resolve NOTE on `odbc_result.o`

simonpcouch:
On `devtools::check_win_devel()`, I see:

```
* checking compiled code ... NOTE
File 'odbc/libs/x64/odbc.dll':
  Found '_ZSt4cerr', possibly from 'std::cerr' (C++)
    Object: 'odbc_result.o'

Compiled code should not call entry points which might terminate R nor
write to stdout/stderr instead of to the console, nor use Fortran I/O
nor system RNGs nor [v]sprintf.

See 'Writing portable packages' in the 'Writing R Extensions' manual.
```

Note that this is a stacked PR into #1004.


simonpcouch:
Noting that this does resolve the NOTE on win-builder!

detule (approved):
LGTM.

PR #1007, opened 2026-06-06
Feat 1006/bound buffers alloc

### LINKED ISSUE #1006: performance: odbc_result: buffer management
The size of the bound buffers is known in advance / we should pre-allocate this memory on the heap

detule:
This has been on my TODO for a while.

By pre-allocating the buffers we are making sure the vectors are not dynamically resized (and contained elements potentially copied) as we are pushing onto them.

Closes #1006


simonpcouch (approved):
Looks good!

PR #1012, opened 2026-06-18
nanodbc: wvarchar: make sure to not read past pdata_ end

### LINKED ISSUE #1011: wstring_convert::to_bytes crash when Snowflake ODBC metadata contains non-ASCII characters in view REMARKS
When using odbc in Snowflake, we ran across an issue where we couldn't list connections:

```
> conn <- DBI::dbConnect(odbc::snowflake())
+ connections::connection_view(conn)
Error:
! wstring_convert::to_bytes
```

After doing some digging we noticed it's specific to `SNOWFLAKE_PUBLIC_DATA_FREE` which contains data with special characters like `é` and `§`. Setting `Sys.setlocale("LC_ALL", "en_US.UTF-8")` does not resolve it.

Steps to reproduce:

The error surfaces when connecting without scoping to a specific database, so that SQLTables enumerates all catalogs:

```
library(DBI)
library(odbc)

conn <- DBI::dbConnect(odbc::snowflake())
DBI::dbListTables(conn)
# Error: wstring_convert::to_bytes
```

It can be reproduced against any Snowflake account with access to the `SNOWFLAKE_PUBLIC_DATA_FREE`, which contains views with non-ASCII characters in their comments:

# Confirm the offending characters
```
view_info <- DBI::dbGetQuery(conn, "
  SELECT TABLE_NAME, COMMENT 
  FROM SNOWFLAKE_PUBLIC_DATA_FREE.INFORMATION_SCHEMA.VIEWS
  WHERE COMMENT IS NOT NULL AND COMMENT != ''
")
```

`non_ascii <- view_info[grepl("[^\x01-\x7F]", view_info$COMMENT, perl = TRUE), ]`

# Extract the specific characters
```
chars <- regmatches(non_ascii$COMMENT, gregexpr("[^\x01-\x7F]", non_ascii$COMMENT, perl = TRUE))
sort(unique(unlist(chars)))
# Includes: é and §
```

These characters cannot be converted to ASCII:
```
iconv(paste(sort(unique(unlist(chars))), collapse = ""), from = "UTF-8", to = "ASCII")
[1] NA
```

Traceback

```
1. +-connections::connection_view(conn)
2. \-connections:::dbi_schemas(con)
3.   \-DBI::dbListObjects(con)
4.     \-odbc::dbListTables(conn)
5.       \-odbc:::odbcConnectionTables(...)
6.         \-odbc:::connection_sql_tables(...)
Error: wstring_convert::to_bytes
```

Expected behavior

Non-ASCII characters in ODBC metadata (specifically the REMARKS column returned by SQLTables) should be handled gracefully, either by using std::codecvt_utf8<wchar_t> explicitly in wstring_convert, or by substituting/skipping unconvertible characters.

Root cause

The C++ code in connection_sql_tables uses `std::wstring_convert` with a locale-dependent codecvt facet to convert ODBC wide-string metadata to narrow strings. The ODBC SQLTables call returns the REMARKS column for all views, and when any view comment contains a non-ASCII character (e.g., é, §), the conversion throws `std::range_error`. Setting the R/C locale via Sys.setlocale() does not affect the C++ locale used by std::wstring_convert. The fix would be to use std::codecvt_utf8<wchar_t> explicitly.

Session info

R version 4.5.1
odbc 1.7.0
Platform: x86_64-pc-linux-gnu
Positron 2026.05.2

detule:
Snowflake ODBC driver, via the `SQLDescribeCol` api reports SQL_WVARCHAR[255] for the `REMARKS` column of the `SQLTables` output, and we allocate buffers against this report.  However, table comments are actually `VARCHAR[MAX]`, and long comments will overflow these buffers.

Write in some defensive programming to guard against this.

Will work with the `nanodbc` team to figure out if this is appropriate for their repo.

Closes #1011

### DIFF
--- a/src/nanodbc/nanodbc.cpp
+++ b/src/nanodbc/nanodbc.cpp
@@ -4082,8 +4082,11 @@ inline void result::result_impl::get_ref_impl<string_type>(short column, string_
             // Type is unicode in the database, convert if necessary
             const SQLWCHAR* s =
                 reinterpret_cast<SQLWCHAR*>(col.pdata_ + rowset_position_ * col.clen_);
-            const string_type::size_type str_size =
-                col.cbdata_[rowset_position_] / sizeof(SQLWCHAR);
+            // Defensive programming: clamp so we never read past pdata_.
+            // See #1011 / Snowflake for SQLTables REMARKS
+            const string_type::size_type str_size = std::min<string_type::size_type>(
+                col.cbdata_[rowset_position_] / sizeof(SQLWCHAR),
+                col.clen_ / sizeof(SQLWCHAR));
             wide_string_type temp(s, s + str_size);
             convert(temp, result);
         }


simonpcouch (approved):
Thanks!

Issue #934, opened 2025-08-14
Numeric connection string parameters may be formatted incorrectly in connection string


BSchamberger:
Hi odbc team,

`odbc:::build_connection_string` creates a connection string out of the parameters passed to `DBI::dbConnect`/`odbc:::odbc_connect`. In the case one of the parameters is a large numeric, it may be converted to scientific format, potentially yielding an incorrect or invalid connection string.

This has happened to me with the [Simba Spark ODBC Driver](https://docs.databricks.com/_extras/documents/Simba-Apache-Spark-ODBC-Connector-Install-and-Configuration-Guide.pdf) `DefaultStringColumnLength` argument/connection attribute. My data contains a field that may store long JSON strings, hence I set `DefaultStringColumnLength` to a large number to err on the safe side
```r
DBI::dbConnect(
  odbc::odbc(),
  Driver                    = "Simba Spark ODBC Driver",
  SparkServerType           = 3,
  Host                      = "[MY_HOST]",
  Port                      = 443,
  AuthMech                  = 3,
  UID                       = "token",
  PWD                       = "[MY_TOKEN]",
  ThriftTransport           = 2,
  HTTPPath                  = "[MY_PATH]",
  DefaultStringColumnLength = 100000
)
```
which leads to the following connection string generated by `odbc:::build_connection_string` 
```r
[1] "Driver=Simba Spark ODBC Driver;SparkServerType=3;...;DefaultStringColumnLength=1e+05"
```
The `DefaultStringColumnLength=1e+05` part is silently ignored by the driver.

The workaround I currently use it to set the parameter to a string directly (`DefaultStringColumnLength = "100000"`). However, it may be nicer if these cases are handled in the package directly. This issue could be solved by changing the definition of `odbc:::build_connection_string` to explicitly avoid scientific notation with `format(args, scientific = FALSE)` instead of using `args` directly:
```r
build_connection_string <- function(args = list(), string = NULL) {

  args_string <- paste(names(args), format(args, scientific = FALSE), sep = "=", collapse = ";")

  if (!is.null(string) && !grepl(";$", string) && length(args) > 0) {
    string <- paste0(string, ";")
  }

  paste0(string, args_string)
}
```
which yields
```r
[1] "Driver=Simba Spark ODBC Driver;SparkServerType=3;...;DefaultStringColumnLength=100000"
```

My environment:
```
R version 4.5.0 (2025-04-11 ucrt)
Platform: x86_64-w64-mingw32/x64
[...]

PR #1016, opened 2026-06-30
fix: avoid scientific notation in numeric connection attributes

LeonidasZhak:
Summary

Avoid scientific notation when `build_connection_string()` formats numeric
connection attributes, so large values such as `DefaultStringColumnLength`
stay literal in the final ODBC connection string.

Thanks to maintainers

Thanks for maintaining `odbc` and for the concrete report in `#934`.

Issue or motivation

`build_connection_string()` currently relies on the default character coercion
 for numeric values. Large or small numerics can therefore become strings like
`1e+05` or `1e-05`, which some drivers do not accept for connection
attributes.

Root cause

`build_connection_string()` pastes argument values directly, so numeric
arguments inherit R's default scientific notation instead of a plain literal
representation suitable for connection-string attributes.

Change

- Format scalar numeric connection arguments with fixed decimal output and no
  scientific notation.
- Preserve the previous fallback stringification for non-scalar values.
- Add regression coverage for large and small numeric connection attributes.
- Note the fix in `NEWS.md`.

Tests

- `Rscript -e 'pkgload::load_all(".", quiet = TRUE); testthat::test_file("tests/testthat/test-utils.R")'`
- `R CMD build .`
- `_R_CHECK_FORCE_SUGGESTS_=false R CMD check --no-manual odbc_1.7.0.9000.tar.gz`

Scope

This stays within connection-string assembly in `build_connection_string()`.
It does not change driver-specific argument helpers, ODBC quoting rules, or
other numeric formatting paths outside connection-string construction.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # odbc (development version)
 
+* Numeric connection string arguments no longer fall back to scientific
+  notation in `build_connection_string()`, which avoids malformed driver
+  attributes such as `DefaultStringColumnLength` (#934).
+
 # odbc 1.7.0
 
 ## Databricks
--- a/R/utils.R
+++ b/R/utils.R
@@ -584,7 +584,25 @@ find_default_driver <- function(paths, fallbacks, label, call = caller_env()) {
 #' @return string A finalized connection string
 #' @noRd
 build_connection_string <- function(args = list(), string = NULL) {
+  format_connection_value <- function(x) {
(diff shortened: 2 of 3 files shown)


detule:
Closes https://github.com/r-dbi/odbc/issues/934

detule (approved):

detule:
Thanks @LeonidasZhak !

LGTM

PR #1020, opened 2026-07-12
Re-enable Windows R-4.1 CI job and pin paws.common (>= 0.7.1)

detule:
The `os: windows-latest, r: '4.1'` R-CMD-check job was removed in #973 because paws.common wouldn't build from source under rtools40/C++11 on win/4.1.

Reported the issue to the `paws.common` maintainers and looks like they addressed the issue --- it now builds from source under C++11, so that constraint is gone. This restores the job and pins the dependency accordingly.


simonpcouch (approved):
Just realized this slipped by me—apologies for that. This looks good to me!

PR #1025, opened 2026-08-21
nanodbc: handle SQL_NO_TOTAL when retrieving binary data in chunks

### LINKED ISSUE #1024: `BINARY` columns silently truncated to `n mod 1024` bytes (Databricks/Spark ODBC; `pyodbc` unaffected)
## What happens

Reading a `BINARY` column over the Databricks ODBC driver returns fewer bytes than the column holds, with no warning, no error and no truncation indicator. The number of bytes returned is not a cap: it is the true length modulo 1,024.

```r
con <- DBI::dbConnect(odbc::databricks(), httpPath = Sys.getenv("DATABRICKS_PATH"))

DBI::dbGetQuery(con, "SELECT cast(repeat('x', 1024) as binary) AS b")$b[[1]] |> length()
#> 1024        # correct

DBI::dbGetQuery(con, "SELECT cast(repeat('x', 1025) as binary) AS b")$b[[1]] |> length()
#> 1           # one byte returned out of 1,025

DBI::dbGetQuery(con, "SELECT cast(repeat('x', 5000) as binary) AS b")$b[[1]] |> length()
#> 904
```

A 1,025-byte value comes back as a single byte. Nothing in the result indicates that 1,024 bytes are missing.

## The length follows an exact rule

```
bytes returned = n mod 1024        (with 0 mapping to 1024)
```

This holds for every size tested, with no deviation:

| asked | returned | `n mod 1024` |
|---|---|---|
| 1,024 | 1,024 | 0 → 1,024 |
| 1,025 | 1 | 1 |
| 1,026 | 2 | 2 |
| 1,030 | 6 | 6 |
| 1,050 | 26 | 26 |
| 1,100 | 76 | 76 |
| 1,200 | 176 | 176 |
| 1,500 | 476 | 476 |
| 2,000 | 976 | 976 |
| 2,048 | 1,024 | 0 → 1,024 |
| 2,500 | 452 | 452 |
| 3,000 | 952 | 952 |
| 5,000 | 904 | 904 |
| 10,000 | 784 | 784 |
| 20,000 | 544 | 544 |
| 30,000 | 304 | 304 |
| 42,000 | 16 | 16 |

Two consequences worth drawing out. A payload of exactly *k* × 1,024 bytes comes back intact, so a spot check at 1 KiB, 2 KiB or 4 KiB passes while values just above those boundaries are destroyed. And the damage is not proportional to size: 1,025 bytes loses 99.9% of the value, while 2,048 loses nothing.

The exact modulo, and the fact that multiples of 1,024 survive, is consistent with a length counter wrapping in a 1,024-byte buffer. That is an inference from black-box behaviour; I have not inspected driver internals. Combined with the `pyodbc` result below, the counter is most likely on the `odbc` side rather than in the driver.

## The bytes returned are correct as far as they go

The returned bytes are a byte-exact prefix of the true value, so this is truncation rather than corruption.

This needs testing with a non-uniform payload to mean anything: with `repeat('x', n)` every byte is identical, so *any* truncation trivially looks like a clean prefix. Repeating it with a payload whose bytes all differ (5,120 bytes cycling `00..ff` twenty times):

```r
q <- paste("SELECT unhex(repeat(concat_ws('',",
           "transform(sequence(0,255), i -> lpad(hex(i),2,'0'))), 20)) AS b")
v <- DBI::dbGetQuery(con, q)$b[[1]]
length(v)
#> 1024                      # 5,120 mod 1024 == 0, so 1,024
identical(v, as.raw(rep(0:255, 20))[seq_along(v)])
#> TRUE                      # byte-exact prefix of the true value
```

So a byte-count check against the known true length *is* enough to detect this, which is worth stating because it makes a workaround possible for callers who know the expected length.

## No connection-string setting tried changes it

`DefaultBinaryColumnLength` was swept at `2000L`, `8000L`, `65535L` and the string `"100000"`. All four connect successfully and all four return the identical wrong lengths (2,000 → 976, 5,000 → 904, 20,000 → 544).

Integer and string literals were used deliberately. On `odbc` 1.7.0, `build_connection_string()` renders round numbers like `100000` in scientific notation as `1e+05`, which the driver silently ignores; that is [r-dbi/odbc#934](https://github.com/r-dbi/odbc/issues/934), fixed by #1016 but not in the current CRAN release. So a naive sweep can look like "the setting has no effect" when the setting never reached the driver. It did here.

Unlike the STRING defect below, I found no client-side setting that fixes this.

## The same bytes arrive intact as `hex()` STRING

This is the sharpest result I have, and it narrows the problem considerably. On the same driver and the same connection, asking for `hex(col)` instead of the `BINARY` column returns the whole payload:

```r
con <- DBI::dbConnect(
  odbc::databricks(),
  DefaultStringColumnLength = 65535
)

n <- 1025
DBI::dbGetQuery(con, sprintf("SELECT cast(repeat('x', %d) as binary) AS b", n))$b[[1]] |> length()
#> 1
DBI::dbGetQuery(con, sprintf("SELECT hex(cast(repeat('x', %d) as binary)) AS h", n))$h |> nchar()
#> 2050        # 1,025 bytes, complete
```

| asked | as `BINARY` | as `hex()` STRING |
|---|---|---|
| 1,024 | 1,024 | 1,024 |
| 1,025 | **1** | 1,025 |
| 2,000 | **976** | 2,000 |
| 5,000 | **904** | 5,000 |
| 20,000 | **544** | 20,000 |

So neither the driver nor the transport is unable to carry the payload. The loss is specific to the `BINARY` type path.

This also gives callers a workaround that stays on ODBC: select `hex(col)` and decode client-side, at the cost of doubling the bytes on the wire.

```r
h <- DBI::dbGetQuery(con, "SELECT hex(geom) AS h FROM tbl")$h
raw_val <- as.raw(strtoi(substring(h, seq(1, nchar(h), 2), seq(2, nchar(h), 2)), 16L))
```

## `pyodbc` is unaffected, which points at `odbc` rather than the driver

This is the part that identifies the owner. Same driver `.so`, same connection string, same warehouse, minutes apart:

| asked | R, `odbc` 1.7.0 | Python, `pyodbc` 5.3.0 |
|---|---|---|
| 512 | 512 | 512 |
| 1,024 | 1,024 | 1,024 |
| 1,025 | **1** | 1,025 |
| 1,500 | **476** | 1,500 |
| 2,000 | **976** | 2,000 |
| 2,048 | 1,024 | 2,048 |
| 5,000 | **904** | 5,000 |
| 20,000 | **544** | 20,000 |
| 42,000 | **16** | 42,000 |

On a real persisted Delta table with a WKB `BINARY` column, read from storage rather than computed:

| server `length()` | R got | Python got |
|---|---|---|
| 1,046 | **22** | 1,046 |
| 1,062 | **38** | 1,062 |
| 1,078 | **54** | 1,078 |

The Python side used the same driver path (`libsparkodbc_sb64.so`), and `SQLGetInfo` reports the identical driver name and version `2.7.7.1016` on both sides.

I also ran R through raw `odbc::odbc()` with the byte-identical connection string `pyodbc` used, bypassing the `odbc::databricks()` helper entirely:

```r
con <- DBI::dbConnect(odbc::odbc(), .connection_string = cs)  # same string pyodbc used
DBI::dbGetQuery(con, "SELECT cast(repeat('x', 1025) as binary) AS b")$b[[1]] |> length()
#> 1
```

Same loss. So neither the `databricks()` helper nor any connection-string difference explains it, which is what leads me to `SQL_C_BINARY` buffer handling in `odbc` itself.

## No table is involved

Worth stating because it makes this cheap to reproduce: the ladder above uses no table at all. `cast(repeat('x', n) as binary)` is a scalar expression evaluated by the warehouse, so no catalog, schema, table or Delta property is in play. One line reproduces it:

```r
DBI::dbGetQuery(con, "SELECT cast(repeat('x', 1025) as binary) AS b")$b[[1]] |> length()
#> 1        # expected 1025
```

The real-table rows above confirm it also happens on data read from storage, but the table is not the variable.

## Relationship to the STRING truncation

STRING columns are also truncated, at 1,023 characters, filed separately as [r-dbi/odbc#1023](https://github.com/r-dbi/odbc/issues/1023). That one looks genuinely like a driver default: it is fixable from R with `DefaultStringColumnLength = 65535`, and it truncates to a fixed length rather than a modulo. This one behaves differently and `pyodbc` is unaffected, so I do not assume a common cause. Flagging the pair in case the buffer-length handling turns out to be related.

## The data is intact server-side

`brickster::DatabricksSQL()`, which uses the SQL Statement Execution API and decodes Arrow rather than going over the Thrift/ODBC wire format, returns the same columns byte-exact from the same warehouse in the same session:

| asked | ODBC | `DatabricksSQL()` |
|---|---|---|
| 1,024 | 1,024 | 1,024 |
| 2,000 | **976** | 2,000 |
| 5,000 | **904** | 5,000 |
| 20,000 | **544** | 20,000 |

So the column is stored correctly and at least one client retrieves it correctly. This is a driver/transport issue, not bad data.

## Why the silence is the problem

The practical impact is that WKB geometry in a `BINARY` column cannot be read over this driver. A truncated WKB blob either fails to parse, or parses into a smaller valid geometry, and neither outcome mentions the driver. Any binary payload has the same exposure: serialised model objects, compressed blobs, hashes, images.

Because there is no warning and no indicator, the failure surfaces far from its cause. A hash comparison fails, a blob will not decompress, a geometry is silently the wrong shape. Given that multiples of 1,024 pass cleanly, a smoke test can easily be built out of exactly the sizes that work.

## Environment

```
R           4.5.2 (2025-10-31)
odbc        1.7.0
DBI         1.3.0
brickster   0.2.14
driver      RStudio Spark ODBC Driver 2.7.7.1016 (libsparkodbc_sb64.so)
DBSQL       2026.20
dbms        hive_metastore 3.1.1
compute     serverless SQL warehouse
```

## What I have not established

- Only `DefaultBinaryColumnLength` was swept. Other driver options may affect this, so "no setting fixes it" is not established: only that no setting I tried did.
- One warehouse, one driver version, one DBSQL version, one `R`/`odbc` pair. Nothing here says anything about other versions.
- Serverless SQL warehouse only. No all-purpose cluster or DBR path tested.
- No payload above 42,000 bytes, and only synthetic payloads: no test against a real WKB geometry column.
- The buffer-wrap reading is inferred from returned lengths, not confirmed against driver source or driver logs.
- **Which `odbc` code path is at fault.** The `pyodbc` comparison narrows it to the client, but I have not read `odbc`'s source or built it with instrumentation, so I cannot point at a specific buffer or `SQLGetData` call. `isql` was tried first and cannot help: it truncates its own binary display at a fixed width.
- **Whether this affects other backends.** Only tested against the Databricks/Spark ODBC driver. If the fault is in generic `SQL_C_BINARY` handling, other drivers with large `BINARY` columns may be affected too, and that is untested.

---

## Note for the Databricks/Simba side

Kept short deliberately: on this evidence there is nothing for Simba to fix. `pyodbc` reads the same `BINARY` columns intact through their driver, so the driver moves the bytes correctly. If anything is worth passing on, it is only that the R client has a `BINARY` buffer bug against their driver, in case they field a report about it from another direction.

jeroen:
Fixes #1024.

## Problem

When retrieving unbound `SQL_C_BINARY` columns, nanodbc reads the value in 1024-byte chunks via repeated `SQLGetData` calls. The loop in `get_ref_impl<std::vector<std::uint8_t>>` only appended a chunk when the length indicator was positive:

```cpp
if (ValueLenOrInd > 0) { ... out.insert(...) ... }
```

Per the ODBC spec, a driver may instead set the indicator to `SQL_NO_TOTAL` (−4) on intermediate `SQL_SUCCESS_WITH_INFO` calls when it does not know the remaining data length — which is what the Databricks/Simba Spark driver does. Since −4 fails the `> 0` test, every full 1024-byte chunk was silently discarded, and only the final chunk (returned with `SQL_SUCCESS` and a real length) was kept. That yields exactly the behavior reported in #1024: `n mod 1024` bytes returned, with exact multiples of 1024 surviving intact.

(One note on the issue's analysis: the returned bytes are the *last* chunk, not a prefix — the reporter's prefix check used a period-256 payload with a length that is a multiple of 1024, where the two are byte-identical.)

## Fix

Handle `SQL_NO_TOTAL` by appending the full buffer, exactly mirroring the existing `SQL_C_CHAR` and `SQL_C_WCHAR` blob paths a few lines above (which already have this branch). On `SQL_SUCCESS_WITH_INFO`/01004 with `SQL_C_BINARY` the driver fills the entire buffer, so appending `buffer_size` is correct.

## Testing

There is no regression test because reproducing this requires a driver that reports `SQL_NO_TOTAL` for binary chunks (the Simba Spark driver does; the drivers in CI report actual remaining lengths, which take the already-correct `ValueLenOrInd > 0` path). Verification against a Databricks warehouse per the repro in #1024 would be welcome.

Upstream nanodbc `main` has the identical gap in the binary path; I'd suggest upstreaming this patch as well.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # odbc (development version)
 
+* Fixed silent truncation of `BINARY`/`VARBINARY` data larger than 1024 bytes
+  when the driver does not report the remaining data length (`SQL_NO_TOTAL`)
+  during chunked retrieval, as seen with the Databricks/Simba Spark
+  driver (#1024).
+
 * Numeric connection string arguments no longer fall back to scientific
   notation in `build_connection_string()`, which avoids malformed driver
   attributes such as `DefaultStringColumnLength` (#934).
--- a/src/nanodbc/nanodbc.cpp
+++ b/src/nanodbc/nanodbc.cpp
@@ -4273,7 +4273,9 @@ inline void result::result_impl::get_ref_impl<std::vector<std::uint8_t>>(
                     buffer,          // TargetValuePtr
                     buffer_size,     // BufferLength
                     &ValueLenOrInd); // StrLen_or_IndPtr
-                if (ValueLenOrInd > 0)
+                if (ValueLenOrInd == SQL_NO_TOTAL)
+                    out.insert(std::end(out), buffer, buffer + buffer_size);
+                else if (ValueLenOrInd > 0)
                 {
                     auto const buffer_size_filled =
                         std::min<std::size_t>(ValueLenOrInd, buffer_size);


jeroen:
Upstreamed to nanodbc as nanodbc/nanodbc#475.

andrie:
Removing the comment.  I meant to post this in #1024

simonpcouch (approved):
Yup, looks good. Thank you!

jacpete:
Hi all I'm checking on the status of this since it's been approved for a month and it looks like we may have a customer hitting this issue in ticket 133328. Is this expected to be merged and released soon?

detule:
Hi @jacpete 

Just merged it.
We just released an update of `odbc` - I will circle back with the team to see if we can release another version "out-of-cycle"; in the meantime Is your customer able to install and use the version of the package from `master`?

Cheers

PR #1026, opened 2026-08-21
databricks(): set DefaultStringColumnLength to 65535 by default

### LINKED ISSUE #1023: `databricks()` silently truncates STRING columns at 1,023 characters
## What happens

`odbc::databricks()` returns STRING columns truncated to 1,023 characters, with no warning, no error and no condition raised. The value you get back is a valid 1,023-character string, so a caller cannot distinguish it from genuinely short data without independently knowing the true length.

```r
con <- DBI::dbConnect(odbc::databricks(), httpPath = Sys.getenv("DATABRICKS_PATH"))
DBI::dbGetQuery(con, "SELECT repeat('x', 5000) AS s")$s |> nchar()
#> 1023
```

Passing `DefaultStringColumnLength = 65535` to `dbConnect()` fixes it completely:

```r
con <- DBI::dbConnect(
  odbc::databricks(),
  DefaultStringColumnLength = 65535
)
DBI::dbGetQuery(con, "SELECT repeat('x', 5000) AS s")$s |> nchar()
#> 5000
```

On v1.7.0, pass `65535` rather than a rounder `100000`: see the #934 interaction below.

Nothing about this is specific to Databricks' spatial or nested types. `repeat('x', n)` is enough, and any free-text column, notes field, JSON blob or WKT geometry over 1,023 characters is affected.

## The boundary is exactly 1,023/1,024

| Characters requested | Default | `DefaultStringColumnLength = 65535` |
|---|---|---|
| 1,000 | 1,000 | 1,000 |
| 1,023 | 1,023 | 1,023 |
| 1,024 | **1,023** | 1,024 |
| 2,000 | **1,023** | 2,000 |
| 5,000 | **1,023** | 5,000 |
| 20,000 | **1,023** | 20,000 |

## Why the silence is the problem

The truncation itself is a driver default and arguably working as configured. What makes it bite is that nothing tells you it happened.

I tested for that directly rather than inferring it from console output. Wrapping the 5,000-character query in `withCallingHandlers()` and capturing both condition classes:

```r
w <- NULL; m <- NULL
res <- withCallingHandlers(
  DBI::dbGetQuery(con, "SELECT repeat('x', 5000) AS s"),
  warning = function(c) { w <<- conditionMessage(c); invokeRestart("muffleWarning") },
  message = function(c) { m <<- conditionMessage(c); invokeRestart("muffleMessage") }
)
w              #> NULL
m              #> NULL
nchar(res$s)   #> 1023
```

No condition of any kind is signalled. The truncation is a clean prefix, not corruption: for the 5,000-character case the returned value is byte-identical to `strrep("x", 1023)`, so the first 1,023 characters are right and the remainder is simply absent.

## This compounds with #934 on the current release

#934 reported that `build_connection_string()` renders large numerics in scientific notation, and
the reporter hit it with `DefaultStringColumnLength` specifically: `100000` became `1e+05`, which the driver silently ignored. That is fixed by #1016, merged after v1.7.0 shipped, so it is not in the current CRAN release.

The two interact badly on 1.7.0. The truncation is silent, and the workaround for it is also silently dropped if you pick a round number:

```r
odbc:::build_connection_string(list(DefaultStringColumnLength = 65535))
#> "DefaultStringColumnLength=65535"    # fine
odbc:::build_connection_string(list(DefaultStringColumnLength = 100000))
#> "DefaultStringColumnLength=1e+05"    # silently ignored by the driver
```

It is not a simple size threshold, which makes it harder to spot. R uses scientific notation only when it is shorter than the fixed form, so round numbers are the ones that break while longer values pass:

| Value passed | Rendered on 1.7.0 |
|---|---|
| `65535` | `65535` |
| `99999` | `99999` |
| `100000` | `1e+05` |
| `123456` | `123456` |
| `200000` | `2e+05` |
| `1000000` | `1e+06` |

On 1.7.0 the working forms are a literal string `"100000"`, an integer `100000L`, or `format(100000, scientific = FALSE)`. Note that `as.character(100000)` does **not** work, because the coercion happens before the string is built and yields `"1e+05"`.

So a user on the current release who hits the truncation, finds `DefaultStringColumnLength`, and reaches for a round number gets a fix that appears to do nothing. Two silent failures in sequence.

## Suggestion

Since `databricks()` exists to supply sensible defaults for this backend, the fix probably belongs there rather than in every caller. Either:

- set a larger `DefaultStringColumnLength` in `databricks()` by default, or
- document in `?databricks` that the driver default caps STRING at 1,023 characters, and name `DefaultStringColumnLength` as the remedy

I have no view on which is right, and the first may well have memory implications you understand better than I do. #856 covers similar ground for `EnableQueryResultDownload`, another Simba parameter with an awkward default, and the discussion there leaned towards documenting rather than overriding a Databricks default. That issue is still open, so I am reading it as related prior discussion rather than a settled precedent. Happy to send a PR for either once you have picked.

If it stays a documentation change, the `?databricks` note is more useful if it names a value that survives #934 on the current release, or points at the string form.

## Environment

```
R 4.5.2 (2025-10-31)
odbc 1.7.0
DBI 1.3.0
driver: RStudio Pro Drivers, Spark ODBC 2.7.7.1016 (libsparkodbc_sb64.so)
DBSQL 2026.20, Azure Databricks serverless SQL warehouse
```

Reproduced on two separate sessions, 2026-08-18 and 2026-08-20, with identical results at every size in the table above.

## What I have not established

The scope here is narrow and I would rather say so than have you assume otherwise:

- one warehouse, one DBSQL version (2026.20), one driver version (2.7.7.1016), so this says nothing about other versions
- serverless SQL warehouse only, with no all-purpose cluster or DBR path tested
- no intermediate `DefaultStringColumnLength` values tried, so I cannot say where the useful floor sits between the default and 65535
- I have not read the driver's own documentation on what its default actually is, so I cannot confirm 1,023 is the documented figure rather than an artefact of this setup

jeroen:
Fixes #1023.

## Problem

The Databricks/Simba Spark ODBC driver describes `STRING` columns to clients as `VARCHAR(DefaultStringColumnLength)` and silently truncates longer values. The driver's own default of 255 is far too small for real-world STRING data (free text, JSON blobs, WKT geometry).

Why the observed boundary in #1023 is 1023 rather than 255: we compile nanodbc with `-DNANODBC_OVERALLOCATE_CHAR`, so a `VARCHAR(255)` column gets a bound buffer of `(255 + 1) × 4 = 1024` bytes, and the driver fills the buffer we hand it rather than stopping at its declared column size — 1023 characters plus the NUL terminator. So the data loss is driver-configured, and the odd number is an artifact of our 4-bytes-per-codepoint overallocation.

## Changes

- `databricks_default_args()` now sets `defaultStringColumnLength = 65535`. With a fetch rowset size of 1 and the 4× overallocation this costs at most ~256 KB of bound buffer per string column, which is negligible.
- User-supplied connection arguments now override built-in defaults **case-insensitively**. Connection string keys are case-insensitive to drivers, and users following the driver docs will pass `DefaultStringColumnLength` (PascalCase); previously that would have been emitted alongside our camelCase default as a duplicate key. This also fixes the same latent hazard for existing defaults like `thriftTransport`.
- Documented the driver's truncation behavior and the new default in `?databricks`.

`man/databricks.Rd` was edited by hand to match the roxygen comment because my local roxygen2 (8.0.0) is newer than the repo's `RoxygenNote` (7.3.3) and regenerating churned every file — feel free to re-`document()`.

## Testing

Added unit tests for the new default and for case-insensitive overrides; the existing camelCase-invariant and override tests still pass.

Note the interaction with #934/#1016: on CRAN odbc 1.7.0, a round-number `DefaultStringColumnLength` passed by users was rendered in scientific notation and silently ignored by the driver. That is already fixed on `main`, and this default is applied after that fix, so `65535` (or any user override) reaches the driver correctly.


R/driver-databricks.R
```diff
@@ -135,7 +144,13 @@ databricks_args <- function(httpPath,
...
 
   auth <- databricks_auth_args(host, uid = uid, pwd = pwd)
-  all <- utils::modifyList(c(args, auth), list(...))
+  dots <- list(...)
+  # Connection string keys are case-insensitive to the driver, so drop any
+  # default that the caller respells (e.g. DefaultStringColumnLength) rather
+  # than emitting the key twice.
+  defaults <- c(args, auth)
+  defaults <- defaults[!tolower(names(defaults)) %in% tolower(names(dots))]
+  all <- utils::modifyList(defaults, dots)
```
simonpcouch: The case-sensitive default override doesn't seem necessary as part of this change. There's already some weird behavior with case-sensitivity here, and I don't know that this change resolves it. e.g. a user could provide `Driver` and the package automatically provides `driver` as a default. Before, both would end up in the connection string. The new path would remove `$driver` in favor of `$Driver`, but then package code later assumes `$driver`.

R/driver-databricks.R
```diff
@@ -168,6 +183,10 @@ databricks_default_args <- function(driver, host, httpPath, useNativeQuery) {
     httpPath = httpPath,
     thriftTransport = 2,
     userAgentEntry = databricks_user_agent(),
+    # The driver reports STRING columns as VARCHAR(DefaultStringColumnLength)
+    # and truncates longer values without warning; its default of 255 is far
+    # too small for real-world STRING data (#1023).
+    defaultStringColumnLength = 65535,
```
simonpcouch, line 183: From my understanding, this default is conservatively set so as to allocate memory effectively. The new default can cause very large allocations for vectorized parameter queries, because `dbBind()` defaults `batch_rows` to the parameter-vector length, which becomes nanodbc’s result rowset size.

simonpcouch: My preference here, if possible, would be to actually forward a truncation alert from the driver as a warning or error.

Apologies for pasting LLM output directly here, but pretty swamped right now. Initial reaction from Codex is that this is viable:

> ODBC has a standard truncation signal: `SQLFetchScroll()` should return `SQL_SUCCESS_WITH_INFO` with SQLSTATE `01004`. The current nanodbc path treats that as ordinary success:
> 
> ```cpp
> if (!success(rc))
>     throw ...;
> return true;
> ```
> 
> Since `success()` includes `SQL_SUCCESS_WITH_INFO`, any diagnostic is discarded. The absence of an R warning therefore does not prove the driver failed to report truncation. [ODBC specifies `01004` for this case.](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/data-length-buffer-length-and-truncation?view=sql-server-ver17)
> 
> The first step should be instrumenting that fetch and checking its diagnostics. If the driver returns `01004`, odbc can surface an exact warning or error without changing the default.
> 
> If it returns plain success, the length indicator might still contain the original length. Comparing that with the bound-buffer size would also detect truncation exactly.
> 
> If neither signal survives, exact detection is impossible from the returned value. A full buffer could mean either truncation or a legitimately exact-length string. odbc could warn that truncation “may” have occurred, but that risks false positives.
> 
> The more robust alternative would be fetching Databricks strings dynamically with `SQLGetData()` rather than preallocating from the reported `VARCHAR` width. That could avoid both silent truncation and the large fixed allocation, though it requires testing that the driver returns the complete value through that path.

simonpcouch (changes requested):
Thanks for making this happen! Increasing that allocation size is not my preferred route here; if possible, it'd be great if we could forward a truncation alert from the driver and stick with the current default.

If you disagree, I might say we wait until @hadley has a moment to look and decide. This change is a bit out of my wheelhouse, so I can't say for sure, but I wouldn't be comfortable approving the current proposal myself. :)

R/driver-databricks.R
hadley: Let's pull that out into a separate PR with a helper that we can use everywhere (I assume it's a problem broadly, not just with databricks).

simonpcouch: You got it, broke that out into https://github.com/r-dbi/odbc/issues/1034

R/driver-databricks.R
simonpcouch: Okay, just checked in with a collaborator and it seems like the approach in this PR as-is will be fine.

hadley (approved):

simonpcouch (approved):

simonpcouch:
Thank you!

